PR #4, opened 2026-03-16
[bazel] Rename Bazel macros

vlpannel:
Currently, many of the repo's Bazel macros are prefixed with `opentitan_*`. As development on this repo is expected to diverge from the original OpenTitan repository, it would be wise to rename such macros. These changes use `pavona_*` instead to avoid confusion.

This PR also updates the Bazel target `//rules/opentitan` to `//rules/pavona`, assuming development will still augment with the same set of rules rather than creating a whole new build target.

For now, it leaves `opentitantool`, `opentitanlib`, etc. alone because these changes would likely break more functionalities.


apinise:
Thanks for working on this! Given the number of changes I wasn't able to check absolutely everything, but I tested a few random SiVal targets from `sw/device/tests/BUILD` and crypto tests in `sw/device/tests/crypto/BUILD`, all of which worked. I tried running some of the lint checks and starlark from `quality/BUILD` and these give an error, but it is because there is no `rfc` directory, once that is there they work fine.

One thing I'd recommend is to regenerate the MODULE.bazel.lock file after having done all of these changes, as I observed locally a few of the digests changed during testing.

`./bazelisk.sh mod deps --lockfile_mode=refresh` should do the trick

Also `sw/device/silicon_creator/lib/drivers/BUILD` has a few linting errors you can resolve with `./bazelisk.sh run //quality:format_Starlark_with_buildifier`

apinise (approved):
LGTM! Might be a few merge conflicts depending on when other PRs get merged.

PR #13, opened 2026-03-17
[topgen] Removing alert and cpu crashdumps and CSRs from rstmgr

LouisTheLuis:
Removing the `alert_handler` and/or the `rv_core_ibex` from the top config should remove the corresponding crashdumps and CSRs.


matutem (approved):
LGTM

PR #18, opened 2026-03-18
[hmac,V2S,S1] Downgrade V stage to V2s and upgrade S stage to S1

matutem:
This has two commits:
- Downgrade V stage to V2s
- Upgrade S stage to S1


qmn (approved):
LGTM, thank you!

PR #19, opened 2026-03-19
[doc] Documentation wording clarifications

vlpannel:
This PR specifies the more vague usage of "OpenTitan" to:
* "Darjeeling" for boot ROM patching
* More top-specific language ("Earlgrey" or "Darjeeling") for the respective AST modules
* "Darjeeling" for Darjeeling's soc_proxy
* "root-of-trust"/"RoT" for mbx
* Generallly more generic language for dma

As Pavona is expected to diverge from OpenTitan, the documentation should reflect this. References to OpenTitan (which originally could refer to any one of: the project, a top level, the repository, an actualized product) have been swapped with more specific language, such as "hardware root-of-trust" or references to a specific top level system (Earlgrey or Darjeeling).


hw/ip/dma/data/dma.hjson
qmn, line 1: I think there's a larger conversation about what DMA is moving data blobs in to and out of. It could be a root of trust, but really it can be anywhere. I think the verbiage needs to be more generic overall: it just moves blobs of memory. It _just so happens_ that if it is within a top-level design that serves as a root of trust, it can handle such transfers securely (that is, meeting the needs of said root of trust).

hw/ip/mbx/data/mbx.hjson
qmn, line 1: Note that unlike the DMA, the mailbox serves a much more security-focused purpose of isolating memory in some integrated RoT from the broader SoC. So language about RoTs is warranted here.

hw/ip/soc_dbg_ctrl/doc/programmers_guide.md
qmn, line 1: Could you please file the corresponding issues?

qmn (commented):
Please have a look at these comments. Indeed, we need to come up with a concise way of referring to a top built from the Pavona ecosystem without calling it a Pavona.

hw/ip/soc_dbg_ctrl/doc/programmers_guide.md
vlpannel: #74

hw/ip/dma/data/dma.hjson
vlpannel: Would it be better instead to say that the DMA moves data from a **source [memory]** to the rest of the SoC?

Agreed that a DMA can be more generic, but this documentation also notes security features of this DMA (I guess the subsystem terminology would still work here as a "secure/private source").

Brief survey of other possible terminology:
- "Private memory", "internal memory", "isolated memory"
- [Xilinx/AMD](https://docs.amd.com/r/en-US/pg034-axi-cdma/Introduction) seems to use "source" address vs "destination" address
- [Wikipedia](https://en.wikipedia.org/wiki/Direct_memory_access) uses "main memory" (though I guess this wouldn't work for an integrated RoT) vs "hardware subsystems"
- "Memory" vs "device"/"peripheral" (but doesn't capture memory isolation/security angle)

hw/ip/dma/data/dma_testplan.hjson
```diff
@@ -258,7 +258,7 @@
 
               Checking:
               - Check if STATUS.aborted bit is set (This indicates DMA operation is aborted)
-              - Check if there all outstanding transactions on OT internal TLUL interface are complete
+              - Check if there all outstanding transactions on secure internal TLUL interface are complete
```
qmn:
```suggestion
              - Check that all outstanding transactions on secure internal TLUL interface are complete
```

qmn (approved):
Just one minor change but LGTM after that goes in.

PR #22, opened 2026-03-23
[spi_device,dv] Fix sim_cfg name

LouisTheLuis:
Make `spi_device` follow the convention of other IPs of having its `base_sim_cfg` named after the IP itself.


vlpannel (approved):
LGTM.

PR #24, opened 2026-03-23
[kmac,doc] Updated KMAC documentation

apinise:


LouisTheLuis (approved):
Not completely familiar with KMAC but given it's mostly stylistic changes, it LGTM.

PR #25, opened 2026-03-23
[dv,dvsim] Standardize generation of scratch directories

LouisTheLuis:
This PR is composed of two commits:
- The first one makes it so that running `dvsim` in a `base_sim_cfg` generates a `_base` scratch subdirectory for consistency sake.
- The second allows for `dvsim` simulations generating separate directories for separate tops. For e.g. simulations of `clkmgr` for `hw/top_earlgrey` and `hw/top_darjeeling` would generate reports / logs on `clkmgr_top_earlgrey` and `clkmgr_top_darjeeling` respectively.


vlpannel (approved):
LGTM.

PR #26, opened 2026-03-23
[dv] Fix overrides in sim_cfg files

LouisTheLuis:
Include override for `acc` and remove obsolete ones for other IP.


apinise (approved):
LGTM!

PR #32, opened 2026-03-24
[entropy_source,rtl] Changed sha3_squeezing net to MuBi4 to match sha3 port

apinise:
The sha3_squeezing_o port in sha3 was previously made a MuBi4 but was not updated in entropy source. This caused CSR reads to entropy source to misinterpret the instantiated sha3 squeezing status.


LouisTheLuis (approved):
LGTM.

Issue #34, opened 2026-03-25
[hw,dv,doc] Replace book, doc_server, repo_server, and results_server variables


LouisTheLuis:
### Description

These are defaulted to OT and a lot of the docs generated with them currently assume the old pages. They should be replaced promptly.

PR #36, opened 2026-03-25
Dashboard creation

LouisTheLuis:
This PR includes 4 commits:
- Creation of a dashboard script that pulls all the corresponding coverage reports and places them in a convenient output directory.
- Changes to `sim_cfg` files for the sake of standardization.
- Fix to `dvsim.py` that makes finding the run reports easier. It changed the reports to be generated under `./scratch/{git_branch or default}/{ip_name}/reports/` instead of `./scratch/{git_branch or default}/reports/{relative path of the IP in the repo}`.
- Include run reports in the aforementioned dashboard script.

Making a note of issue #34, which needs to be addressed later.


util/create-dashboard.sh
```diff
@@ -0,0 +1,305 @@
...
+  # Pattern 5: hw/{top}/ip_autogen/{ip_name}/dv/{sim_cfg}
+  for cfg_file in hw/${top}/ip_autogen/*/dv/*_sim_cfg.hjson; do
+    [ -f "$cfg_file" ] || continue
+    sim_name=$(basename "$cfg_file" "_sim_cfg.hjson")
+    [[ "$sim_name" == *base* ]] && continue
+    ip_name=$(echo "$cfg_file" | cut -d'/' -f4)
+    JOBS+=("${cfg_file}|${ip_name}|${sim_name}|autogen|${top}")
+  done
+
+  # Pattern 6: hw/{top}/ip_autogen/{ip_name}/dv/{sim_cfg}
```
matutem: This seems identical to Pattern 5 above. What am I missing?

util/create-dashboard.sh
```diff
@@ -0,0 +1,305 @@
+#!/bin/bash
+
```
matutem, line 2: This seems to combine two things: running dvsim and retrieving the results. It may be useful to separate the two? I normally run regressions directly using dvsim, and I suppose the weekly and nightly do as well.

util/create-dashboard.sh
LouisTheLuis: Oop, the comment doesn't reflect it. Pattern 6 is basically a solution for the corner case of `rstmgr_cnsty_check`.

matutem: Are you also listing the prim/dv items?

matutem:
It is not obvious from the comments, but perhaps the PR description should explain how the paths for the reports under scratch changed.

util/create-dashboard.sh
LouisTheLuis: Yep, those are covered by pattern 3.

util/create-dashboard.sh
LouisTheLuis: Ideally this could be a catch-all tool for this process. I do agree that it can be split into two scripts (or three, as publishing is also a process separated from running `dvsim` and collecting the reports). 

I'll update the script to be split into multitop / earlgrey / darjeeling as well.

vlpannel (commented):
LGTM. No comments, but would recommend more eyes to also check.

PR #50, opened 2026-03-26
 [ci,dv] Enforce standard for sim_cfgs

LouisTheLuis:
This PR includes two commits:
- Fix certain `sim_cfg` names.
- Create a `sim_cfg` checker that makes sure that each `sim_cfg` file follows a series of constraints.

The reasoning behind this commit is that in order for `dvsim` to properly generate scratch directories that can be easily collected (e.g. for a dashboard), the `sim_cfg` files should follow a particular pattern. This enforces that in such a way that if any user wants to include another `sim_cfg` file, it's retrievable easily by any other script.


qmn:
This looks good! It would be good to document the legal names for a sim config in the documentation, probably the DVSim README. There are apparently about four cases:
* A single sim config,
* A sim config with a base and multiple variants that import that base,
* A sim config with multiple variants (but no base), and
* A sim_**cfgs** file, only for tops, which imports individual sim configs above.
These all have implications for what the name should be.

EDIT: And it looks like you need to chmod +x the script.

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,157 @@
...
+    This function searches for *_sim_cfg.hjson or *_sim_cfgs.hjson
+    files and groups them in lists based on which directory they exist
+    within.
+    For e.g. for a ipname with {ipname}_base_sim_cfg.hjson, and 
+    {ipname}_variant_sim_cfg.hjson, they would both be grouped in:
+    
+    grouped["/hw/ip/{ipname}"] = [{ipname} sim cfgs]
+    """
+    grouped = defaultdict(list)
+    for path in SEARCH_ROOT.rglob("*"):
```
matutem: Could you let rglob do more of the work? Try something like this:

```
all_scs = SEARCH_ROOT.glob("**/*_sim_cfg[s]*.hjson")
grouped = {str(p.parent): (p.name, get_fields_from_hjson(p)) for p in all_scs}
```

vlpannel: Seconded, with minor modification:
```diff
  all_scs = SEARCH_ROOT.glob("**/*_sim_cfg[s]*.hjson")
+ dirs = [p.parent for p in all_scs]
+ grouped = {str(d): [(s.name, get_fields_from_hjson(s)) for s in all_scs if s.parent == dir] for d in dirs}
```

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+                # Pattern 1: base + variant mode
+                # The base file is the one with 'name' field
+                base_files = [(f, d) for f, d in sim_cfg_files if "name" in d]
+
+                if len(base_files) != 1:
+                    for file, _ in base_files:
+                        mismatches.append(
+                            f"Filename {file} - exactly one file with 'name' field "
+                            f"expected in base+variant directory"
+                        )
```
vlpannel: Not sure if this is applicable, but I can see an edge case here: what if there are no files with the `name` field? No mismatches will be added, but there is not _exactly_ one base sim_cfg.

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+                        elif "name" in fields:
+                            msg = f"Filename {file} - variant file must not contain 'name' field"
+                            mismatches.append(msg)
+                        elif "variant" not in fields:
+                            msg = f"Filename {file} - variant file must contain 'variant' field"
+                            mismatches.append(msg)
+                        else:
+                            expected = f"{base_name}_{fields['variant']}_sim_cfg.hjson"
+                            if file != expected:
+                                mismatches.append(f"Filename {file} should be named {expected}")
```
vlpannel: You've already checked for if `name` is in fields on lines 68-90 (there's only one base cfg here), so the only thing left to check is if `variant` is in the fields or not.

This could be shortened to just:
```python
if "variant" in fields:
    expected = f"{base_name}_{fields['variant']}_sim_cfg.hjson"
    if file != expected:
        mismatches.append(f"Filename {file} should be named {expected}")
```

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+                    elif "variant" in fields:
+                        msg = f"Filename {file} - file must not contain 'variant' field"
+                        mismatches.append(msg)
+                    else:
+                        name = fields['name']
+                        valid = [f"{name}_sim_cfg.hjson", f"{name}_base_sim_cfg.hjson"]
+                        if file not in valid:
+                            mismatches.append(
+                                f"Filename {file} should be named {valid[0]} or {valid[1]}"
+                            )
```
vlpannel: Similar to previous comment, you've already filtered out some things: there is no `variant` in any of the sim_cfgs (lines 61, 63, 104).

Hence this can be shortened to:
```python
if "name" in fields:
    name = fields['name']
    valid = [f"{name}_sim_cfg.hjson", f"{name}_base_sim_cfg.hjson"]
    if file not in valid:
        mismatches.append(
            f"Filename {file} should be named {valid[0]} or {valid[1]}"
        )
```

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+        for file, fields in sim_cfgs_files:
+            dir_path = Path(directory)
+            if dir_path == SEARCH_ROOT / "hw" / "dv":
+                expected = "top_sim_cfgs.hjson"
+                if file != expected:
+                    mismatches.append(f"Filename {file} should be named {expected}")
+                elif 'name' not in fields:
+                    mismatches.append(f"Filename {file} should contain field 'name'")
+            elif (dir_path.name == "dv" and
+                  dir_path.parent.parent == SEARCH_ROOT / "hw" and
```
vlpannel: Nit: pathlib allows use of `parents` iterable.
```suggestion
                  dir_path.parents[1] == SEARCH_ROOT / "hw" and
```

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+                  dir_path.parent.name.startswith("top_")):
```
vlpannel: Could also use `Path.glob` here:
```suggestion
            elif dir_path.resolve() in SEARCH_ROOT.glob("hw/top_*/dv"):
```

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+                topname = dir_path.parent.name
+                expected = f"{topname}_sim_cfgs.hjson"
+                if file != expected:
+                    mismatches.append(f"Filename {file} should be named {expected}")
+                elif 'name' not in fields:
```
vlpannel, line 143: Maybe this should be an `if`, not an `elif` (_unless only the top\_\* file needs the `name` field; if so, then disregard this comment_).

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+import sys
+from pathlib import Path
+from collections import defaultdict
+import hjson
+
+SEARCH_ROOT = Path(__file__).resolve().parents[2]
```
vlpannel, line 11: At first, I was getting some mismatches when running this script locally because I'd previously generated some docs on another branch (which copies files into the `build-site/` dir). Git checkout did not update this directory, as it's ignored by git.

Nit: it might be wise to have this search also ignore the globs listed in the `.gitignore` or something of the like.

ci/scripts/check-sim-cfgs.py
```diff
@@ -0,0 +1,163 @@
...
+    grouped["/hw/ip/{ipname}"] = [{name} sim cfgs]
+    """
+    grouped = defaultdict(list)
+    for path in SEARCH_ROOT.rglob("*"):
+        if path.is_file() and (
+            path.name.endswith("_sim_cfg.hjson") or
+            path.name.endswith("_sim_cfgs.hjson")
+        ):
+            grouped[str(path.parent)].append((
+                path.name,
```
vlpannel: Would it be better to leave this as the full path?

It would be nice for the error messages to show full paths rather than just the file names (otherwise, it will be hard to find the file to debug).
```suggestion
                path,
```

vlpannel (approved):
With some minor changes for code readability, but overall LGTM. Nice work on condensing a somewhat complicated flow graph of conditions into a Python script.

ci/scripts/check-sim-cfgs.py
vlpannel: Actually, see comment below.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Done. Unfortunately dictionary interpretation is not ideal for this problem because each key `str(p.parent)` corresponds to a list of `.hjson` files living in that directory. Using a dictionary interpretation rewrites the key instead of appending; hence the `for` loop. 

Also, `rglob` doesn't seem to support optional characters like a regex would. And using `_sim_cfg*.hjson` wouldn't be ideal because there are files like `partner_chip_sim_cfg_sample.hjson` that are not used for being ran by `dvsim`.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Good catch! That should be a mismatch as well. The `for loop` would find an empty list and wouldn't print. Fixed.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Fixed.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Fixed.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Fixed.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: It does, otherwise it allows for `dvsim` runs to generate `-sim-*/` scratch directories instead of the more standard `top_*-sim-*/` directories.

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Using `p.name` is mostly a benefit of not having to construct the entire path whenever we compare. That said, we can print {directory}/{file} for each for better debugging. 
Uploading that :+1:

ci/scripts/check-sim-cfgs.py
LouisTheLuis: Maybe this can work?
```python
    gitignore = SEARCH_ROOT / ".gitignore"
    spec = pathspec.PathSpec.from_lines("gitwildmatch", gitignore.read_text().splitlines()) \
        if gitignore.exists() else None

    all_scs = (
        list(SEARCH_ROOT.rglob("*_sim_cfg.hjson")) +
        list(SEARCH_ROOT.rglob("*_sim_cfgs.hjson"))
    )
    if spec:
        all_scs = [p for p in all_scs if not spec.match_file(p.relative_to(SEARCH_ROOT))]
```

The idea is essentially collect from the `.gitignore` and then exclude any file that had a path within it.

PR #51, opened 2026-03-26
[util,ci] Automate collection of IPs

LouisTheLuis:
This PR introduces a script used by CI and `hw/Makefile` to auto-generate all top `sim_cfgs.hjson` files. This is useful for running project-wide or top-wide regressions and automates the process of adding a new IP to the repo.

### DIFF
--- a/hw/Makefile
+++ b/hw/Makefile
@@ -89,6 +89,9 @@ $(ips_reg): %_reg:
 
 top_and_cmdgen: top cmdgen
 
-all: top cmdgen regs
+dvgen:
+	${PRJ_DIR}/util/gen_ip_collection.py --flow sim_dv --tops top_earlgrey top_darjeeling
+
+all: top cmdgen regs dvgen
 
 .PHONY: all
--- a/util/gen_ip_collection.py
+++ b/util/gen_ip_collection.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""
+gen_ip_collection.py
+Collects IP and top-level sim cfg files and generates batch hjson files.
+Usage:
+    ./gen_ip_collection.py --flow sim_dv --tops top_earlgrey top_darjeeling
+"""
+import argparse
+import glob
+import os
+from pathlib import Path
+
+SEARCH_ROOT = Path(__file__).resolve().parents[1]
+LOWRISC_COPYRIGHT_LINE = "// Copyright lowRISC contributors (OpenTitan project)."
+ZERORISC_COPYRIGHT_LINE = "// Copyright zeroRISC Inc."
+LICENSE_LINES = """\
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0"""
+
+COPYRIGHT_HEADER_SOLO = f"{ZERORISC_COPYRIGHT_LINE}\n{LICENSE_LINES}"
+COPYRIGHT_HEADER_JOINT = (
+    f"{LOWRISC_COPYRIGHT_LINE}\n{ZERORISC_COPYRIGHT_LINE}\n{LICENSE_LINES}"
+)
+WARNHDR = """//
+// ------------------- W A R N I N G: A U T O - G E N E R A T E D   C O D E !! -------------------//
+// PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED WITH THE FOLLOWING COMMAND:
+"""
+
+
+def get_copyright_header(path, flow, tops):
+    """Return the appropriate copyright header for this file."""
+    tops_str = ' '.join(tops)
+    GENCMD = (f"// util/gen_ip_collection.py --flow {flow}\n"
+              f"//                --tops {tops_str}\n")
+    if os.path.exists(path):
+        try:
+            with open(path) as f:
+                first_line = f.readline().strip()
+            if first_line == LOWRISC_COPYRIGHT_LINE:
+                return f"{COPYRIGHT_HEADER_JOINT}\n\n{WARNHDR}{GENCMD}"
+        except OSError as e:
+            print(f"Warning: could not read {path}: {e}")
+    return f"{COPYRIGHT_HEADER_SOLO}\n\n{WARNHDR}{GENCMD}"
+
+
+# =============================
+#   Specific to sim_dv flow
+# =============================
+
+def collect_sim_dv(tops):
+    """
+    Collect all sim_cfg hjson files into two buckets:
+      - ip_cfgs:  shared across all tops (hw/ip/... and hw/dv/...)
+      - top_cfgs: dict keyed by topname, each containing top-specific cfgs
+    """
+    ip_cfgs = []
+    top_cfgs = {top: [] for top in tops}
+
+    def skip(sim_name):
+        return "base" in sim_name
+
+    # ---- IP bucket ----
+    # Pattern 1: hw/ip/{ip_name}/dv/{sim_cfg}
+    # Pattern 2: hw/ip/{ip_name}/dv/uvm/{sim_cfg}
+    # Pattern 3: hw/ip/prim/dv/{prim_variant}/{sim_cfg}
+    # Pattern 4: hw/dv/sv/{ip_name}/dv/{sim_cfg}
+    for cfg in (
+        sorted(glob.glob("hw/ip/*/dv/*_sim_cfg.hjson"))
+        + sorted(glob.glob("hw/ip/*/dv/uvm/*_sim_cfg.hjson"))
+        + sorted(glob.glob("hw/ip/prim/dv/*/*_sim_cfg.hjson"))
+        + sorted(glob.glob("hw/dv/sv/*/*/*_sim_cfg.hjson"))
+    ):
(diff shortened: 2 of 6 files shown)


util/gen_ip_collection.py
```diff
@@ -0,0 +1,301 @@
...
+LOWRISC_COPYRIGHT_LINE = "// Copyright lowRISC contributors (OpenTitan project)."
+ZERORISC_COPYRIGHT_LINE = "// Copyright zeroRISC Inc."
+LICENSE_LINES = """\
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0"""
+
+COPYRIGHT_HEADER_SOLO = f"{ZERORISC_COPYRIGHT_LINE}\n{LICENSE_LINES}"
+COPYRIGHT_HEADER_JOINT = (
+    f"{LOWRISC_COPYRIGHT_LINE}\n{ZERORISC_COPYRIGHT_LINE}\n{LICENSE_LINES}"
+)
```
vlpannel, line 26: Not sure what the copyright requirements should _actually_ be here, but I would also put in a comment noting that the output is autogenerated code.

For an example, see [`util/topgen.py`](https://github.com/pavona/pavona-staging/blob/main/util/topgen.py#L58-L70).

util/gen_ip_collection.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # Pattern 2: hw/ip/{ip_name}/dv/uvm/{sim_cfg}
+    for cfg in sorted(glob.glob("hw/ip/*/dv/uvm/*_sim_cfg.hjson")):
+        sim_name = Path(cfg).stem.replace("_sim_cfg", "")
+        if skip(sim_name):
+            continue
+        ip_cfgs.append(cfg)
+
+    # Pattern 3: hw/ip/prim/dv/{prim_variant}/{sim_cfg}
+    for cfg in sorted(glob.glob("hw/ip/prim/dv/*/*_sim_cfg.hjson")):
```
vlpannel: Nit: since `sorted` returns a list, it might be easier to concatenate these lists for one big for loop:
```python
# Searching for patterns:
#    - hw/ip/{ip_name}/dv/{sim_cfg}
#    - hw/ip/{ip_name}/dv/uvm/{sim_cfg}
#    - hw/ip/prim/dv/{prim_variant}/{sim_cfg}
#    - etc....
for cfg in (
    sorted(glob.glob("hw/ip/*/dv/*_sim_cfg.hjson"))
    + sorted(glob.glob("hw/ip/*/dv/uvm/*_sim_cfg.hjson"))
    + sorted(glob.glob("hw/ip/prim/dv/*/*_sim_cfg.hjson"))
  # + ...
):
    sim_name = Path(cfg).stem.replace("_sim_cfg", "")
    if skip(sim_name):
        continue
    ip_cfgs.append(cfg)
```

util/gen_ip_collection.py
```diff
@@ -0,0 +1,301 @@
...
+        ip_cfgs.append(cfg)
+
+    # Pattern 4: tl_agent
+    for cfg in sorted(glob.glob("hw/dv/sv/*/*/*_sim_cfg.hjson")):
+        ip_cfgs.append(cfg)
+
+    # ---- Top-specific bucket ----
+    for top in tops:
+        # Pattern 5: hw/{top}/ip_autogen/{ip_name}/dv/{sim_cfg}
+        for cfg in sorted(glob.glob(f"hw/{top}/ip_autogen/*/dv/*_sim_cfg.hjson")):
```
vlpannel: Nit: see previous comment on sorted list concatenation.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,301 @@
...
+    ip_cfgs, top_cfgs = collect_uvm_dv(tops)
+    print(f"  Found {len(ip_cfgs)} shared IP cfgs")
+    for top, cfgs in top_cfgs.items():
+        print(f"  Found {len(cfgs)} cfgs for {top}")
+
+    # Per-top hjson files
+    for top in tops:
+        out_path = f"hw/{top}/dv/{top}_sim_cfgs.hjson"
+        copyright_header = get_copyright_header(out_path)
+        content = generate_top_hjson(top, ip_cfgs, top_cfgs[top], copyright_header)
```
vlpannel: `generate_top_hjson` collects all IPs, including those that are not within a top's list of modules. I'm assuming you're deliberately testing even the IPs not within a given top?

vlpannel (commented):
Some minor remarks, but will defer to other reviewers for more input.

util/gen_ip_collection.py
LouisTheLuis: This works pretty well! Done for both.

util/gen_ip_collection.py
LouisTheLuis: Done.

util/gen_ip_collection.py
LouisTheLuis: Yes, given that these are used for dashboards, it is ideal that all IPs are there; including the ones that are not existing within any top (e.g. `acc_pqc`). It sounds like a good question however; in terms of dashboards, what is the best way to present the data given the constraints (such as not enough time to run three separate batch `-i all` simulations per week)?

LouisTheLuis: Ideally, `dvsim` should be able to support nested `use_cfgs`, but that is not currently the case.

util/gen_ip_collection.py
LouisTheLuis: Added a comment as such. For the license header, my understanding is that we want to keep the old headers intact.

util/gen_ip_collection.py
matutem: I think it will be useful to generate the report components incrementally (for example, once all tests of any IP are done we can generate that IPs full report). That way we will still get some results even if the full run fails at some point.

hw/Makefile
```diff
@@ -47,7 +47,7 @@ IPS ?= aes           \
        uart          \
        usbdev
 
-TOPS ?= top_darjeeling top_earlgrey top_englishbreakfast
+TOPS ?= top_earlgrey top_darjeeling top_englishbreakfast
```
matutem: It was in alphabetical order, why change it?

LouisTheLuis: Oop, this is a typo from when I thought of collecting IPs directly from the `TOPS` variable. Fixed.

matutem:
I see we dropped englishbreakfast. By design?

LouisTheLuis:
@matutem For the `gen_ip_collection.py` run? I didn't include `top_englishbreakfast` in the target because I realized it currently doesn't have an already existing batch `sim_cfg`. I was under the presumption that it didn't have DV support because of that; if my presumption is wrong then I can add it as well.

util/gen_ip_collection.py
matutem: This is most likely a subsequent PR, and will need an issue if you think it is a good idea.

LouisTheLuis: Another idea I think would be useful in `dvsim`: for batch runs, include a field (`batchgroups`?) that allows for the results to be grouped into separate report "buckets".

That way, we could simply run `hw/dv/top_sim_cfgs.hjson` (generated in this PR) and group the IPs in the final coverage run report as `batchgroup: earlgrey` (all the earlgrey IP), `batchgroup: darjeeling` (all the darjeeling IP), and `batchgroup: ip` (all the IP). 

`gen_ip_collections.py` could then be extended to pull the corresponding IP for each from their `{topname}.hjson`. 

And that way, only one run would be needed for generating all the coverage reports.

matutem:
In practive englishbreakfast only adds a few ip_autogens, typically generated differently from other tops, so it is handy. Plus running it weekly means it is less likely to suffer bit-rot. We could also run nightly for less tests, like we do for the other tops. I think we should either regress blocks or remove them.

LouisTheLuis:
Updated this PR to further streamline the code. Additionally, added this:
```
  overrides: [
    {
        name: scratch_path
        value: "{scratch_base_path}/{name}-{flow}"
    }
  ]
```
after some conversations. As batch configs allow for a mixed-tool flow, and some companies have licenses to support both VCS and Xcelium, it's a good idea to probably keep a batch run directory "tool-less".

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+            if skip(sim_name):
+                continue
+            top_cfgs[top].append(cfg)
+
+    return ip_cfgs, top_cfgs
+
+
+def format_use_cfgs(cfgs, indent=13):
+    """Format a list of cfg paths as hjson use_cfgs entries."""
+    lines = []
```
matutem: Use list comprehensions: replace lines 100, 102, and 103 by
    lines = [f'{pad}"{{proj_root}}/{cfg}",' for cfg in cfgs]

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def write_file(path, content, dry_run):
+    if dry_run:
+        print(f"\n{'=' * 60}")
+        print(f"Would write: {path}")
+        print('=' * 60)
+        print(content)
+    else:
+        os.makedirs(os.path.dirname(path), exist_ok=True)
```
matutem: File system operations should be within try-except blocks to deal with potential errors. Please review all other file ops in this script.

LouisTheLuis: Done.

util/gen_ip_collection.py
LouisTheLuis: Done.

matutem (approved):
LGTM

PR #52, opened 2026-03-27
[dvsim] Fix dvsim to produce better coverage reports

LouisTheLuis:
This PR includes 3 commits. One by one, they do the following:
- Change the `scratch` directory structure to include the run reports in `*-sim-*/reports/{timestamp or latest}/report.html` as opposed to the previous structure, where it would mimic the directory structure of the repository (e.g. for an IP in `hw/ip`, the report would be in `hw/ip/{ipname}/report.html`.
- Change absolute links to coverage reports in run reports to relative links. Also change links to test plans from `html` links to a book to links to the repository `hjson` files (to be changed later, note issue #34).
- Allow coverage summary numbers (e.g. group, code, line, etc.) to show up in `*_batch_sim_cfgs.hjson` run reports for each IP included, as well as some neater formatting for the HTML. Remove coverage reports link from run reports.

This commit is building on top of #51.

### DIFF
--- a/hw/Makefile
+++ b/hw/Makefile
@@ -89,6 +89,9 @@ $(ips_reg): %_reg:
 
 top_and_cmdgen: top cmdgen
 
-all: top cmdgen regs
+dvgen:
+	${PRJ_DIR}/util/gen_ip_collection.py --flow sim_dv --tops top_earlgrey top_darjeeling
+
+all: top cmdgen regs dvgen
 
 .PHONY: all
--- a/hw/data/common_project_cfg.hjson
+++ b/hw/data/common_project_cfg.hjson
@@ -2,11 +2,11 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 {
-  project:          opentitan
-  repo_server:      "github.com/lowrisc/opentitan"
-  book:             opentitan.org/book
-  doc_server:       docs.opentitan.org
-  results_server:   reports.opentitan.org
+  project:          pavona
+  repo_server:      "github.com/pavona/pavona-staging"
+  book:             pavona.org/book
+  doc_server:       docs.pavona.org
+  results_server:   reports.pavona.org
 
   // Default directory structure for the output
   scratch_base_path:  "{scratch_root}/{branch}"
@@ -14,8 +14,7 @@
 
   // Common data structure
   build_pass_patterns: []
-  // TODO: Add back FuseSoC fail pattern after
-  // https://github.com/lowRISC/opentitan/issues/7348 is resolved.
+  // TODO: Add back FuseSoC fail pattern
   build_fail_patterns: []
 
   exports: [
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -145,8 +145,7 @@ class FlowCfg():
         self._expand()
 
         # Construct the path variables after variable expansion.
-        self.results_dir = (Path(self.scratch_base_path) / "reports" /
-                            self.rel_path / "latest")
+        self.results_dir = (Path(self.scratch_path) / "reports" / "latest")
         self.results_page = (self.results_dir / self.results_html_name)
 
         tmp_path = (self.results_server + "/" + self.rel_path +
@@ -423,8 +422,7 @@ class FlowCfg():
             log.info("[scratch_path]: [%s] [%s]", item.name, item.scratch_path)
             item.write_results(self.results_html_name, item.results_md,
                                json_str)
-            log.log(VERBOSE, "[report]: [%s] [%s/report.html]", item.name,
-                    item.results_dir)
+            log.info("[report]: [%s] [%s/report.html]", item.name, item.results_dir)
             self.errors_seen |= item.errors_seen
 
         if self.is_primary_cfg:
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -421,21 +421,22 @@ class FlowCfg():
             log.info("[results]: [%s]:\n%s\n", item.name, result)
             log.info("[scratch_path]: [%s] [%s]", item.name, item.scratch_path)
             item.write_results(self.results_html_name, item.results_md,
-                               json_str)
-            log.info("[report]: [%s] [%s/report.html]", item.name, item.results_dir)
+                               json_str, is_primary_cfg=False)
+            log.info("[report]: [%s] [%s/report.html]\n\n", item.name, item.results_dir)
             self.errors_seen |= item.errors_seen
 
         if self.is_primary_cfg:
             self.gen_results_summary()
             self.write_results(self.results_html_name,
-                               self.results_summary_md)
+                               self.results_summary_md, is_primary_cfg=True)
+            log.info("[report]: [%s] [%s/report.html]", self.name, self.results_dir)
 
     def gen_results_summary(self):
         '''Public facing API to generate summary results for each IP/cfg file
         '''
         return
 
-    def write_results(self, html_filename, text_md, json_str=None):
+    def write_results(self, html_filename, text_md, json_str=None, is_primary_cfg=False):
         """Write results to files.
 
         This function converts text_md to HTML and writes the result to a file
@@ -452,7 +453,7 @@ class FlowCfg():
         # Write results to the report area.
         with open(self.results_dir / html_filename, "w") as f:
             f.write(
-                md_results_to_html(self.results_title, self.css_file, text_md))
+                md_results_to_html(self.results_title, self.css_file, text_md, is_primary_cfg))
 
         if json_str is not None:
             filename = Path(html_filename).with_suffix('.json')
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -838,8 +838,8 @@ class SimCfg(FlowCfg):
 
                 # self.rel_path gives us the path to the directory
                 # containing the sim_cfg file...
-                testplan = "https://{}/{}".format(
-                    self.book,
+                testplan = "https://{}/tree/main/{}".format(
+                    self.repo_server,
                     Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.html"
                 )
 
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -840,7 +840,7 @@ class SimCfg(FlowCfg):
                 # containing the sim_cfg file...
                 testplan = "https://{}/tree/main/{}".format(
                     self.repo_server,
-                    Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.html"
+                    Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.hjson"
                 )
 
             results_str += f"### [Testplan]({testplan})\n"
@@ -879,7 +879,7 @@ class SimCfg(FlowCfg):
                         if self.args.publish:
                             cov_report_page_path = "cov_report"
                         else:
-                            cov_report_page_path = self.cov_report_dir
+                            cov_report_page_path = "../../cov_report"
                         cov_report_page_path += "/" + self.cov_report_page
                         results_str += "({})\n\n".format(cov_report_page_path)
                     results_str += self.cov_report_deploy.cov_results
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -875,18 +875,20 @@ class SimCfg(FlowCfg):
                     results_str += "\n## Coverage Results\n"
                     # Link the dashboard page using "cov_report_page" value.
                     if hasattr(self, "cov_report_page"):
-                        results_str += "\n### [Coverage Dashboard]"
-                        if self.args.publish:
-                            cov_report_page_path = "cov_report"
-                        else:
-                            cov_report_page_path = "../../cov_report"
-                        cov_report_page_path += "/" + self.cov_report_page
-                        results_str += "({})\n\n".format(cov_report_page_path)
+                        results_str += "\n### Coverage Dashboard\n\n"
+                        # TODO: modify to include differentiation between public and private
+                        # dashboard usage.
+                        # if self.args.publish:
+                        #     cov_report_page_path = "cov_report"
+                        # else:
+                        #     cov_report_page_path = "../../cov_report"
+                        # cov_report_page_path += "/" + self.cov_report_page
+                        # results_str += "({})\n\n".format(cov_report_page_path)
                     results_str += self.cov_report_deploy.cov_results
                     self.results_summary[
-                        "Coverage"] = self.cov_report_deploy.cov_total
+                        "Overall Coverage"] = self.cov_report_deploy.cov_total
                 else:
-                    self.results_summary["Coverage"] = "--"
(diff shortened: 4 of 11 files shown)


matutem:
Luis, how do you suggest I try this to be able to view generated reports? I assume this means any coverage dvsim run will do the trick.

LouisTheLuis:
@matutem I would suggest running any batch `sim_cfg` (e.g. `top_darjeeling_sim_cfgs.hjson`) and run with the `--cov` flag. This will generate files in the `/scratch/{branch_name}/` directory. 

To view the files, I would do something like `python3 -m http.server $PORT --bind 0.0.0.0` for a particular `$PORT` of your preference. Based on the workstation you are working from, you should be able to look at the directory from `http://{workstation}:$PORT` in any browser.

For the batch summary, the `report.html` will be in `/scratch/{branch_name}/{batch_name-sim-*}/reports/latest/report.html`. You can navigate there and look at the coverage summaries.

**Note:** I usually delete most of the `use_cfgs` for running a batch `sim_cfg` so that I can run `-r 1 -i smoke --cov` and get quick results for testing.

matutem:
Something is funny: I checked out your PR to the branch pr_52, and then run
  ./util/dvsim/dvsim.py hw/top_earlgrey/dv/top_earlgrey_sim_cfgs.hjson --select-cfgs clkmgr -r 1 -v m -ro +max_quit_count=40 --waves fsdb -i all -r 2 --cov
The contents of scratch/pr_52 are

(ot-python) matute@shared-workstation:~/work/ps_a$ ls -l scratch/pr_52/
total 8
drwxr-xr-x  3 matute matute 4096 Apr  1 20:25 -sim-{tool}
drwxr-xr-x 66 matute matute 4096 Apr  1 20:25 clkmgr_top_earlgrey-sim-vcs

Some expansion didn't go well. Perhaps you are not handling a partial run using `select-cfgs`?

Nevertheless, the contents of -sim* are
(ot-python) matute@shared-workstation:~/work/ps_a$ find scratch/pr_52/-sim-\{tool\}/
scratch/pr_52/-sim-{tool}/
scratch/pr_52/-sim-{tool}/reports
scratch/pr_52/-sim-{tool}/reports/latest
scratch/pr_52/-sim-{tool}/reports/latest/report.html

util/dvsim/SimCfg.py
```diff
@@ -875,18 +875,18 @@ def create_bucket_report(buckets):
...
                     if hasattr(self, "cov_report_page"):
-                        results_str += "\n### [Coverage Dashboard]"
-                        if self.args.publish:
-                            cov_report_page_path = "cov_report"
-                        else:
-                            cov_report_page_path = self.cov_report_dir
-                        cov_report_page_path += "/" + self.cov_report_page
-                        results_str += "({})\n\n".format(cov_report_page_path)
+                        results_str += "\n### Coverage Dashboard\n\n"
+                        # if self.args.publish:
```
apinise, line 881: What is the purpose commenting out this block of code? Is it going to be added back once we are uploading to a real dashboard?

util/dvsim/FlowCfg.py
```diff
@@ -173,18 +172,13 @@ def _expand(self):
...
-        partial = self.is_primary_cfg
-
         # If custom dump script is exist, replace with run_script attribute.
         if self.args.dump_script is not None:
             self.run_script = '{proj_root}/' + self.args.dump_script
 
         self.__dict__ = find_and_substitute_wildcards(self.__dict__,
                                                       self.__dict__,
-                                                      self.ignored_wildcards,
-                                                      ignore_error=partial)
```
vlpannel, line 187: Assuming this will not affect other flow cfgs, right? What was the reason this change was made?

util/dvsim/SimCfg.py
```diff
@@ -911,28 +911,33 @@ def gen_results_summary(self):
...
+                rows.append(row)
 
-                # If header is set, ensure its the same for all cfgs.
-                if header:
-                    assert header == cfg.results_summary.keys()
-                else:
-                    header = cfg.results_summary.keys()
-                table.append(row.values())
+        if rows:
+            all_keys = list(dict.fromkeys(key for row in rows for key in row))
```
vlpannel, line 930:
```suggestion
            all_keys = list(key for row in rows for key in row.keys())
```
Nit: don't need intermediate function

util/dvsim/SimCfg.py
```diff
@@ -911,28 +911,33 @@ def gen_results_summary(self):
...
+            # Move the 5th key to the end
+            key_to_move = all_keys.pop(4)
+            all_keys.append(key_to_move)
```
vlpannel: I'm not too familiar with DV dashboards: what is the purpose of this?

util/dvsim/SimCfg.py
LouisTheLuis: That's the big question! Ideally, the tool can be used by anybody in such a way that whoever uses it with their corresponding tool at hand (VCS, Xcelium or QuestaSim) can have the Coverage Dashboard linked to their main report.

Problem is, for public view, we might not want these to be available. This might be resolved later with another commit where we distinguish local publishing (e.g. to any repo) from CI publishing (e.g. to our own dashboard).

util/dvsim/FlowCfg.py
LouisTheLuis: This fix was made so that you could run batch runs (e.g. `top_earlgrey_sim_cfgs.hjson`) and `dvsim` wouldn't crash over not being able to populate the `{tool}` section of the scratch directory. This would make it so that when you run something like `./util/dvsim/dvsim.py top_earlgrey_sim_cfgs.hjson --select-cfgs clkmgr`, it would generate a directory `scratch/main/top_earlgrey_batch-sim-{tool}/`.

I think if a wildcard is not defined, it should error out. Additionally, this encourages better practice as some batch files might include `xcelium` and `vcs` runs; and what happened when you didn't have both available is that some would run, some wouldn't, and then the final report would be incomplete (because it cannot fill coverage numbers for, say, Xcelium runs if you don't have Xcelium in the first place).

util/dvsim/SimCfg.py
LouisTheLuis: This is specifically to move the Overall Coverage column to the end. It's mostly a stylistic choice.

util/dvsim/SimCfg.py
LouisTheLuis: I think it's needed, the table generated below with:
```
table_txt = tabulate(table,
                    headers=all_keys,
                    tablefmt="pipe",
                    colalign=colalign)
```
will generate a column per element of the `all_keys` list. So we want to avoid duplication; otherwise the list becomes very long as it includes one set of keys per each `sim_cfg`.

util/dvsim/SimCfg.py
vlpannel:
```suggestion
            # Move the "Overall Coverage" key to the end
            all_keys.remove("Overall Coverage")
            all_keys.append("Overall Coverage")
```
Nit: slightly better readability but does the exact same thing in code.

util/dvsim/SimCfg.py
vlpannel: Ah, makes sense now. No need for change then. Using the Python `set` here would also work (also eliminates duplicates).

vlpannel (approved):

util/dvsim/SimCfg.py
LouisTheLuis: Done.

util/dvsim/SimCfg.py
LouisTheLuis: Added a TODO here.

hw/data/common_project_cfg.hjson
```diff
@@ -2,8 +2,8 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 {
-  project:          opentitan
-  repo_server:      "github.com/lowrisc/opentitan"
+  project:          pavona
+  repo_server:      "github.com/pavona/pavona-staging"
   book:             opentitan.org/book
```
matutem: Don't this and the following two lines need purging opentitan?

util/dvsim/FlowCfg.py
```diff
@@ -419,25 +418,27 @@ def gen_results(self, results):
                         if hasattr(item, '_gen_json_results')
                         else None)
             result = item._gen_results(results)
+            print("\n\n")
```
matutem: I think this doesn't belong here, also at line 430.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+        action="store_true",
+        help="Print what would be generated without writing files",
+    )
+    return parser.parse_args()
+
+
+def main():
+    os.chdir(SEARCH_ROOT)
+    args = parse_args()
+    if args.flow == "uvm_dv":
```
matutem: This flow is currently called "sim" (see SimCfg.py line 43), and more commonly "sim_dv". I don't think we need yet another name.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+    """Format a list of cfg paths as hjson use_cfgs entries."""
+    lines = []
+    pad = " " * indent
+    for cfg in cfgs:
+        lines.append(f'{pad}"{{proj_root}}/{cfg}",')
+    return "\n".join(lines)
+
+
+def generate_hjson(ip_cfgs, top_cfgs, tops, copyright_header, top_specific=False):
+    """Generate hjson for a batch sim_cfgs"""
```
matutem, line 109: How do we know we are not missing some _sim_cfg.hjson files? Perhaps we need an independent test(s) of this flow that performs some checks like this.

hw/dv/top_sim_cfgs.hjson
```diff
@@ -0,0 +1,111 @@
+// Copyright zeroRISC Inc.
```
matutem, line 1: Could we rename this file to something more indicative of what it is? A name like "all_sim_cfgs.hjson" may be more accurate?

hw/Makefile
```diff
@@ -89,6 +89,9 @@ $(ips_reg): %_reg:
 
 top_and_cmdgen: top cmdgen
 
-all: top cmdgen regs
+dvgen:
+	${PRJ_DIR}/util/gen_ip_collection.py --flow uvm_dv --tops top_earlgrey top_darjeeling
```
matutem: Change `uvm_dv` to `sim_dv` here?

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # Per-top hjson files
+    for top in tops:
+        out_path = f"hw/{top}/dv/{top}_sim_cfgs.hjson"
+        copyright_header = get_copyright_header(out_path, flow, top_cfgs)
+        content = generate_hjson(ip_cfgs, top_cfgs, [top], copyright_header, True)
+        write_file(out_path, content, dry_run)
+
+    # Global hjson file
+    out_path = "hw/dv/top_sim_cfgs.hjson"
```
matutem: Here we could rename it to something more indicative, as suggested in a separate comment.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def write_file(path, content, dry_run):
+    if dry_run:
+        print(f"\n{'=' * 60}")
+        print(f"Would write: {path}")
+        print('=' * 60)
+        print(content)
+    else:
+        os.makedirs(os.path.dirname(path), exist_ok=True)
```
matutem: Add try-except here and below.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+// PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED WITH THE FOLLOWING COMMAND:
+"""
+
+
+def get_copyright_header(path, flow, tops):
+    """Return the appropriate copyright header for this file."""
+    tops_str = ' '.join(tops)
+    GENCMD = (f"// util/gen_ip_collection.py --flow {flow}\n"
+              f"//                --tops {tops_str}\n")
+    if os.path.exists(path):
```
matutem, line 38: try-except here

util/gen_ip_collection.py
```diff
@@ -0,0 +1,252 @@
...
+            if skip(sim_name):
+                continue
+            top_cfgs[top].append(cfg)
+
+    return ip_cfgs, top_cfgs
+
+
+def format_use_cfgs(cfgs, indent=13):
+    """Format a list of cfg paths as hjson use_cfgs entries."""
+    lines = []
```
matutem: Use list comprehension for `lines`: it is way more pythonic!

hw/data/common_project_cfg.hjson
LouisTheLuis: Yes. The question is whether we should leave them blank or wait until we have an available webpage for them. I opened issue #34 regarding this. I will also put a TODO just in case.

util/gen_ip_collection.py
LouisTheLuis: As a matter of fact, there are some that are not being directly run (e.g. `spid` sim_cfgs). In line 51 we can see the patterns that are being looked at. Part of this effort is also to try to keep sim_cfgs in the right place; we don't want to run weeklies from `pre_dv` directories, for example.

util/gen_ip_collection.py
LouisTheLuis: Done.

hw/dv/top_sim_cfgs.hjson
LouisTheLuis: Good idea! Changed.

util/gen_ip_collection.py
LouisTheLuis: Done.

util/gen_ip_collection.py
LouisTheLuis: Done.

util/gen_ip_collection.py
LouisTheLuis: Done.

util/gen_ip_collection.py
LouisTheLuis: Done.

hw/Makefile
LouisTheLuis: Done.

util/dvsim/FlowCfg.py
LouisTheLuis: Done.

hw/data/common_project_cfg.hjson
matutem: Excellent, thanks.

util/gen_ip_collection.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python3
```
matutem, line 1: I think we may want to eventually transition this utility to be based on the complete top configs. I think those are the authoritative source for what goes in each top, including the paths. There are some IPs that have not yet been integrated in a top, and we could add them as special cases. No immediate action needed.

LouisTheLuis: I'm thinking the same thing... I just realized that for lints, the automatization over what goes in each `{topname}_lint_cfgs.hjson` is not as simple as importing files. I assume this might also be true of FPV cfgs.

matutem (approved):
LGTM

PR #53, opened 2026-03-27
[dvsim,dv] Include --publish flag and weekly regression

LouisTheLuis:
Now the `--publish` flag requires an argument corresponding to a repository to which `dvsim` collects the reports and `git push`es to. Note that it currently requires a `$SSH_KEY_PASSPHRASE` environment variable to be defined with the passphrase to push to a private repository (this is to be improved later on).

In addition to this, it includes a `weekly` argument that automatically mimics the `-i all`. Now, `nightly` runs the `-i all_once` with `-cov` enabled.

This PR is to be merged after #50, #51, and #52.

### DIFF
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -838,8 +838,8 @@ class SimCfg(FlowCfg):
 
                 # self.rel_path gives us the path to the directory
                 # containing the sim_cfg file...
-                testplan = "https://{}/{}".format(
-                    self.book,
+                testplan = "https://{}/tree/main/{}".format(
+                    self.repo_server,
                     Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.html"
                 )
 
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -840,7 +840,7 @@ class SimCfg(FlowCfg):
                 # containing the sim_cfg file...
                 testplan = "https://{}/tree/main/{}".format(
                     self.repo_server,
-                    Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.html"
+                    Path(self.rel_path).parent / 'data' / f"{self.name}_testplan.hjson"
                 )
 
             results_str += f"### [Testplan]({testplan})\n"
@@ -879,7 +879,7 @@ class SimCfg(FlowCfg):
                         if self.args.publish:
                             cov_report_page_path = "cov_report"
                         else:
-                            cov_report_page_path = self.cov_report_dir
+                            cov_report_page_path = "../../cov_report"
                         cov_report_page_path += "/" + self.cov_report_page
                         results_str += "({})\n\n".format(cov_report_page_path)
                     results_str += self.cov_report_deploy.cov_results
--- a/util/dvsim/SimCfg.py
+++ b/util/dvsim/SimCfg.py
@@ -21,7 +21,6 @@ from Deploy import CompileSim, CovAnalyze, CovMerge, CovReport, CovUnr, RunTest
 from FlowCfg import FlowCfg
 from modes import BuildMode, Mode, RunMode, find_mode
 from Regression import Regression
-from results_server import ResultsServer
 from SimResults import SimResults
 from tabulate import tabulate
 from Test import Test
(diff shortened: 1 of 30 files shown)


util/dvsim/doc/glossary.md
```diff
@@ -100,17 +109,36 @@ Please see [TBD]() for more details.
 
 DVSim provides these standard regression targets for all DUTs:
 - `smoke` (typically run in CI checks)
-- `nightly` (a full regression with coverage enabled, that is run every night)
+- `nightly` (a full regression with a single seed and coverage enabled, that is run every night)
+- `weekly` (a full regression with coverage enabled, that is run every night)
```
qmn:
```suggestion
- `weekly` (a full regression with coverage enabled, that is run every weekend)
```

util/dvsim/SimCfg.py
```diff
@@ -875,18 +874,18 @@ def create_bucket_report(buckets):
...
-                            cov_report_page_path = self.cov_report_dir
-                        cov_report_page_path += "/" + self.cov_report_page
-                        results_str += "({})\n\n".format(cov_report_page_path)
+                        results_str += "\n### Coverage Dashboard\n\n"
+                        # if self.args.publish:
+                        #     cov_report_page_path = "cov_report"
+                        # else:
+                        #     cov_report_page_path = "../../cov_report"
+                        # cov_report_page_path += "/" + self.cov_report_page
+                        # results_str += "({})\n\n".format(cov_report_page_path)
```
qmn, line 885: Suggest deleting these lines?

util/dvsim/SimCfg.py
```diff
@@ -911,28 +910,33 @@ def gen_results_summary(self):
...
 
-                # If header is set, ensure its the same for all cfgs.
-                if header:
-                    assert header == cfg.results_summary.keys()
-                else:
-                    header = cfg.results_summary.keys()
-                table.append(row.values())
+        if rows:
+            all_keys = list(dict.fromkeys(key for row in rows for key in row))
+            # Move the 5th key to the end
```
qmn: What's the 5th key such that we want to move it to the end?

util/gen_ip_collection.py
qmn, line 1: It is somewhat challenging to understand how gen_ip_collection.py works. From what I gather, it does the following:

1. Collect sim cfgs for block IPs as well as top-specific IPs and the top-level IP itself.
2. For each top, create a single batch sim cfg (see note below).
3. Create a global sim cfg.

Note for (2): I spent some time trying to understand the get_copyright_header() function. It appears to say: "if the top sim cfg originally had the joint copyright, then preserve it; otherwise, place the solo copyright; if there was no file to begin with, place the solo copyright."

I also see that there's a note to maintain alphabetical order but it appears they're sorted when they're being collected, not when they're being outputted.

Lastly, I think there's a lot of duplication in between generate_top_hjson and generate_global_hjson which could be factored out into one method.

qmn (commented):
Please have a look at the following comments.

util/dvsim/SimCfg.py
LouisTheLuis: I'm considering making it optional (CI vs. local dashboard).

util/dvsim/SimCfg.py
LouisTheLuis: Overall Coverage; it's a stylistic choice to have it after all the other coverage metrics rather than at the front.

util/dvsim/doc/glossary.md
LouisTheLuis: Done.

util/dvsim/SimCfg.py
LouisTheLuis: Added a TODO in #52.

util/gen_ip_collection.py
LouisTheLuis: (2) is indeed the rationale; the idea is to not override the original copyright header. It can be improved once we decide how to proceed with copyright.

For the maintaining alphabetical order, I think I will get rid of it; it gets sorted by top anyways. Made fixes over the duplication in #51.

util/dvsim/SimCfg.py
LouisTheLuis: Fixed as noted in #52.

util/dvsim/SimCfg.py
```diff
@@ -875,18 +874,20 @@ def create_bucket_report(buckets):
...
-                        results_str += "\n### [Coverage Dashboard]"
-                        if self.args.publish:
-                            cov_report_page_path = "cov_report"
-                        else:
-                            cov_report_page_path = self.cov_report_dir
-                        cov_report_page_path += "/" + self.cov_report_page
-                        results_str += "({})\n\n".format(cov_report_page_path)
+                        results_str += "\n### Coverage Dashboard\n\n"
+                        # TODO: modify to include differentiation between public and private
+                        # dashboard usage.
```
qmn, line 879: Looks good for now, maybe this can be an arg you can set later.

qmn (approved):
LGTM, thanks for your improvements to DVSim!

PR #54, opened 2026-03-27
[gitless] Make the repo usable without git refs

vlpannel:
This repo should be mostly usable even if taken from a [git archive](https://git-scm.com/docs/git-archive) and/or if ported into a version control system that is not git.

This means that anything that unnecessarily relies on git refs should be revised. Of course, not all the functionalities of this repo (such as checking diffs) can be used without git history in place, but a number of them can. It is important to note that git is currently listed as one of the repository installation requirements, so some things (like patching) should still work without the repository itself being a git repo.

This PR replaces unnecessary uses of calls to git (such as using `git rev-parse` to find the repo top or creating paths based on the current git branch name) with some appropriate default behavior whenever unable to use git.

It also creates a "sanity check" testing script to ensure that the "git-less" version of the repo should work about as well as the "git-ful" one: `ci/scripts/check-gitless.sh` should pass for both a cloned version of the repo and a version taken from a `git archive` tarball.

### DIFF
--- a/.github/workflows/gitless.yml
+++ b/.github/workflows/gitless.yml
@@ -1,0 +1,37 @@
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Gitless repo test
+on:
+  workflow_call:
+
+jobs:
+  gitless:
+    name: Run gitless check
+    runs-on:
+      group: pavona-large
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
+        with:
+          fetch-depth: 0
+      - name: Prepare environment
+        uses: ./.github/actions/prepare-env
+      - name: Work from git repo archive
+        run: |
+          git archive --format=tar.gz -o "../gitless.tar.gz" HEAD
+          cd ..
+          mkdir gitless
+          tar -xzf gitless.tar.gz -C gitless --strip-components=1
+          cd gitless
+          export REPO_TOP=$(realpath .)
+      - name: Run gitless check
+        continue-on-error: true
+        run: ./ci/scripts/check-gitless.sh -e
+      - name: Output gitless log
+        run: |
+          log_output=$(cat gitless.log)  # will be empty if no errors
+          if [ -n "$log_output" ]; then
+            echo $log_output
+            exit 1
+          fi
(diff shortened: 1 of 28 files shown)


util/topgen/templates/toplevel.rs.tpl
```diff
@@ -47,6 +47,7 @@ ${helper.file_header.render()}
 // To regenerate this file follow OpenTitan topgen documentations.
 
 #![allow(dead_code)]
+#![allow(unused_imports)]
```
rtorok-zr: What warning were you getting from `rustc` that made this `allow` necessary? There may be a better solution.

vlpannel: This is a good question; this was an attempt to fix an earlier CI failure, but it may have already been fixed on `main`.  I'll try to remove this commit and see if the problem has already been fixed elsewhere, and if not, will try something else.

vlpannel: Okay, this was unnecessary and the commit has been dropped.

.github/workflows/gitless.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  workflow_call:
+
+jobs:
+  gitless:
+    name: Run gitless check
+    runs-on:
+      group: pavona-large
+    steps:
+      - uses: actions/checkout@v4
```
qmn: Please pin to SHA.

qmn (commented):
Hi, this looks great! This should make it easier to cut releases, and will be useful for those who have to consume the repository as a tarball and not a Git repo.

Should the second commit, [[sim_cfg,doc] Fix batch sim_cfg names in docs and code](https://github.com/pavona/pavona-staging/pull/54/changes/6fdee93c5b33f40a665d3337d5164e2ba4d3f37f), be hoisted into a separate PR?

I have only one minor comment so far.

vlpannel:
> Should the second commit, [[sim_cfg,doc] Fix batch sim_cfg names in docs and code](https://github.com/pavona/pavona-staging/pull/54/changes/6fdee93c5b33f40a665d3337d5164e2ba4d3f37f), be hoisted into a separate PR?

Yes, I think this is appropriate; commit was originally made to fix link check errors in CI, but will move over to a separate PR.

qmn (approved):
After squashing the pinning SHA commit and also eliding the CONTRIBUTING.md fix from this PR, this LGTM. Thanks for supporting this important release flow!

PR #59, opened 2026-03-30
[rfc] Add versioning RFC

vlpannel:
This PR is for an RFC regarding versioning--version scheme/naming as well as a release development schedule are discussed.

The RFC proposes:
* Date-based versioning implemented via git tags and release branches
* A controlled release branch development cycle
* Versioning the repo as a whole as opposed to individual parts

This RFC includes a large amount of additional information to provide context/rationale for date-based versioning and versioning the repo as a whole. The "Supplementary" section at the end of the document is to be considered as helpful but not essential reading.


rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
+# Versioning RFC
+
+* Status: Proposed
+* Authors: [Viveca Pannell](mailto:vpannell@zerorisc.com), [Quan Nguyen](mailto:qmn@zerorisc.com)
+* Last updated: Dec 19, 2025
+
+## Introduction
+
+This RFC covers basic repo versioning–namely, which schemes to use and how to implement it.
```
adurbin-rivos: What does 'versioning-namely' mean? I'm not sure I understand the sentence.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+The repository needs versioning to help its users reason about the quality and compatibility of its parts, especially as they change over time.
+We propose to track all parts of the repo with only one version number.
+We will not use semantic versioning; we will use calendar versioning (`YYYY.MM`) to coincide with major repo releases, with revision numbers to represent subsequent patches.
```
adurbin-rivos: How is the revision number conveyed in the encoding scheme? I see it's mentioned somewhat below... May want to start with the anticipated scheme in its entirety.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+### Proposal
+
+![Summary of versioning](./overview.svg "Summary of versioning")
+
+The purpose of versioning is to keep track of development progress and give users a common baseline point at which to most easily use the repo tools at hand.
+
+We will version the repo as a whole.
+We use a release-date-plus-patch-number versioning scheme of `[YYYY].[MM].p[#]`.
```
adurbin-rivos: Above 'revision numbers' are mentioned, but this talks about patch numbers. I'm a confused what is what. Moreover, is there an implicit 'p0' do we unconditionally add patch numbers?

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+Other docs may override or require revision of what is written here, including future RFCs.
+
+## Versioning Scheme
+
+*Supplementary topics: [monolithic repo rationale](#keeping-the-repo-as-a-single-whole), [repo split considerations](#possible-repo-segmentation:-device-side-and-host-side-code), [calver rationale](#why-calver/calendar-based-versioning?), [FuseSoC side note](#calver-with-fusesoc), [patch number rationale](#advantages-to-patch-number-for-any-size-of-change), [files which track versions](#hardware-files-which-track-versions), [how big of a change](#how-big-of-a-change-warrants-a-new-version?)*
+
+A simple versioning scheme can go a long way: it can be both easy to implement and provide a quick way to supply information.
+We use a calendar versioning (“calver”) scheme of “*YYYY*.*MM*.p*\#*”, filled in with the year and month of a given major repository release, plus a patch number.
+
+Two items that have the same calver date will be considered compatible, corresponding to the same repo release.
```
adurbin-rivos: The use of 'items' here seems a little confusing. Isn't the only thing that applies as an item in this context the tag with the same date?

rfc/rfc-2026-XX-Versioning/rfc_2026_XX_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+## Versioning Scheme
+
+*Supplementary topics: [monolithic repo rationale](#keeping-the-repo-as-a-single-whole), [repo split considerations](#possible-repo-segmentation:-device-side-and-host-side-code), [calver rationale](#why-calver/calendar-based-versioning?), [FuseSoC side note](#calver-with-fusesoc), [patch number rationale](#advantages-to-patch-number-for-any-size-of-change), [files which track versions](#hardware-files-which-track-versions), [how big of a change](#how-big-of-a-change-warrants-a-new-version?)*
+
+A simple versioning scheme can go a long way: it can be both easy to implement and provide a quick way to supply information.
+We use a calendar versioning (“calver”) scheme of “*YYYY*.*MM*.p*\#*”, filled in with the year and month of a given major repository release, plus a patch number.
+
+Two items that have the same calver date will be considered compatible, corresponding to the same repo release.
+
+This version number provides information on: 1\) the version being stable or deprecated, and 2\) how many bug fixes it has received.
```
adurbin-rivos: How does the version encoding indicate stability or deprecation on its own? On the bug fixes comment is it implied that there's a new tag for every bug fix?

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+![Example of release branch evolution; initial release must occur in February 2026](./development.svg "Example of release branch evolution; initial release must occur in February 2026")
+
+In accordance with good practices, development should take place on separate branches, not directly on master or on any release branch.
+Changes can then be merged into or rebased onto these central branches after making a pull request (PR) and getting reviewer approval.
+
+An important thing to note: release numbers never change nor decrement.
+After the initial release, any updates on a release branch *must* constitute a new patch, and previous patches should never be amended or rescinded (a new patch may undo some of the things changed in a previous patch, but this does not change the git history nor patch numbering).
+
+Once a suitable point on master on master is selected as the start of the new release, that commit should be tagged with `cut:[YYYY].[MM]` to designate the point at which the release branch will diverge.
```
adurbin-rivos: 'on master on master' -> remove one of those?

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+Before initial release, additional commits may be made on the release branch (most likely cherry-picked from master) to prepare a minimally buggy, well-tested, and well-documented release.
+Master is considered unstable and where the majority of development is pushed, while releases are designed to be more carefully-controlled versions of the repo.
```
adurbin-rivos: Above we have the following:

```
We’d like to avoid:
- Reliance on an unstable tip of master
```

These two statements seem to be in contradiction.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+As releases progress from in-development to being public and stable, the release branch will finally be tagged with `release/[YYYY].[MM].p0`.
```
adurbin-rivos: Since we're using 'p0' here:
- Presumably we're making a new branch with a new date code as well.
- And then we use the tagging scheme?
- What happens to older dated branch that was being used to stabilize it?


If I'm understanding correctly, we have a 'cut:[YYYY].[MM]' that corresponds to a 'release/[YYYY].[MM]'  branch of the same date. Patches are applied on that branch subsequently making new 'release/[YYYY].[MM].p#' tags which ultimately will result in a new branch? But we can't go back to 'p0' of the same date because that would conflict with the stabilization tags.

I'm thinking we may want a variant of an a release that is '[YYYY].[MM].r#' that signifies a release vs patch.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+All development for new features and improvements should instead go to master, and release branches may, if desired, backport select improvements from master up until the initial release.
+
+Generally speaking, changes on master should only be incorporated into the release branch before the initial release, at which point the two branches are considered to have diverged.
+If the same bug is found on both branches post-release, two separate PRs–one against each branch–is warranted; if there are no conflicts, it may be feasible to cherry-pick commits from master.
+After initial release, any sets of commits cherry-picked to a release branch should constitute a new patch as well.
+
+### Schedule
+
+Repo releases should come out on a consistent schedule, always on time for the version number’s release date–once a release date is selected, the initial release must be made during that month, cutting out any unfinished features as necessary.
+We will start with twice per year releases and adjust this frequency as necessary.
```
adurbin-rivos: s/We/The project'/ or 's/We/The Pavona project/'

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+A regular, predictable release schedule chosen and well-communicated ahead of time will help instill trust in the repo; therefore, it may be advantageous to select a specific cadence (eg., "every March and every September") rather than individual release months.
```
adurbin-rivos: 'well-communicated ahead of time' -- missing a noun? Or rejigger the sentence?

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+1. Select a suitable commit; tag it as `cut:[YYYY].[MM]`.
+
+2. Create the release branch `release/[YYYY].[MM]`, starting at that commit.
+
+3. Before the full release is ready, ensure quality of new initial release and publish metrics.
+   1. Run suite of tests to determine baseline functionality of repo and post test results: system-level tests, block-level test, coverage tests. Publish results.
+   2. Update user instructions and documentation in repo: for running tests, for extracting RTL from FuseSoC, top level block diagrams, purposes of top level signals, memory maps, for building hardware and software, etc.
+   3. Fix any bugs found along the way.
+
+4. At initial release, tag the latest commit on the release branch as `release/[YYYY].[MM].p[P]`.
```
adurbin-rivos: Previous descriptions of the encoding used '#' instead of 'P'.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+5. Create a new release package on GitHub (see below).
+
+### Pull Requests for Patches
+
+Contributors’ Checklist:
+
+1. Update metadata/version-concerned files with new version numbers as needed.
+2. Make PR on GitHub to bring changes into release branch. Follow standard contribution procedures.
+   1. Specify significant changes introduced by this patch in PR comments.
+3. Once PR is merged, ensure the latest release branch commit is tagged with the new patch number.
```
adurbin-rivos: Why would a contributor be responsible for creating the tags? I don't think anyone should be able to create tags.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+- [ ] Is this patch being made before the version’s end-of-life?
+
+### Create Release Package
+
+\[Still to be tested\]
+
+1. Clone/checkout desired release. Make sure it is up-to-date with origin/the shared copy of the repo.
+2. Write release notes (mandatory for initial releases, optional for patches; Markdown format allowed) somewhere outside of the repo.
+3. Tar the repo, git history and all (GitHub release packaging will provide a default source code tarball and .zip file without git history in addition to this).
+4. Create a new repo release on GitHub titled “[YYYY].[MM].p[P]” from the `release/[YYYY].[MM].p[P]` git tag, copying release notes into the description box and uploading the with-git-history tarball.
```
adurbin-rivos: Same comment as above relating to '#' vs 'P'.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+Additionally, especially for hardware assets, developers should always err on the side of making things more parametrizable–especially when development makes certain design decisions that are not inarguably improvements.
+
+Host-side code should all work together, being like a “toolbox” full of tools that are all compatible with one another.
+Because everything on the host-side should work off the same set of axioms, it is expected that host side code only fully updates at the time of repo release, with new releases being opportunities for new tooling schemes.
+
+Device-side code will use a revision number instead of a patch number.
+This way, individual components may be updated without having to update the versioning of every component.
+In other words, it will be easier to pull together components from different points in time if so desired, which is especially important given that components may have varying rates of development over time.
+
+For each device-side component, whenever the component is updated, the commit that merges those changes into the release branch should also be tagged as `<component>@[YYYY].[MM].r[#]`.
```
adurbin-rivos: Since this section was about a topic of discussion that was ultimately discarded, perhaps we shouldn't provide more details on the implementation? It may make for some consusion.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+The main warning from detractors of calver (aside from backwards compatibility in projects that have been using another versioning scheme for a long time) is that calver must still mean *something* about the code other than just a datetime.
+In our case, it is a version of the repo that is usable as a starting point: minimal bugs, well-coordinated parts, adequate documentation, and published test results at a fixed point in time.
+
+Changes between versions should be clearly communicated (especially “breaking” ones that will make integrating previous version parts more difficult), and releases should always be made on a meaningful (predictable, consistent, appropriate for the speed of development) schedule.
+
+### Calver with FuseSoC {#calver-with-fusesoc}
+
+An important note about calver and FuseSoC: FuseSoC uses the okonomiyaki Python library to determine versions, which assumes semver.
+This can [pose a problem](https://github.com/olofk/fusesoc/issues/565) when calver versions are used in FuseSoC core files, so instead we will only use the patch number as the version number in hardware core files.
```
adurbin-rivos: I think we should elaborate on the mechanism of how patch number is only used. There's a potential order of operations of tagging and updating core files in addition to the specifics of how one encodes the proper version on fusesoc files.

The other thing that comes to mind: if someone checks out 2 different tags. How does someone distinguish between the version of an IP? Patch numbers reset at 0 for each branch.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+
+For each new patch, the version number on every FuseSoC core file should accordingly increment/match the patch number.
+
+### Advantages to patch number for any size of change {#advantages-to-patch-number-for-any-size-of-change}
+
+While patch number does not specify the size of any given update, it’s more important to know just when anything changes.
+Trying to qualify a change may prove difficult, and git logs should be adequate for examining specific revisions.
+
+Why not just use commit hashes along the release branch instead of patch number, keeping only the versioning at the latest point in the release branch?
+Patch number inherently has an order and can differentiate between pre-release status on the branch and public release; with commit hashes, it’s unclear when one hash precedes the other, and there’s no way to tell at what point the release became stable.
```
adurbin-rivos: I'm not sure the current scheme differentiates a pre-release vs a public release.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -0,0 +1,472 @@
...
+Patch number is also easier to remember than commit hashes.
+
+### Hardware files which track versions {#hardware-files-which-track-versions}
+
+Correct version information should also be maintained in version-tracking hardware files like HJSON metadata and FuseSoC core files.
+HJSONs should clearly note the full version number of the repo (release date and patch number), but see the above note about FuseSoC version numbering.
```
adurbin-rivos: How do we update all the hjson files with the anticipated metadata? That's a separate step needed for every tag generation.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -85,7 +87,8 @@ Changes can then be merged into or rebased onto these central branches after mak
 An important thing to note: release numbers never change nor decrement.
 After the initial release, any updates on a release branch *must* constitute a new patch, and previous patches should never be amended or rescinded (a new patch may undo some of the things changed in a previous patch, but this does not change the git history nor patch numbering).
 
-Once a suitable point on master on master is selected as the start of the new release, that commit should be tagged with `cut:[YYYY].[MM]` to designate the point at which the release branch will diverge.
+Once a suitable point on master on master is selected as the start of the new release, that commit should be tagged with `cut/[YYYY].[MM]` to designate the point at which the release branch will diverge.
+The year and month still remain the release year and month, not the date when the release branch is created.
```
adurbin-rivos: I'm not sure why this sentence is necessary. It seems superfluous, unless you just want to explicitly indicate there's a release branch being made with the intent of a release being marked '[YYYY].'[MM]'.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -428,11 +441,11 @@ Differences between release branches will be small enough that users will have e
 
 ### GitHub alternatives consideration
 
-Just in case we need to switch over to another platform (hint: [recent news](https://github.blog/changelog/2025-12-16-coming-soon-simpler-pricing-and-a-better-experience-for-github-actions/) may affect CI costs), it’s useful to consider any alterations with the versioning and release process that would need to be made, were we to switch to [another code hosting platform](https://en.wikipedia.org/wiki/Comparison_of_source-code-hosting_facilities#Version_control_systems).
+Just in case we ever need to move away from GitHub (this was prompted by [recent news](https://github.blog/changelog/2025-12-16-coming-soon-simpler-pricing-and-a-better-experience-for-github-actions/) that may affect CI costs), it’s useful to consider any alterations with the versioning and release process that would need to be made, were we to switch to [another code hosting platform](https://en.wikipedia.org/wiki/Comparison_of_source-code-hosting_facilities#Version_control_systems).
```
adurbin-rivos: Use of 'we' again. I suggest making it more formal using 'project' or 'Pavona project'.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -427,7 +427,7 @@ Git tags have been selected as the central mode of storing version information f
   * Can be signed and verified.
 
 Git tags have some limitations that can be worked around, especially with the fact that git tags are commit-specific.
-Since rebasing or amending commits do not retain git tags (these actions create new commits), we will enforce a no-rebase and no-amend policy in the upstream repo: once changes are brought in to the upstream release branch, they are final.
+Since rebasing or amending commits do not retain git tags (these actions create new commits), we will enforce a no-rebase and no-amend policy on the release branch: once changes are brought in to it, they are final.
```
adurbin-rivos: Use of 'we' again.

rfc/rfc-2026-02-Versioning/rfc_2026_02_versioning.md
```diff
@@ -309,7 +312,7 @@ It does not cover:
 ### Keeping the repo as a single whole
 
 Original ideations about repo versioning treated version numbers as potentially being assigned to various parts within the repo instead.
-We examined trying to version various levels of hardware (prims, IP blocks, and tops) in depth, then considered the below-described division of the repo into host and device.
+We examined trying to version various levels of hardware (prims, IP blocks, and tops) in depth, then considered division of the repo into host and device.
```
adurbin-rivos, line 300: 'we'

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -414,6 +345,13 @@ We currently have no way of quantifying the size of a change, and instead of was
 The only changes that could possibly be excluded here are minor typo fixes in comments and documentation or code reorganization for simple readability.
 These changes do not affect repo usage and can easily be lumped in with the next set of functional changes.
 
+While patch number does not specify the size of any given update, it’s more important to know just when anything changes.
```
adurbin-rivos: s/just//

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
```diff
@@ -105,7 +109,7 @@ After initial release, any sets of commits made to a release branch should const
 
 A regular, predictable release schedule chosen and well-communicated ahead of time will help instill trust in the repo; therefore, it is advantageous to select a specific cadence rather than individual release months.
 
-Repo releases should always come out on time–once a release date is selected, the initial release must be made during that month, rolling back any unfinished features as necessary.
+Repo releases should always come out on time–once a release date is selected, the initial release *must* be made during that month, rolling back any unfinished features as necessary.
```
adurbin-rivos: s/time-once/time. Once/

adurbin-rivos (commented):
My apologies for some of the comments if they don't apply in the progression. When I review changes I walk all the commits individually. I think we need to look at the completed whole for consistency as I suspect there may be some inconsistencies lurking.

General question: when can we squash these commits? Or we can't because GitHub is annoying about dropping comments? It'd be easier to digest as one change. I would expect there to be previous versions of a PR, but I suspect GitHub drops that which makes it harder?

vlpannel:
> General question: when can we squash these commits? Or we can't because GitHub is annoying about dropping comments? It'd be easier to digest as one change. I would expect there to be previous versions of a PR, but I suspect GitHub drops that which makes it harder?

Completely agree about squashing these commits--I was originally hesitant to do so given RFCs should preserve the revision history of the document, but there are no previous discussions on this subject to preserve, so this history of revision isn't important.

Apologies that the git history made this more difficult to review; the commits have been squashed now (though I think future commits likely won't since they'll correspond to discussion on this PR thread).

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
vlpannel: This is supposed to be an em-dash character. As this may cause confusion, I'll swap it with "--".

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
vlpannel: Right; "revision number" was part of a previous idea and should be replaced only with the "patch number" terminology now.

"p0" is not implicit; a release will be manually tagged with the initial patch.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
vlpannel: The idea is to allow development along master (unstable), then to periodically cut one release branch at a given date and stabilize it.

The release branch can then be something that we point new/non-contributing users to: it will have minimal bugs and doesn't receive much modification, so they can happily use that release for an extended period of time without having to worry about incoming changes.

I guess "reliance" here means solely using the repo at the tip of tree; we want other endorsed versions of the repo for non-contributors.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
vlpannel: The release branch should not be patch specific. Rather, all the patch tags are at different points along the release branch.

The only commit that should both have a tag and a new branch cut from it is the `cut:[YYYY].[MM]` tag.

* A branch that is not ready yet will simply have no `p0` tag on it (pre-release).
* Once it is ready, it will get a `p0`.
* After patch 0 has been established, if there is a bug that needs a fix, someone can push a fix to the release branch and tag the last commit of that fix with `release/[YYYY].[MM].p1`.
* Another bug fix will similarly be pushed to the same release branch and be tagged as `p2`.
* etc.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
vlpannel: We considered creating automation for tagging, but at the time, we thought we wouldn't have enough time before repo launch to do so. Automation might be feasible now, but is there any other option for tagging to consider considered?

adurbin-rivos:
> > General question: when can we squash these commits? Or we can't because GitHub is annoying about dropping comments? It'd be easier to digest as one change. I would expect there to be previous versions of a PR, but I suspect GitHub drops that which makes it harder?
> 
> Completely agree about squashing these commits--I was originally hesitant to do so given RFCs should preserve the revision history of the document, but there are no previous discussions on this subject to preserve, so this history of revision isn't important.
> 
> Apologies that the git history made this more difficult to review; the commits have been squashed now (though I think future commits likely won't since they'll correspond to discussion on this PR thread).

@vlpannel could you please let me know when you want me to re-review? I didn't know if you just squashed or made other changes as well.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
adurbin-rivos: Apologies that my motivation wasn't clear. The wording in the commit says "will finally be tagged with...p0" in order to make it public. That's not exactly the flow described. I think the sentence needs to be rewritten or more elaboration is required.

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
adurbin-rivos: I understand the intent; my commentary is that the description indicates not wanting to rely on an unstable tip of master but then here it's stated that master is is considered unstable. I think we should align the language to make the thoughts more consistent between here and there. Or omit some of the explicitness because it is sort of implied that 'main' is the development branch. Stability there is reliant on the developers and processes, obviously, but release branches are absolutely expected to be stable.

For that matter, 'master' should be 'main', no?

rfc/rfc-2025-03-Versioning/rfc_2025_03_versioning.md
adurbin-rivos: My commentary was centered around someone may be cherry-picking fixes over, but I was expecting someone else may be coordinating the release, etc. If there's more than one fix in-flight then ordering matters as well which makes me think whoever is in charge of the release does the tagging -- not the person applying fixes.

I think I have another comment about the fusesoc stuff around patches as well that applies to such a situation as well. If you want patch numbers of releases to match with the tag there's a whole other stuff aligning the fusesoc core files too which lends itself to the the person coordinating the release to perform.

vlpannel:
I've made a few changes based on PR feedback and, after re-reading the RFC, intend to reorganize the document a little bit (structure is a bit confusing).

rfc/rfc-2026-XX-Versioning/rfc_2026_XX_versioning.md
```diff
@@ -6,11 +6,11 @@
...
 ## Introduction
 
-This RFC covers basic repo versioning–namely, which schemes to use and how to implement it.
+This RFC covers basic repo versioning--namely, which schemes to use and how to implement it.
 
 The repository needs versioning to help its users reason about the quality and compatibility of its parts, especially as they change over time.
-We propose to track all parts of the repo with only one version number.
-We will not use semantic versioning; we will use calendar versioning (`YYYY.MM`) to coincide with major repo releases, with revision numbers to represent subsequent patches.
+The Pavona project will track all parts of the repo with only one version number.
+Instead of semantic versioning, it will use calendar versioning (`YYYY.MM`) to coincide with major repo releases, with revision numbers to represent subsequent patches.
```
adurbin-rivos: Sorry to quibble: if I'm not mistaken 'revision number' used here is the only place in the doc. Maybe I'm over indexing, but I think we should omit that noun usage and stick to describing the scheme (patch numbers).

vlpannel: I think I removed the term somewhere else in the doc but missed this instance. Thanks!

adurbin-rivos (approved):
I think this is a good baseline. We can tweak more once it lands and people find anything they want to change, imo.

adurbin-rivos (approved):

PR #61, opened 2026-03-30
[ac_range_check,sv] Fix missing reset safe files

LouisTheLuis:
Some files were lost during merge. Now sequences are passing again.


qmn (approved):
Thanks for finding these, LGTM

matutem (approved):
Approved since the files went missing during repo transfers, so the changes were previously approved.

PR #62, opened 2026-03-30
[verilator] Build verilator in-tree using Bazel

rtorok-zr:
This PR removes the dependency on a nonhermetic verilator installation on the host, replacing it with a Bazel repository dependency in `third_party/verilator`.

The verilator binary can now be run hermetically within the repo using `bazel run @verilator//:verilator`. Test targets using a `sim_verilator` environment automatically use the hermetic target thanks to an update to `rules/fusesoc.bzl`.

This PR also bumps the verilator version from v4.210 to v5.046.

Dependent on #88

### DIFF
--- a/third_party/verilator/BUILD.verilator.bazel
+++ b/third_party/verilator/BUILD.verilator.bazel
@@ -1,0 +1,11 @@
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+exports_files(glob(["**"]))
+
+filegroup(
+    name = "all_srcs",
+    srcs = glob(["**"]),
+    visibility = ["//visibility:public"],
+)
--- a/util/fusesoc_build.py
+++ b/util/fusesoc_build.py
@@ -18,6 +18,13 @@ if __name__ == "__main__":
     if path_env is not None:
         path_env = ":" + path_env
     path_env = os.path.dirname(sys.executable) + path_env
+
+    # VERILATOR_BINARY` is passed in as a relative path to the Bazel
+    # `execroot`. Convert this to an absolute path in PATH, if present.
+    if "VERILATOR_BINARY" in os.environ:
+        verilator_binary = os.environ["VERILATOR_BINARY"]
+        verilator_dirname = os.path.dirname(os.path.join(os.getcwd(), verilator_binary))
+        path_env += os.pathsep + verilator_dirname
     os.environ["PATH"] = path_env
 
     # Start fusesoc
--- a/util/fusesoc_build.py
+++ b/util/fusesoc_build.py
@@ -7,9 +7,11 @@ is used for generators.
 """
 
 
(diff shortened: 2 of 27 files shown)


third_party/verilator/BUILD.verilator.bazel
```diff
@@ -0,0 +1,50 @@
...
+# is executable, Bazel does not believe it is runnable. If you want to run
+# verilator , use the runnable wrapper like so: `bazel run @verilator//:verilator`.
+filegroup(
+    name = "verilator_bin",
+    srcs = [":build_verilator"],
+    output_group = "verilator_bin",
+    visibility = ["//visibility:public"],
+)
+
+genrule(
```
rtorok-zr: This pretty blatantly copies what lowRISC did for `third_party/openocd`. Newer versions of `rules_foreign_cc` than the v0.9.0 in this repo have a `runnable_binary` target, but I was not able to get it working in a way that did not rebuild all of Verilator from source every time I ran it. Somehow the caching was being messed up.

matutem: Does this mean this WILL always rebuild verilator from source? That would be pretty bad news. Could you explain?

util/fusesoc_build.py
```diff
@@ -18,6 +18,13 @@
     if path_env is not None:
         path_env = ":" + path_env
     path_env = os.path.dirname(sys.executable) + path_env
+
+    # VERILATOR_BINARY` is passed in as a relative path to the Bazel
+    # `execroot`. Convert this to an absolute path in PATH, if present.
+    verilator_binary = os.environ["VERILATOR_BINARY"]
+    if verilator_binary is not None:
+        verilator_dirname = os.path.dirname(os.getcwd() + "/" + verilator_binary)
```
qmn:
```suggestion
        verilator_dirname = os.path.dirname(os.path.join(os.getcwd(), verilator_binary))
```

util/fusesoc_build.py
```diff
@@ -18,6 +18,13 @@
...
+        path_env += ":" + verilator_dirname
```
qmn:
```suggestion
        path_env += os.pathsep + verilator_dirname
```

(not to be confused with os.sep = `/`)

qmn (commented):
Very impressive, thanks so much! Is it required that we combine this with the bump to v5.046? Would it be possible to parcel this out to two commits: 1) building Verilator in-tree and 2) the version bump (which should include the `--no-timing` flag changes)?

third_party/verilator/BUILD.verilator.bazel
rtorok-zr: The alternative solution I considered was doing that, but doing it this way does cache properly after it is built once.

rtorok-zr: This is outdated. Marking as resolved.

qmn (approved):
Awesome! Thanks so much for this work. It'll be much easier to get started on the project now. LGTM

PR #67, opened 2026-04-01
[otp_macro,rtl] Clear rdata before zeroize read

### LINKED ISSUE #63: [otp_ctrl,rtl] Fix count-ones to consider the granule size
### Description

The zeroize command should return the count of ones in the item that was zeroized to determine if zeroization was successful. The implementation in otp_ctrl_dai.sv simply counts the number of ones in the data returned by the macro.

The otp_macro returns 64-bit for each read and only updates the word that was read, so in case of 32 bit reads the higher bits are unmodified. For dai reads the result is placed in the direct_access_rdata_0 and 1: 0 for 32 bit reads and both for 64 bit reads.

Zeroization computes the number of ones in the otp_macro read data and places them only in direct_access_rdata_0.

For 64-bit granule accesses the result is fine, but for 32-bit granule the top 32 bits are whatever was left from the most recent read. This is problematic since it is not possible to determine the number of bits that were actually set.

### Impact

There is a relatively simple work-around since the recommended flow for zeroization for any partition is to first zeroize the zeroization mark, which is 64 bits, and then zeroize the individual fields. If there are no intervening read commands this means when zeroizing items of granule 32 the top 32 bits are going to be all 1's left from zeroizing the mark (except for any stuck at 0 bits), so it is possible to put a bound on the number of 1's for subsequent 32-bit granule zeroizations: you can simply subtract 32 from the count returned by the zeroize command for items of granule 32.

I expect the zeroization software will need to account for this behavior.

### How to fix it

It would be possible to fix this in either otp_ctrl_dai.sv or in otp_macro.sv. I think it is best to fix it in otp_macro.sv.

matutem:
Fixes #63

### DIFF
--- a/hw/ip/otp_macro/rtl/otp_macro.sv
+++ b/hw/ip/otp_macro/rtl/otp_macro.sv
@@ -226,6 +226,7 @@ module otp_macro
   logic valid_d, valid_q;
   logic integrity_en_d, integrity_en_q;
   logic req, wren, rvalid;
+  logic clear_rdata;
   logic [1:0] rerror;
   otp_macro_addr_t addr_q;
   logic [SizeWidth-1:0] size_q;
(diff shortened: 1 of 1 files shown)


LouisTheLuis (approved):
LGTM.

PR #68, opened 2026-04-01
[kmac,rtl] Made the ACC connection agnostic to app index

apinise:
Porting PR from https://github.com/zerorisc/expo/pull/216

This PR will fix issues with potential top-level configurations that have multiple `ACC` connections. It replaces some logic in `kmac_app.sv` to act based on the current app mode rather than just the single `ACC` connection. It also updated the hash modes in `kmac.py` model in `ACC` to match the encoding used by verilog packages. This fixes an issue with using `accsim` and a real `KMAC` connection to `ACC`.


qmn (approved):
LGTM, thanks!

PR #71, opened 2026-04-01
[spi_host,dv] Work-around for simulator defects and clean up

matutem:
This has 4 commits:
- Fix the size of a bin in command_cg covergroup: this is just to avoid a warning and should have no side-effects since the simulator resized the array as expected
- Fix the csb size: it was sized with the wrong parameter, but won't affect things since this block uses a single external endpoint
- Fix SV format of covergroups, no functional change
- Add workaround for simulator defects: this is a real change in DV. Without it the spi_host_passthrough_mode test was consistently failing


LouisTheLuis (approved):
LGTM.

PR #75, opened 2026-04-02
[dvsim] Include batchgroups option in sim_cfgs

LouisTheLuis:
With this commit, now `sim_cfgs` can have their `use_cfgs` be divided in several groups for more clear reporting.
Additionally, `--publish-prev` and `--cov_results` flag are introduced for publishing already existing `latest/` reports and for including tool-specific coverage reports in the `report.html`, respectively.
The directories are also renamed (this allows for users to publish lint runs and formal runs, if desired) and a bug for running batch `sim_cfgs` is fixed.

Builds on top of #53.

### DIFF
--- a/util/dvsim/CfgJson.py
+++ b/util/dvsim/CfgJson.py
@@ -89,8 +89,8 @@ def _load_single_file(target, path, is_first, arg_keys):
                 raise RuntimeError('{!r}: File is included by another one, '
                                    'but defines "use_cfgs".'
                                    .format(path))
-            if not isinstance(dict_val, list):
-                raise RuntimeError('{!r}: use_cfgs must be a list. Saw {!r}.'
+            if not isinstance(dict_val, list) and not isinstance(dict_val, dict):
+                raise RuntimeError('{!r}: use_cfgs must be a list or a dict. Saw {!r}.'
                                    .format(path, dict_val))
 
         # Otherwise, update target with this attribute
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -545,6 +545,21 @@ class FlowCfg():
             shutil.copy2(latest_report, os.path.join(dest_dir, 'report.html'))
             log.info("Copied report for %s", item.name)
 
+            if self.args.cov_reports:
+                cov_dir = os.path.join(item.scratch_path, 'cov_report')
+                if not os.path.exists(cov_dir):
+                    log.warning("No coverage reports found for %s at %s, skipping.", item.name, 
+                                cov_dir)
+                else:
+                    cov_dest_dir = os.path.join(repo_dir,
+                                                f"{self.name}_{flow}",
+                                                f"{self.name}_{flow}_{self.timestamp}",
+                                                os.path.basename(item.scratch_path),
+                                                'cov_report')
+                    os.makedirs(cov_dest_dir, exist_ok=True)
+                    shutil.copytree(cov_dir, cov_dest_dir, dirs_exist_ok=True)
+                    log.info("Copied coverage reports for %s", item.name)
+
         subprocess.run(["git", "-C", repo_dir, "add", "-A"], check=True, env=env)
         subprocess.run(["git", "-C", repo_dir, "commit", "-m",
                         f"[dashboard] Update {self.name} {self.flow} {self.timestamp}"],
@@ -585,10 +600,21 @@ class FlowCfg():
     def clone_or_pull(self, repository: str, local_path: str, env: dict):
         """Clone the repo if it doesn't exist locally, otherwise pull latest."""
         try:
-            if not os.path.exists(local_path):
-                subprocess.run(["git", "clone", repository, local_path], check=True, env=env)
+            if os.path.exists(local_path):
+                result = subprocess.run(
+                    ["git", "-C", local_path, "remote", "get-url", "origin"],
+                    check=True, capture_output=True, text=True, env=env
+                )
+                existing_remote = result.stdout.strip()
+                if existing_remote != repository:
+                    log.warning("Existing repo at %s points to %s, expected %s. Recloning.",
+                                local_path, existing_remote, repository)
+                    shutil.rmtree(local_path)
+                    subprocess.run(["git", "clone", repository, local_path], check=True, env=env)
+                else:
+                    subprocess.run(["git", "-C", local_path, "pull"], check=True, env=env)
             else:
-                subprocess.run(["git", "-C", local_path, "pull"], check=True, env=env)
+                subprocess.run(["git", "clone", repository, local_path], check=True, env=env)
         except FileNotFoundError:
             log.error("git is not installed or not on PATH. Cannot publish results.")
             raise
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -553,20 +553,23 @@ class FlowCfg():
         log.info("Results published successfully.")
 
     def _setup_ssh_env(self) -> dict:
-        """Start an ssh-agent, add the key via askpass, and return the env dict."""
+        """Return an env dict with SSH configured, using passphrase or deploy key."""
+        env = os.environ.copy()
+        env["GIT_SSH_COMMAND"] = "ssh -o StrictHostKeyChecking=no"
+
         passphrase = os.environ.get("SSH_KEY_PASSPHRASE")
+        if not passphrase:
+            # Rely on deploy key already configured in the environment
+            log.info("SSH_KEY_PASSPHRASE not set, assuming deploy key is available.")
+            return env
 
         with tempfile.NamedTemporaryFile(mode='w', suffix='.sh', delete=False) as f:
             f.write(f'#!/bin/sh\necho "{passphrase}"\n')
             askpass_script = f.name
         os.chmod(askpass_script, stat.S_IRWXU)
-
-        env = os.environ.copy()
         env["SSH_ASKPASS"] = askpass_script
         env["SSH_ASKPASS_REQUIRE"] = "force"
         env["DISPLAY"] = ""
-        env["GIT_SSH_COMMAND"] = "ssh -o StrictHostKeyChecking=no"
-
         try:
             agent = subprocess.run(["ssh-agent", "-s"], check=True, capture_output=True, text=True)
             for line in agent.stdout.splitlines():
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -497,7 +497,7 @@ class FlowCfg():
                                         relative_to)
         return "[%s](%s)" % (link_text, relative_link)
 
-    def publish_results(self, repository: str, flow: str):
+    def publish_results(self, repository: str, test: str):
         """Publish these results to a corresponding repository."""
         log.info("Publishing results to %s", repository)
         repo_dir = os.path.expanduser("./scratch/results_repo")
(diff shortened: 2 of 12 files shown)


matutem:
Does this impact the functionality provided by --select-cfgs arg?

LouisTheLuis:
@matutem I tested the functionality and it seems to behave the same as before... That said, I found something interesting (or to perhaps focus on). When I run something like:

```
./util/dvsim/dvsim.py ./hw/dv/all_sim_cfgs.hjson -r 1 --select-cfgs alert_handler -v m -ro +max_quit_count=40 --build-opts '-kdb -debug_access+all+reverse' --run-opts '+UVM_VERDI_TRACE=HIER +print_topology=1' -i smoke -t vcs --cov
```

It runs both `alert_handler` instances for both `top_darjeeling` and `top_earlgrey`. That's to be expected, as both contain the same name in their `cfg`; but perhaps it would be ideal to also be able to select a specific one with `--select-cfgs`.

matutem:
@LouisTheLuis regarding all_sim_cfgs, rather than complicating this I suggest if we wanted to just run the earlgrey alert_handler we could simply run with top_earlgrey_sim_cfgs and select alert_handler.

util/dvsim/CfgJson.py
```diff
@@ -89,8 +89,8 @@ def _load_single_file(target, path, is_first, arg_keys):
                 raise RuntimeError('{!r}: File is included by another one, '
                                    'but defines "use_cfgs".'
                                    .format(path))
-            if not isinstance(dict_val, list):
-                raise RuntimeError('{!r}: use_cfgs must be a list. Saw {!r}.'
+            if not isinstance(dict_val, list) and not isinstance(dict_val, dict):
+                raise RuntimeError('{!r}: use_cfgs must be a list or a dict. Saw {!r}.'
```
matutem, line 93: These checks are rather poor: a list of what, could be a list of ints or a list of dicts. Same for checking it is a dict of dicts. What we really want is a list of strings, or a dict of lists of strings. I think we may want to look into pydantic.

LouisTheLuis: To be fair, the options for `dvsim` are:

- List of strings (e.g. `top_darjeeling_sim_cfgs.hjson`)
- List of dicts (e.g. `top_darjeeling_lint_cfgs.hjson`)
- Dictionary of string:list (of strings or dicts) key-value pairs (e.g. `all_sim_cfgs.hjson`)

util/dvsim/FlowCfg.py
```diff
@@ -534,6 +534,20 @@ def publish_results(self, repository: str, flow: str):
...
+                if not os.path.exists(cov_dir):
+                    log.warning("No coverage reports found for %s at %s, skipping.", item.name, 
+                                cov_dir)
+                cov_dest_dir = os.path.join(repo_dir,
+                                            f"{self.name}_{flow}",
+                                            f"{self.name}_{flow}_{self.timestamp}",
+                                            os.path.basename(item.scratch_path),
+                                            'cov_report')
+                os.makedirs(cov_dest_dir, exist_ok=True)
+                shutil.copytree(cov_dir, cov_dest_dir, dirs_exist_ok=True)
```
vlpannel: Will this error out if `cov_dir` does not exist? Previously only raised warning (line 540) if it doesn't exist.

util/dvsim/CfgJson.py
matutem: Ah, you are right lint has more structure. I guess this is okay.

util/dvsim/dvsim.py
```diff
@@ -618,6 +618,11 @@ def parse_args():
                       help=('Rather than building or running any tests, '
                             'analyze the coverage from the last run.'))
 
+    covg.add_argument("--cov-reports",
```
matutem, line 639: How is this different from `--cov`? I thought with that flag we generated reports.

util/dvsim/dvsim.py
```diff
@@ -772,13 +813,25 @@ def main():
...
+            ["git", "ls-remote", args.publish],
+            capture_output=True,
+            env={**os.environ, "GIT_SSH_COMMAND": "ssh -o BatchMode=yes"}
+        ).returncode == 0
+
+        if not has_passphrase and not has_deploy_key:
+            log.fatal("--publish requires either SSH_KEY_PASSPHRASE to be set"
+                      " or a deploy key with access to the repository")
+            sys.exit(1)
+
```
matutem: This is duplicate from 779-789. Could you move the code to a function and invoke that instead?

util/dvsim/FlowCfg.py
```diff
@@ -508,37 +526,54 @@ def publish_results(self, repository: str, flow: str):
...
+                                    f"{self.name}_{test}_{self.flow}_{self.timestamp}",
                                     os.path.basename(item.scratch_path),
                                     'reports',
                                     'latest')
             os.makedirs(dest_dir, exist_ok=True)
             shutil.copy2(latest_report, os.path.join(dest_dir, 'report.html'))
             log.info("Copied report for %s", item.name)
 
+            if self.args.cov_reports:
+                cov_dir = os.path.join(item.scratch_path, 'cov_report')
```
matutem, line 550: Could you use the pathlib Path API instead? It would be good to gradually migrate away from os.path.

util/dvsim/FlowCfg.py
```diff
@@ -508,37 +526,54 @@ def publish_results(self, repository: str, flow: str):
...
+                if not os.path.exists(cov_dir):
+                    log.warning("No coverage reports found for %s at %s, skipping.", item.name,
```
matutem, line 552: If cov_dir doesn't exist we probably should not run the code up to line 550, so place that code in the `else`.

util/dvsim/FlowCfg.py
```diff
@@ -508,37 +526,54 @@ def publish_results(self, repository: str, flow: str):
...
+                                            os.path.basename(item.scratch_path),
+                                            'cov_report')
+                os.makedirs(cov_dest_dir, exist_ok=True)
+                shutil.copytree(cov_dir, cov_dest_dir, dirs_exist_ok=True)
+                log.info("Copied coverage reports for %s", item.name)
+
         subprocess.run(["git", "-C", repo_dir, "add", "-A"], check=True, env=env)
         subprocess.run(["git", "-C", repo_dir, "commit", "-m",
-                        f"[dashboard] Update {self.name} {self.flow} {self.timestamp}"],
+                        f"[dashboard] Update {self.name}_test_{self.flow} {self.timestamp}"],
```
matutem:
```suggestion
                        f"[dashboard] Update {self.name}_{test}_{self.flow} {self.timestamp}"],
```

util/dvsim/FlowCfg.py
```diff
@@ -550,7 +551,7 @@ def publish_results(self, repository: str, flow: str):
 
         subprocess.run(["git", "-C", repo_dir, "add", "-A"], check=True, env=env)
         subprocess.run(["git", "-C", repo_dir, "commit", "-m",
-                        f"[dashboard] Update {self.name} {self.flow} {self.timestamp}"],
+                        f"[dashboard] Update {self.name}_test_{self.flow} {self.timestamp}"],
```
vlpannel:
```suggestion
                        f"[dashboard] Update {self.name}_{test}_{self.flow} {self.timestamp}"],
```

util/dvsim/dvsim.py
LouisTheLuis: Regular `--cov` doesn't hyperlink the tool-specific reports (generated by VCS or Xcelium, with all the coverage stuff). The idea is to let `--cov-reports` do that for a user that has the licenses available and wants to include the Synopsys/Cadence reports.

util/dvsim/FormalCfg.py
```diff
@@ -146,14 +147,33 @@ def gen_results_summary(self):
...
-                table.append(cfg.result_summary[cfg.name])
+                row = cfg.result_summary[cfg.name]
             except KeyError as e:
-                table.append([cfg.name, "ERROR", "N/A", "N/A", "N/A"])
+                row = [cfg.name, "ERROR", "N/A", "N/A", "N/A"]
                 log.error(
                     "cfg: %s could not find generated results_summary: %s",
                     cfg.name, e)
+
+                if batchgroups:
```
matutem: Should this be un-indented? Otherwise it is only run when there is an exception.

matutem: It seems there is some code that is shared by a few types of Cfgs. Could some of it be placed in a FlowCfg function so we don't repeat it?

util/dvsim/dvsim.py
LouisTheLuis: Done.

util/dvsim/FlowCfg.py
LouisTheLuis: Fixed.

util/dvsim/FormalCfg.py
LouisTheLuis: Done.

util/dvsim/FlowCfg.py
LouisTheLuis: Fixed.

util/dvsim/FlowCfg.py
```diff
@@ -508,37 +526,55 @@ def publish_results(self, repository: str, flow: str):
...
 
+            if self.args.cov_reports:
+                cov_dir = os.path.join(item.scratch_path, 'cov_report')
+                if not os.path.exists(cov_dir):
+                    log.warning("No coverage reports found for %s at %s, skipping.", item.name,
+                                cov_dir)
+                else:
+                    cov_dest_dir = os.path.join(repo_dir,
+                                                f"{self.name}_{flow}",
+                                                f"{self.name}_{flow}_{self.timestamp}",
```
vlpannel:
```suggestion
                                                f"{self.name}_{self.flow}",
                                                f"{self.name}_{self.flow}_{self.timestamp}",
```

vlpannel (approved):
LGTM, though I understand that dvsim is a complex beast (there may be something I'm missing).

util/dvsim/FlowCfg.py
LouisTheLuis: If I modify it, I would rather do it in a separate PR. I took the initial decision of using `os.path` because for the purposes of doing `git clone` and `git pull` operations, it was easier to deal with strings rather than `Path` objects.

util/dvsim/FlowCfg.py
LouisTheLuis: Already fixed by placing it under an `else:` block.

util/dvsim/FlowCfg.py
LouisTheLuis: Fixed.

util/dvsim/FlowCfg.py
LouisTheLuis: Fixed.

util/dvsim/FormalCfg.py
LouisTheLuis: I refactored the `if batchgroups` section into a single check in `FlowCfg.py`. Some stuff could be refactored further, that said, there are some assumptions and small nits within each method (`SimCfg` has to move the "Overall Coverage" key to the end). I believe I can refactor `gen_results_summary` into a couple of methods; that said, I will consider fixing it later (I for one don't want to modify much in the `FormalCfg` flow, for example, as we don't have access to FPV tools and I can't really fully test the flow).

LouisTheLuis: I'll make a note of this in the project and revisit after.

PR #76, opened 2026-04-02
[doc,hw,sw] Remove links to documents that aren't publicly visible

qmn:
These all result in Access Denied, which makes it hard to understand design decisions.

Going forward, these kinds of diagrams and documents need to be on the repo, whether as part of a PR's comments or in the source itself. It's also just easier to keep track of.

### DIFF
--- a/doc/security/specs/ownership_transfer/README.md
+++ b/doc/security/specs/ownership_transfer/README.md
@@ -184,10 +184,6 @@ Definitions:
 <table>
   <tr>
     <td>
-    <!--
-    svgs are made in google docs, and can be found here:
-    https://docs.google.com/drawings/d/1hnC2EgkYpkxhVJ6I0prxdoQbglstqfwxvBYh_JIt8TA/edit?usp=sharing
-    -->
       <img src="ownership_transfer_fig1.svg" width="900" alt="fig1">
     </td>
   </tr>
--- a/hw/ip/acc/doc/theory_of_operation.md
+++ b/hw/ip/acc/doc/theory_of_operation.md
@@ -87,11 +87,6 @@ If a fault causes the state to become zero, ACC raises a `BAD_INTERNAL_STATE` fa
 
 ### Operational States
 
-<!--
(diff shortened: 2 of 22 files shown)


hw/ip/acc/doc/theory_of_operation.md
```diff
@@ -88,8 +88,6 @@ If a fault causes the state to become zero, ACC raises a `BAD_INTERNAL_STATE` fa
 ### Operational States
 
 <!--
-Source: https://docs.google.com/drawings/d/1C0D4UriRk5pKGFoFtAXYLcJ1oBG1BCDd2omCLPYHtr0/edit
-
 Download the SVG from Google Draw, open it in Inkscape once and save it without changes to add width/height information to the image.
 -->
```
vlpannel:
```suggestion
```
The whole comment can probably be deleted.

vlpannel (approved):
LGTM with one suggestion.

doc/security/specs/ownership_transfer/README.md
```diff
@@ -184,10 +184,6 @@ Definitions:
 <table>
   <tr>
     <td>
-    <!--
```
matutem, line 187: It is understandable but uncomfortable to loose access to the drawing tool. I think @ysekhar was using a FOSS format and generator. It would be good to suggest or recommend some tooling to keep the images editable. What do you use?

matutem (approved):
I think we would benefit from editable images, so may want to use some FOSS data format. How an we come up with some contending foss flows?

LGTM

hw/ip/acc/doc/theory_of_operation.md
qmn: Good point, thanks!

doc/security/specs/ownership_transfer/README.md
qmn: I use Inkscape (or, if I'm feeling sufficiently bold and changes sufficiently small, Vim). As long as we have access to the drawing source in an open format, it should be fine.

PR #77, opened 2026-04-03
[otp_ctrl,dv] Fix value of OTP_MACRO_FULL_WIDTH

matutem:
This value should use the otp data word width, which is otp_ctrl_macro_pkg::OtpWidth.
The old code was using otp_ctrl_reg_pkg::OtpWidth, which is 2 (bytes).

Notice OTP_MACRO_FULL_WIDTH was only used in uvm_info messages so it would not affect test outcomes.


qmn (approved):
LGTM, thank you!

PR #85, opened 2026-04-07
[otp_macro,lint] Enable lint of otp_macro via top_cfgs

matutem:
Add otp_macro in earlgrey and darjeeling top_cfgs. Add missing mapping for otp_ctrl_macro_pkg in the top core file for the tops above.


LouisTheLuis (approved):
LGTM.

PR #88, opened 2026-04-07
[sw,build] Make LLVM 22.1.2 the default C/C++ build toolchain

rtorok-zr:
This PR replaces the host `gcc` toolchain and `@lowrisc_rv32_toolchain` with LLVM 22.1.2 as the primary toolchain for C/C++ compilation, linking, and `binutils`.

Enabling the repository to work with LLVM required many small fixes to address differences in behavior between LLVM and `gcc`. Highlights include:
1. RISC-V `zbr`, `zbf`, and `zbp` extensions to the bitmanip instruction set were [removed from LLVM](https://github.com/lowRISC/ibex/pull/965) c. December 2022. However, Ibex still supports these instructions, so there were replaced in the assembly code with equivalent [`.insn` directives](https://sourceware.org/binutils/docs/as/RISC_002dV_002dFormats.html).
2. `llvm-ar` does not flatten archives by default. This required using "quick append" with the `L` option to retain the past behavior of `ot_static_library`.
3. The ACC build script adds a flag to `llvm-objcopy` to copy the `.text` section to itself to clear the `executable` flag, which is not cleared by `--set-section-flags`, unlike in GNU `objcopy`. Retaining the `executable` flag would make the memory protection enable execution in the `.rodata` region, which could increase the attack surface for ROP-style attacks, and also caused a python script in QEMU that generates flash images to complain that the `.elf` and `.bin` files disagreed about the code size, because the ELF would incorrectly have the executable bit set on the `.rodata` section.
4. Linkage for ACC binaries needed to be adjusted to give the `.scratchpad` region the `ALLOC` annotation, (which it should have had to begin with). In `ld.lld`, this is required to assign the region to the correct address, unlike GNU, which kindly honored our address request even though our configuration was strictly incorrect.
6. Worked around a rough edge in `ld.lld` with performing arithmetic on code symbols, which affected how we defined `_manifest_entry_point` in linker scripts.
7. Manually `KEEP` the `.data` symbols in ACC binaries to ensure they are exported and avoid linker errors because  symbols unreferenced in the script they are declared, like `_acc_remote_app_acc_char_dmem_write_mem`, can be optimized out. With GNU, this was achieved with the `-gc-keep-exported` symbol, which `ld.lld` does not support.

I recommend reviewing one commit at a time; the penultimate commit produced many auto-generated formatting fixes that are responsible for the majority of the diff size, because our `clang-format` was upgraded from `v10.0.0`.


sw/device/tests/penetrationtests/firmware/fi/ibex_fi.c
qmn, line 1: Can you file an issue about the register assignment instability?

What error happened to the FI tests with the bad clobbers?

sw/device/tests/rstmgr_cpu_info_test.c
```diff
@@ -287,7 +287,7 @@ bool test_main(void) {
       check_prev_state(dump.previous_fault_state,
                        (dif_rv_core_ibex_previous_crash_dump_state_t){
                            .mtval = (uint32_t)kIllegalAddr1,
-                           .mpec = (uint32_t)kDoubleFaultFirstAddr + 4,
```
qmn, line 290: This seems to be a functional change unrelated to the commit. Could you elaborate what this change was about?

sw/device/tests/rv_core_ibex_isa_test.S
qmn, line 1: The `.insn` directives are fine for now. For better maintainability, in a future PR, I'd recommend making assembler macros that hide away the guts of the `.insn` directive.

sw/device/silicon_creator/lib/cert/dice_cwt.c
```diff
@@ -107,16 +107,18 @@ enum cwt_cert_expectations {
...
 
 // Reusable buffer for checking cose sign1 cert identity.
 static char expected_cose_sign1_id[kDiceCwtCoseSign1IdSizeBytes] = {
-    0x02,                            // mapKey 2 (subject id)
-    0x78, kIssuerSubjectNameLength,  // 64-byte text header
-    // Remaining bytes will be filled during check.
+    0x02,  // mapKey 2 (subject id)
+    0x78,
+    kIssuerSubjectNameLength,  // 64-byte text header
+                               // Remaining bytes will be filled during check.
```
qmn: I disagree with clang-format here, this last comment belongs at the same level of the other bytes but I'm not sure how to get around it.

qmn (approved):
Simply tremendous work!

I understand that the purpose of this PR is to remove the dependence on the host compiler (and also to get Verilator built on LLVM). But would this PR also mean that we no longer support GCC? I can imagine that someone, for whatever reason, would want to loop in their own toolchain. By what mechanism would someone switch to their toolchain (or, for that matter, back to the host compiler)? I ask mostly for my edification.

These commits LGTM, just be sure to trim the fix to CONTRIBUTING.md.

sw/device/tests/rstmgr_cpu_info_test.c
rtorok-zr: I meant to mention this to you earlier. When I changed to LLVM, all of a sudden the fault address was off by 4 bytes. The `+ 4` lines up with an analogous check later in the test, and I'm not sure why this was passing without it on `gcc`. I'm guessing it's a difference in how the binary generated by the compiler causes the hardware to increment the PC before the fault happens or not, but I might need someone to look at this later.

sw/device/silicon_creator/lib/cert/dice_cwt.c
rtorok-zr: Added `// clang-format off` to the struct to get the desired indentation.

sw/device/tests/penetrationtests/firmware/fi/ibex_fi.c
rtorok-zr: The value of `a2`/`x12` did not match the expected value for tests that used `kIbexFiSubcommandCharCombi` because the `a2` was clobbered when it tried to save the register state at the end of the test.

rtorok-zr: Opened https://github.com/pavona/pavona-staging/issues/141

PR #90, opened 2026-04-08
[dv,clkmgr] Fix clkmgr_frequency_vseq

matutem:
The calib_rdy_i is unconnected so it should not affect the test. This test should always consider it to be MuBi4True.


LouisTheLuis (approved):
LGTM.

PR #92, opened 2026-04-08
[otp_ctrl,rtl] Fix defect introduced in lowRISC/opentitan#28108

matutem:
The PR mentioned above attempts to allow reads of zeroization markers with no concern for integrity errors, but it fails to restrict this only to zeroizable partitions.


LouisTheLuis (approved):
LGTM.

PR #93, opened 2026-04-08
[sram_ctrl,dv] Fix path of $assertoff

matutem:
Update the path in an $assertoff statement to consider the racl adapter.


LouisTheLuis (approved):
LGTM.

PR #97, opened 2026-04-09
[doc] Augmentations for generating documents

vlpannel:
This PR includes three main commits:
* Add an increased verbosity flag to cmdgen to print out the diff (what will be updated).
* Augment `util/selfdoc.py` to capture stdout from command line.
* Use the new selfdoc stdout functionality to keep tools' command line usages up to date (using `util/selfdoc.py quote:path/to/tool.py --help`).

These are a few simple changes that will keep the repo documentation more up-to-date as changes accrue over time.

### DIFF
--- a/util/doc/vendor.md
+++ b/util/doc/vendor.md
@@ -8,21 +8,32 @@ The `util/vendor` tool can go beyond simply copying in source files: it can patc
 
 ## Tool usage overview
 
-```text
-usage: vendor [-h] [--refresh-patches] [--commit] [--verbose] file
+<!-- BEGIN CMDGEN util/selfdoc.py quote:util/vendor.py --help -->
+```
+$ util/vendor.py --help
+usage: vendor [-h] [--update] [--refresh-patches] [--commit]
+              [--desc-override DESC_OVERRIDES] [--verbose]
+              file
 
-vendor, copy source code from upstream into this repository
+A tool to copy source code from upstream into this repository. For an
+introduction to using this tool, see doc/contributing/hw/vendor.md in this
+repository. For full documentation, see util/doc/vendor.md.
 
 positional arguments:
-  file               vendoring description file (*.vendor.hjson)
+  file                  vendoring description file (*.vendor.hjson)
 
-optional arguments:
-  -h, --help         show this help message and exit
-  --update, -U       Update locked version of repository with upstream changes
-  --refresh-patches  Refresh the patches from the patch repository
-  --commit, -c       Commit the changes
-  --verbose, -v      Verbose
+options:
+  -h, --help            show this help message and exit
+  --update, -U          Update locked version of repository with upstream
+                        changes
+  --refresh-patches     Refresh the patches from the patch repository
+  --commit, -c          Commit the changes
+  --desc-override DESC_OVERRIDES, -D DESC_OVERRIDES
+                        Override a setting in the description file. Format:
+                        -Dsome.key=value. Can be used multiple times.
+  --verbose, -v         Verbose
 ```
+<!-- END CMDGEN -->
 
 ## The vendor description file
 
--- a/util/raclgen/README.md
+++ b/util/raclgen/README.md
@@ -18,15 +18,16 @@ For more information about the concept of RACL itself, see the [general RACL ove
 
 ## Usage
 
-```shell
+<!-- BEGIN CMDGEN util/selfdoc.py quote:util/raclgen.py --help -->
+```
 $ util/raclgen.py --help
-usage:
-raclgen.py --doc DOC
-    Generates markdown documentation of the RACL configuration for a given top.
+usage: 
+    raclgen.py --doc DOC
+        Generates markdown documentation of the RACL configuration for a given top.
 
-raclgen.py --racl-config RACL_CONFIG --ip IP --mapping MAPPING [--if-name IF_NAME]
-    Generates the RACL policy selection vector for the given IP, RACL mapping, and interface name.
-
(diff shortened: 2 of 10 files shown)


util/doc/vendor.md
```diff
@@ -8,21 +8,35 @@ The `util/vendor` tool can go beyond simply copying in source files: it can patc
...
+```
+$ util/vendor.py --help
+usage: vendor [-h] [--update] [--refresh-patches] [--commit]
+              [--desc-override DESC_OVERRIDES] [--verbose]
+              file
+
+A tool to copy source code from upstream into this repository. For an
+introduction to using this tool, see doc/contributing/hw/vendor.md in this
+repository (on the internet at
+https://opentitan.org/book/doc/contributing/hw/vendor.html). For full
```
matutem: This opentitan.org link should be replaced by a relative one. Also below.

util/raclgen/README.md
```diff
@@ -18,15 +18,16 @@ For more information about the concept of RACL itself, see the [general RACL ove
 
 ## Usage
 
-```shell
+<!-- BEGIN CMDGEN util/selfdoc.py quote:util/raclgen.py --help -->
+```
 $ util/raclgen.py --help
-usage:
+usage:
```
matutem: Please remove the extra space at the end of line 23 in raclgen.py.

util/raclgen/README.md
```diff
@@ -18,15 +18,16 @@ For more information about the concept of RACL itself, see the [general RACL ove
...
 $ util/raclgen.py --help
-usage:
+usage: 
 raclgen.py --doc DOC
     Generates markdown documentation of the RACL configuration for a given top.
 
 raclgen.py --racl-config RACL_CONFIG --ip IP --mapping MAPPING [--if-name IF_NAME]
     Generates the RACL policy selection vector for the given IP, RACL mapping, and interface name.
-
+
```
matutem: The multi-line usage string in raclgen.py line 23 probably needs to use python's textwrap.dedent and be formatted with indentation, so perhaps this will work better?

....           (usage=textwrap.dedent('''\
      %(prog)s --doc DOC
          Generates...
           last line.
      '''))

The first back-slash is to avoid the extra new line. You need to indent the string with white spaces. All the lines will have the minumum number of whitespaces that each of the lines have.

dedent and multi-line strings tend to go together 😃 

This doesn't render well, but check https://ben11kehoe.medium.com/python-multi-line-string-literals-with-textwrap-dedent-d4cf3c4a6d02 for an explanation.

util/gen_system_list.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/python3
+
```
matutem: Please add a file docu-string explaining what this does. Another advantage of file-level docu-strings is that the can be used by the help message of the tool, so they are ideal for both documenting the file and the tool usage, all in one :)

util/gen_system_list.py
```diff
@@ -0,0 +1,97 @@
...
+from pathlib import Path
+import hjson
+import re
+import sys
+
```
matutem: I don't see how this relates to the rest of the changes. Perhaps split it into a separate commit in this same PR?

util/selfdoc.py
```diff
@@ -8,6 +8,8 @@
 import tlgen
 import topgen.selfdoc as topgen_selfdoc
 from typing import TextIO
+import subprocess
+from cmdgen import REPO_ROOT
```
matutem, line 12: This resolved REPO_ROOT seems pretty useful. Perhaps it should go in basegen and all tools could import it, rather than each one needing to do the `parent` song-and-dance. Not necessarily on this PR.

util/selfdoc.py
```diff
@@ -21,12 +23,18 @@ def generate_selfdocs(tool: str, fout: TextIO):
         fout.write(tlgen.selfdoc(heading=3, cmd='tlgen.py --doc'))
     elif tool == "topgen":
         topgen_selfdoc.document(fout)
+    elif "quote:" in tool:  # document stdout
```
matutem: Could you make this `quote:` string optional? Meaning this can be a plain else, and internally it detects and strips an optional quote, runs the subprocess, and the fout.write could optionally add quotes around the string. This way the quotes processing becomes an optional added bonus.

Okay to file an issue and fix it in a separate PR.

util/selfdoc.py
vlpannel: Agreed; there are a number of things that could go in basegen/some common library for all the tooling, `REPO_ROOT` being one of them. For another PR.

util/raclgen/README.md
vlpannel: I think the raclgen doc intentionally uses the newline to separate using raclgen for documentation versus getting RACL policies. This is also why there is the extra space after `usage:` (argparse expects more text on the same line but there is only a new line.

In a new commit, I've decided to keep it multiline, make the indentation more nicely formatted, and simply use a backspace character to rid of the extra space.

Let me know if this formatting should be changed further.

util/gen_system_list.py
vlpannel: This new file moved to #101.

util/gen_system_list.py
vlpannel: Moved to #101.

util/raclgen/README.md
matutem: Sounds good. If we had a markdown lint checker it would flag the extra character :)

matutem (approved):
LGTM

PR #98, opened 2026-04-10
[prim_sha2,rtl] Adds a block load interface to prim_sha2

apinise:
These commits added a block load interface to the `prim_sha2` IP and created a separate `prim_sha2_compression` module with the wide interface. The previous `prim_sha2_wrapper` now instantiates the compression core and interfaces with it, maintaining compatibility with `HMAC`.


matutem (approved):
LGTM

PR #99, opened 2026-04-10
[ci,prepare-env] Remove Google Cloud authentication

qmn:
We'll need to set up Pavona's GCP caches in due time, but for now, get nightlies and weeklies going.


qmn:
Mooted by #118

PR #100, opened 2026-04-12
[pwrmgr,dv] Add pwrmgr_rstreqs assertions

matutem:
This has two commits for pwrmgr DV:
- Add an argument to responder macros that control verbosity
- Bind interface pwrmgr_rstreqs_sva_if to pwrmgr: contains assertions checking reset requests and causes.


ysekhar (approved):
LGTM

PR #102, opened 2026-04-13
[util,make] Enhance Makefile and improve reggen/gen_dv

matutem:
This has 2 simple commits:
- Enforce cmdgen to run after top
- Clean error handling and use mako template for reggen/gen_dv


util/reggen/ral_pkg.core.tpl
```diff
@@ -0,0 +1,24 @@
+CAPI=2:
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+# DO NOT EDIT:
+# This was generated by reggen
+name: pavona:dv:${lblock}_ral_pkg
+description: "${lblock.upper()} RAL package"
```
vlpannel:
```suggestion
description: "${lblock.title()} RAL package"
```
I'm guessing you just want the first letter of each word capitalized, right? Ignore this if you're trying to make it all caps.

util/reggen/ral_pkg.core.tpl
```diff
@@ -0,0 +1,24 @@
+CAPI=2:
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
```
vlpannel: Nit: I'm still unsure of how we are treating copyright for autogenerated files, and templates can still have comments without rendering them using `##` instead. This can be a future PR (and topic of discussion) though.

hw/Makefile
```diff
@@ -78,7 +79,7 @@ $(tops_gen): %_gen:
 	${PRJ_DIR}/util/topgen.py -t $(top_hjson) -s $(top_seed_hjson) -o ${PRJ_DIR}/hw/$*/ ${toolflags}
 
 # This runs cmdgen on all md files of the repo using xargs for parallelism
-cmdgen:
+cmdgen: top
```
vlpannel, line 83: Nit: I'm not super clear on the rationale for having cmdgen always run after top with this makefile, but I imagine it might be useful. I can see a trivial case where you need cmdgen to run on files not related to topgen output (eg., changing multiple docs that aren't top-specific), in which case I guess you could just run cmdgen manually (using `util/cmdgen.py -u '**/*.md'`).

Why not always run cmdgen as part of the top target? (This would basically eliminate the need for the cmdgen and top_and_cmdgen targets.)

vlpannel (approved):
Overall LGTM with some nits.

hw/Makefile
matutem: cmdgen is run after top because top generates md files from templates. Basically cmdgen needs to see all the updated md files under top*/ip_autogen.

We could have per-top cmdgen, where each top would run topgen and when done, run cmdgen for the given top's md files. On top of that, each top can be run in parallel, as in `make -j 4 ...`. The whole top_and_cmdgen is pretty fast already, but we could still improve it.

As for running parts of make manually, that is always an option, but the idea of this Makefile is that running `all` really does it all.

util/reggen/ral_pkg.core.tpl
matutem: I see this report (https://www.copyright.gov/newsnet/2025/1060.html) says even some AI generated content can be copyrighted if there is sufficient human intervention beyond clocking some buttons. I also see this: https://opensource.stackexchange.com/questions/10786/can-mechanically-generated-code-be-licensed. I think in our case if we hold a copyright to the generator and to the template a case can be made it can be copyrighted.

But frankly I agree this needs legal guidance.

util/reggen/ral_pkg.core.tpl
matutem: Yes, I think it is best to have the complete block in upper case. Think of RV_DM for example: the name is composed of abbreviations, and it is comon use to have abbreviations like this be in upper case. It would be hard to distinguish what is an abbreviation and what isn't, so all upper seems better.

A sensible alternative would be to just use whatever name was passed to the generator, with no modifications, and we would then pass the buck to the <block>.hjson file. I would also support that.

hw/Makefile
vlpannel: This makes sense. I agree that top_and_cmdgen is not all that slow, so I think this is good as is with the cmdgen requiring the top (no need for per-top cmdgen).

util/reggen/ral_pkg.core.tpl
vlpannel: Agreed. This is interesting but should be reserved for a future PR.

matutem: Good point. Done.

PR #103, opened 2026-04-15
[dv,dvsim] Allow use of graded seeds in simulations

LouisTheLuis:
This PR contains four commits:
- The first one introduces the top name as a `variant` for each `ip_template` simulation configuration file. This allows for easier handling by `dvsim`.
- The second one includes a script that allows `urg` to be ran on a batch of VCS runs, allowing for a CSV file to be generated for analysis, and a `hjson` file that can be used to override the seeds in a simulation run.
- The third one includes a new flag to `dvsim` that allows the user to include the `hjson` file and override the seeds. Through this, `urg` can be run through `grade_regression_vcs.py` and the generated file can be used in order to achieve similar coverage for a smaller amount of runs.
-  The fourth one fixes a bug in the Lint (slow) script.

### DIFF
--- a/util/dvsim/CfgFactory.py
+++ b/util/dvsim/CfgFactory.py
@@ -7,6 +7,7 @@ import sys
 import os
 
 from CfgJson import load_hjson
+from utils import parse_hjson
 
 import FormalCfg
 import CdcCfg
@@ -65,7 +66,7 @@ def _load_cfg(path, initial_values):
     return (found_cls, hjson_data)
 
 
-def _make_child_cfg(path, args, initial_values):
+def _make_child_cfg(path, args, initial_values, seed_index=None, default_reseed=1):
     try:
         cls, hjson_data = _load_cfg(path, initial_values)
     except RuntimeError as err:
@@ -80,17 +81,57 @@ def _make_child_cfg(path, args, initial_values):
                            'itself included from another configuration.'
                            .format(path))
 
+    if seed_index:
+        reseed_override(hjson_data, seed_index, default_reseed)
+
     # Call cls as a constructor. Note that we pass None as the mk_config
     # argument: this is not supposed to load anything else.
     return cls(path, hjson_data, args, None)
 
 
+def reseed_override(hjson_data: dict, seed_index: dict, default_reseed: int):
+    '''Override resets for each test that has a reseed if the flag --reseed-source was passed.
+    '''
+    hjson_name = hjson_data["name"]
+    log.debug("========== %s ==========", hjson_name)
+    log.debug("before:")
+    count_reseeds(hjson_data)
+
+    hit = False
+    variant = hjson_data.get("variant")
+    for hjson_test in hjson_data.get("tests", []):
+        resolved_name = hjson_test["name"].replace("{name}", hjson_data["name"])
+
+        for candidate in [hjson_test["name"], resolved_name]:
+            key = (candidate, hjson_data["name"], variant)
+            if key in seed_index:
+                hit = True
+                hjson_test["reseed"] = seed_index[key]["reseed"]
+                break  # Assume one match per test
+
+    if "reseed" in hjson_data.keys() and hit:
+        hjson_data["reseed"] = default_reseed
+
+    log.debug("after:")
+    count_reseeds(hjson_data)
+
+
+def count_reseeds(hjson_data):
+    '''Helper function used in debug for counting seeds after override
+    '''
+    total = 0
+    for test in hjson_data.get("tests", []):
+        global_reseed = hjson_data.get("reseed", 1)
+        reseed = test.get("reseed", global_reseed)
+        testname = test["name"].replace("{name}", hjson_data["name"])
+        log.debug(" %s: %d", testname, reseed)
+        total += reseed
+    log.debug("  Total: %d", total)
+    return total
+
+
 def make_cfg(path, args, proj_root):
     '''Make a flow config by loading the config file at path
-
-    args is the arguments passed to the dvsim.py tool and proj_root is the top
-    of the project.
-
     '''
     initial_values = {
         'proj_root': proj_root,
@@ -105,9 +146,26 @@ def make_cfg(path, args, proj_root):
         log.error(str(err))
         sys.exit(1)
 
+    seed_index = {}
+    default_reseed = 1
+    if args.reseed_source:
+        seed_data = parse_hjson(args.reseed_source)
+
+        # Collect test seeds in a lookup dictionary
+        for seed_test in seed_data.get("tests", []):
+            variant = seed_test.get("variant")
+            key = (seed_test["test"], seed_test["name"], variant)
+            seed_index[key] = seed_test
+
+        default_reseed = seed_data.get("reseed", 1)
+
+        # If it is a single config, just change the reseeds now
+        if 'use_cfgs' not in hjson_data and seed_data:
+            reseed_override(hjson_data, seed_index, default_reseed)
+
     def factory(child_path):
         child_ivs = initial_values.copy()
         child_ivs['flow'] = hjson_data['flow']
-        return _make_child_cfg(child_path, args, child_ivs)
+        return _make_child_cfg(child_path, args, child_ivs, seed_index, default_reseed)
 
     return cls(path, hjson_data, args, factory)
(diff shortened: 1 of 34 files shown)


matutem:
Why do we need the top name in the hjson? I thought the segmenting of batch files you introduced could already be used for this.

LouisTheLuis:
@matutem The point of the batchgroups is that they can be arbitrary; we should not rely on them defining which `variant` an IP corresponds to. 
If we want to decide how to reseed `alert_handler` in two separate tops, we should look at their variants; both could be grouped in the same `batchgroup`, after all (one scenario for this is two instances of the same block within a top).

util/grade_regression_vcs.py
```diff
@@ -0,0 +1,327 @@
...
+        "urg", "-full64",
+        "-dir", str(merged),
+        "-grade", "testfile", "cost", "cputime",
+        "-report", str(report),
+    ]
+    if dry_run:
+        log.info(f"[{unit}] would run: {' '.join(cmd)}")
+        return 0, 0.0
+
+    report.mkdir(parents=True, exist_ok=True)
```
matutem: try-except since mkdir could fail.

util/grade_regression_vcs.py
```diff
@@ -0,0 +1,327 @@
...
+    log_path = report / "urg_grade.log"
+    start = time.time()
+    with log_path.open("w") as log_file:
```
matutem: try-except around the log_file ops. Notice subproces can also raise an exception.

util/grade_regression_vcs.py
```diff
@@ -0,0 +1,327 @@
...
+            log.warning(f"urg failed for {len(failed)} unit(s); continuing to aggregate the rest")
+
+        if args.dry_run:
+            sys.exit()
+
+    # -------- Phase 2: Aggregate --------
+    rows = []
+    per_unit_summary = []
+    missing = []
+    hjson_entries = []
```
matutem, line 247: I think we should generate a hjson file per unit, not a global one.

util/dvsim/CfgFactory.py
```diff
@@ -80,11 +81,62 @@ def _make_child_cfg(path, args, initial_values):
...
+    if seed_data:
+        reseed_override(hjson_data, seed_data)
+
     # Call cls as a constructor. Note that we pass None as the mk_config
     # argument: this is not supposed to load anything else.
     return cls(path, hjson_data, args, None)
 
 
+def reseed_override(hjson_data: dict, seed_data: dict):
+    '''Override resets for test if the flag --reseed-source was passed.
```
matutem: Override reseeds for ...

util/dvsim/CfgFactory.py
```diff
@@ -80,11 +81,62 @@ def _make_child_cfg(path, args, initial_values):
...
+    '''
+    hjson_name = hjson_data["name"]
+    log.debug(f"========== %s ==========", hjson_name)
+    log.debug("before:")
+    count_reseeds(hjson_data)
+
+    hit = False
+    for hjson_test in hjson_data.get("tests", []):
+        for seed_test in seed_data.get("tests", []):
```
matutem: This is O(N^2). If you sort both you can process them linearly, so the whole thing ends up O(N log(N)) due to sorting.

util/dvsim/CfgFactory.py
```diff
@@ -80,11 +81,62 @@ def _make_child_cfg(path, args, initial_values):
...
+            cfg_match = seed_test["name"] == hjson_data["name"]
+            if "variant" in seed_test:
+                cfg_match = cfg_match and seed_test["variant"] == hjson_data.get("variant")
+
+            if name_match and cfg_match:
+                hit = True
+                hjson_test["reseed"] = seed_test["reseed"]
+                break # Assume one match per test
+
+    if "reseed" in hjson_data.keys() and hit:
```
matutem, line 112: Huh? I don't understand what a global reseed does.

LouisTheLuis: It'd probably be better called "default". It's the reseed value used when tests don't explicitly assign one. Some tests don't contribute to coverage; so we reset the global reseed to 1 (still run the test for functionality purposes, but reduce the amount of times it's run).

util/dvsim/dvsim.py
```diff
@@ -589,6 +589,13 @@ def parse_args():
                              'maintaining the ratio of numbers of runs '
                              'between different tests.'))
 
+    seedg.add_argument("--reseed-source",
+                       "-rs",
+                       metavar="PATH",
+                       help=('Path to a HJSON file used as the source for '
```
matutem, line 595: Maybe mention that this reseeds are based on test grading.

util/dvsim/CfgFactory.py
```diff
@@ -80,11 +81,62 @@ def _make_child_cfg(path, args, initial_values):
...
+    log.debug("before:")
+    count_reseeds(hjson_data)
+
+    hit = False
+    for hjson_test in hjson_data.get("tests", []):
+        for seed_test in seed_data.get("tests", []):
+            # Check name matches (either direct or with {name} substituted)
+            name_match = (
+                hjson_test["name"] == seed_test["test"] or
+                hjson_test["name"].replace("{name}", hjson_data["name"]) == seed_test["test"]
```
matutem: I don't understand what this is doing. Why will the test names diverge?

util/dvsim/CfgFactory.py
matutem: Remove extra line.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: As we currently don't have a clear view of the structure of the repository, it is preferable to keep a single one for weekly run usage. Eventually, the script can be expanded to include per block reseed generation.

util/dvsim/CfgFactory.py
```diff
@@ -105,9 +157,17 @@ def make_cfg(path, args, proj_root):
         log.error(str(err))
         sys.exit(1)
 
+    seed_data = None
+    if args.reseed_source:
+        print(args.verbosity)
+        seed_data = parse_hjson(args.reseed_source)
+        # If it's a single config, just change the reseeds now
+        if 'use_cfgs' not in hjson_data and seed_data:
```
matutem, line 163: At this point you process each sim_cfg independently. And the seed data is global, so the overhead of finding the test among all tests run weekly is non-trivial.

util/dvsim/CfgFactory.py
LouisTheLuis: Could be simplified. Fixed.

util/dvsim/CfgFactory.py
LouisTheLuis: Done.

util/dvsim/dvsim.py
LouisTheLuis: Done.

util/dvsim/CfgFactory.py
LouisTheLuis: Fixed; now it iterates over both linearly, so it will be $O(N)$ (or more like, $O(N+M)$ ).

util/dvsim/CfgFactory.py
LouisTheLuis: For clarity: `Override resets for each test that has a reseed if the flag --reseed-source was passed`. Fixed.

util/dvsim/CfgFactory.py
LouisTheLuis: I think I can refactor the dictionary to an index before passing the seeds; that way we get rid of the overhead of checking the list itself. Fixed.

matutem (approved):
Looks good to get the ball rolling.

qmn (approved):
LGTM!

qmn:
Actually, I missed something in Lint (slow), there seems to be a variable used before assignment. Is that an easy fix?

LouisTheLuis:
@qmn Just fixed it on another commit.

PR #105, opened 2026-04-16
[dvsim] Fix quotation issue with SlurmLauncher

LouisTheLuis:
The original `SlurmLauncher` was calling `srun` with `bash -c ''` which caused some syntax bugs. Also fixed some of the variable names.

### DIFF
--- a/util/dvsim/SlurmLauncher.py
+++ b/util/dvsim/SlurmLauncher.py
@@ -7,6 +7,8 @@ import os
 import shlex
 import shutil
 import subprocess
+import tempfile
+import sys
 
 from Launcher import ErrorMessage, Launcher, LauncherError
 
@@ -31,6 +33,7 @@ class SlurmLauncher(Launcher):
         # Popen object when launching the job.
         self.process = None
         self.slurm_log_file = self.deploy.get_log_path() + '.slurm'
+        self.slurm_script_file = None
 
     def _do_launch(self):
         # replace the values in the shell's env vars if the keys match.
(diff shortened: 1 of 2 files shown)


util/dvsim/SlurmLauncher.py
```diff
@@ -31,6 +33,7 @@ def __init__(self, deploy):
         # Popen object when launching the job.
         self.process = None
         self.slurm_log_file = self.deploy.get_log_path() + '.slurm'
+        self.slurm_script_file = None
```
matutem, line 36: Could you locally use pathlib (and Path objects) instead of os? Same for slurm_log_file. In another PR is you wish.

util/dvsim/SlurmLauncher.py
```diff
@@ -44,25 +47,37 @@ def _do_launch(self):
...
+            with tempfile.NamedTemporaryFile(mode='w', suffix='.sh',
+                                             dir=script_dir, delete=False) as f:
+                f.write('#!/bin/bash\n')
+                if SLURM_SETUP_CMD:
+                    f.write(f'{SLURM_SETUP_CMD}\n')
+                f.write(f'{self.deploy.cmd}\n')
+                self.slurm_script_file = f.name
+            os.chmod(self.slurm_script_file, 0o755)
+        except OSError as e:
+            log.error(f"OSError: {e}\nCouldn't make directory {script_dir}")
```
matutem, line 66: Notice the write above could have also thrown an exception. To be cleaner you could put makedirs in a try block, and the write in another one.

matutem (approved):
With some simple suggestions

util/dvsim/SlurmLauncher.py
LouisTheLuis: Done.

util/dvsim/SlurmLauncher.py
LouisTheLuis: Done.

PR #108, opened 2026-04-17
[top_level,spi] Tighten width handling of csb and active_csb

### LINKED ISSUE #107: [chip,spi] Fix failure in chip_sw_spi_device_pass_through
### Description

This test fails for both earlgrey and darjeeling. The message is

* `UVM_ERROR @ * us: (spi_if.sv:74) [spi_if] Check failed $countones(NUM_CSB'(~csb)) <= * (* [*] vs * [*])` has 3 failures:
    * Test chip_sw_spi_device_pass_through has 3 failures.
        * 0.chip_sw_spi_device_pass_through.102004364901188594737673180229607246574949988651022080204912095529969100087635\
          [snip]
                UVM_ERROR @  10.220001 us: (spi_if.sv:74) [spi_if] Check failed $countones(NUM_CSB'(~csb)) <= 1 (2 [0x2] vs 1 [0x1])

Notice the spi_if.csb signal is declared with width spi_agent_pkg::NUM_CSB. When tracing this failure it is found that csb top two bits are not set to 1, as should be the case since they are unconnected.

This test fails for both darjeeling and earlgrey.

matutem:
The assignments to csb in chip_if seem to be handled incorrectly by the simulator. This explicitly uses the declared width of csb and active_csb to avoid any confusion.
Make sure csb is also initialized on initial reset.

Fixes #107

### DIFF
--- a/hw/dv/sv/spi_agent/spi_host_driver.sv
+++ b/hw/dv/sv/spi_agent/spi_host_driver.sv
@@ -18,11 +18,12 @@ class spi_host_driver extends spi_driver;
 
   // Resets signals
   virtual task reset_signals();
+    cfg.vif.csb = '1;
     forever begin
       @(negedge cfg.vif.rst_n or negedge cfg.vif.disconnected);
       if (cfg.vif.disconnected) continue;
       under_reset = 1'b1;
-      active_csb  = 1'b0;
+      active_csb  = CSB_WIDTH'(0);
       cfg.vif.sck <= cfg.sck_polarity[0];
       cfg.vif.sio_out <= 'x;
       cfg.vif.csb <= '1;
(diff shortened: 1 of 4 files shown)


hw/dv/sv/spi_agent/spi_host_driver.sv
```diff
@@ -18,11 +18,12 @@ class spi_host_driver extends spi_driver;
 
   // Resets signals
   virtual task reset_signals();
+    cfg.vif.csb = '1;
     forever begin
       @(negedge cfg.vif.rst_n or negedge cfg.vif.disconnected);
       if (cfg.vif.disconnected) continue;
       under_reset = 1'b1;
-      active_csb  = 1'b0;
+      active_csb  = CSB_WIDTH'(0);
```
ysekhar, line 26: My only concern is that if we have the opportunity to rewrite the driver to be reset safe we need to do it properly too.

ysekhar (approved):
LGTM in my view.

hw/dv/sv/spi_agent/spi_host_driver.sv
matutem: So we should :smiley:

PR #109, opened 2026-04-17
[ci] Allow workflow-dispatch in weekly for debug

LouisTheLuis:


qmn (approved):
LGTM, this also matches how the Nightly does it

PR #115, opened 2026-04-22
[dvsim] Including seeds through hjson

LouisTheLuis:
This PR contains 3 commits:
- The first one deals with a bug in the generation of batch simulation reports, resulting in the reports failing to include detailed coverage figures (e.g. Group Coverage, etc.).
- The second one updates the `grade_regression_vcs.py` script to also fetch the seeds for each corresponding test, thus allowing this to be passed to `dvsim.py` using the `--reseed-source` flag. It also renamed certain functions for the sake of clarity (as "override" is already used to described the function of the `-r` flag).
- The third one implements the inclusion of custom seeds into `dvsim.py`. Now, `dvsim.py` will be able to prioritize the seeds provided for each test either through the `--reseed-source` flag or through the main `sim_cfg`.


LouisTheLuis:
For a small example of how this works, let's consider `ac_range_check_sim_cfg.hjson`. This `sim_cfg` file has a tests section:

```
  // List of test specifications.
  tests: [
    {
      name: ac_range_check_smoke
      uvm_test_seq: ac_range_check_smoke_vseq
      reseed: 20
    }
    {
      name: ac_range_check_smoke_racl
      uvm_test_seq: ac_range_check_smoke_racl_vseq
      reseed: 20
    }
    {
      name: ac_range_check_smoke_high_threshold
      uvm_test_seq: ac_range_check_smoke_high_threshold_vseq
      reseed: 20
    }
    {
      name: ac_range_check_rst_safe_smoke
      uvm_test: ac_range_check_rst_safe_base_test
      uvm_test_seq: ac_range_check_rst_safe_smoke_vseq
      reseed: 20
    }
    {
      name: ac_range_check_bypass
      uvm_test_seq: ac_range_check_bypass_vseq
      reseed: 1
    }
    {
      name: ac_range_check_lock_range
      uvm_test_seq: ac_range_check_lock_range_vseq
      reseed: 20
    }

    // TODO: add more tests here
  ]
```

If we wanted the first 3 seeds of `ac_range_check_smoke` to be 1, 2, and 3 respectively, we now can define it as follows:

```
  // List of test specifications.
  tests: [
    {
      name: ac_range_check_smoke
      uvm_test_seq: ac_range_check_smoke_vseq
      reseed: 20
      tests: [
        "1"
        "2"
        "3"
      ]
    }
    {
      name: ac_range_check_smoke_racl
      uvm_test_seq: ac_range_check_smoke_racl_vseq
      reseed: 20
    }
    {
      name: ac_range_check_smoke_high_threshold
      uvm_test_seq: ac_range_check_smoke_high_threshold_vseq
      reseed: 20
    }
    {
      name: ac_range_check_rst_safe_smoke
      uvm_test: ac_range_check_rst_safe_base_test
      uvm_test_seq: ac_range_check_rst_safe_smoke_vseq
      reseed: 20
    }
    {
      name: ac_range_check_bypass
      uvm_test_seq: ac_range_check_bypass_vseq
      reseed: 1
    }
    {
      name: ac_range_check_lock_range
      uvm_test_seq: ac_range_check_lock_range_vseq
      reseed: 20
    }

    // TODO: add more tests here
  ]
```

The end result of this is that once `dvsim.py` runs `ac_range_check_smoke`, it will first run seeds `1`, `2`, and `3`, before running Python-generated randomized seeds for the 17 remaining reseeds. Similarly, we can "override" the seeds by calling `dvsim.py` with the `--reseed-source` flag and specifying the seeds there. An example can be seen in `grading_report.hjson`:

```
{
  reseed: 1
  tests:
  [
    {
      test: ac_range_check_alert_test
      name: ac_range_check
      reseed: 3
      seeds:
      [
        "1926663374011160319064726394148343593605046832920935114263797626204125776400"
      ]
      variant: darjeeling
    }
...
```

util/dvsim/CfgFactory.py
```diff
@@ -82,15 +82,15 @@ def _make_child_cfg(path, args, initial_values, seed_index=None, default_reseed=
...
 
     # Call cls as a constructor. Note that we pass None as the mk_config
     # argument: this is not supposed to load anything else.
     return cls(path, hjson_data, args, None)
 
 
-def reseed_override(hjson_data: dict, seed_index: dict, default_reseed: int):
-    '''Override resets for each test that has a reseed if the flag --reseed-source was passed.
+def reseed_source(hjson_data: dict, seed_index: dict, default_reseed: int):
+    '''Source resets for each test that has a reseed if the flag --reseed-source was passed.
```
matutem, line 93: I thought override made a much better name. As for the comment, it would be '''Override reseed for each ...'''. reseed_source is not that intuitive, perhaps it can be interpreted as reseed from the given source file?

hjson_data is also vague, this is the grading derived reseed count contained in hjson data format.

Perhaps the concept is graded_reseed(s)?

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,47 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
-def parse_test_file(path: Path) -> Counter:
+def parse_test_file(path: Path, unit_sim_dir: Path) -> tuple[Counter, dict[str, list[str]]]:
     """
     Parses the corresponding graded file and counts the instances of a test
     within that file.
     """
-    INDEX_PREFIX_RE = re.compile(r"^\d+\.")  # leading run index
-    SEED_SUFFIX_RE = re.compile(r"\.\d+$")   # trailing seed
+    INDEX_PREFIX_RE = re.compile(r"^(\d+)\.")    # leading run index
+    SEED_SUFFIX_RE = re.compile(r"\.(\d+)$")   # VCS run index
```
matutem: I thought this was the seed VCS uses, so `# VCS seed`?

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,47 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
-                name = token.rsplit("/", 1)[-1]
-                name = INDEX_PREFIX_RE.sub("", name, 1)
+                snippet = token.rsplit("/", 1)[-1]
+
+                idx_m = INDEX_PREFIX_RE.match(snippet)
+                run_idx = idx_m.group(1) if idx_m else None
+
+                name = INDEX_PREFIX_RE.sub("", snippet, 1)
                 name = SEED_SUFFIX_RE.sub("", name)
+
```
matutem, line 126: Check that `name` is not empty.

hw/dv/grading/grading_report.hjson
```diff
@@ -10,6350 +10,14445 @@
       test: ac_range_check_alert_test
       name: ac_range_check
       reseed: 3
```
ysekhar, line 12: Question: What happens when reseed is set and seed is also set? 
My take is in this scenarios is that reseed will run 3  tests but seed will always be the same? 
Or will it use us the seed in the list first and then run with random seed ?

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,47 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
                 counts[name] += 1
+
+                # Find seed passed by dvsim
+                if unit_sim_dir is not None and run_idx is not None:
+                    prefix = f"{run_idx}.{name}"
+                    for match in unit_sim_dir.glob(f"{prefix}*"):
+                        if match.is_dir():
+                            long_seed = match.name[len(prefix) + 1:]
```
matutem: I am confused here, I think long seed should be the last token in the snippet. It may be clearer if you  extract it with SEED_SUFFIX_RE from the snippet.

util/grade_regression_vcs.py
```diff
@@ -260,8 +276,12 @@ def main() -> int:
             missing.append((unit, "gradedtests.txt missing"))
             continue
 
-        graded = parse_test_file(graded_path)
-        unique = parse_test_file(unique_path) if unique_path.is_file() else Counter()
+        unit_sim_dir = cov_merge.parent / "passed"
```
matutem: unit_sim_dir -> passing_tests_dir?

hw/dv/grading/grading_report.hjson
LouisTheLuis: Within a sim_cfg, `reseed` only defines the number of times that a test is ran with separate seeds. So what would happen with `reseed: 6; seeds: ["1", "2", "3"]` is that the first three seeds would be popped (running 1, 2 and 3, respectively) and then three new random seeds will be generated and run.

If what you're asking is what would happen with the `-s` or `-fs` flags, look at the `get_seed()` method from `RunTest`:

```
    def get_seed(seedlist: List[str]):
        # If --seeds option is passed, then those custom seeds are consumed
        # first. If --fixed-seed <val> is also passed, the subsequent tests
        # (once the custom seeds are consumed) will be run with the fixed seed.
        if not RunTest.seeds:
            if RunTest.fixed_seed is not None:
                return RunTest.fixed_seed
            for i in range(1000):
                seed = random.getrandbits(256)
                RunTest.seeds.append(seed)
        # If the test already specifies some seeds, use those first
        if len(seedlist) != 0:
            return int(seedlist.pop(0))
        return RunTest.seeds.pop(0)
```

The seed provided by `-fs` takes priority; then, it's the seeds provided by the `hjson`, the seeds provided by the `-s` flag, and finally, the randomized seeds. Probably a better idea would be to give the ones provided by `-s` priority (by appending `seedlist` at the end).

util/dvsim/CfgFactory.py
LouisTheLuis: The reason why I changed from `override` to `source` is because override is already used to describe what `-s` and `-fs` do within `FlowCfg.py` and `SimCfg.py`. It seems like a bit of semantic overload, so I would prefer to keep the definitions separate.

Modified to `graded_reseeds`.

util/grade_regression_vcs.py
LouisTheLuis: Fixed.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: This would imply a directory like `{digit}.{seed}` or one not following any pattern.
Added a warning.

util/grade_regression_vcs.py
LouisTheLuis: To clarify: the numbers at the end of the snippet are the short seed (the one VCS stores). The long seeds show up in the names of the directories inside `/passed`. It's also ideal to use the long seeds, to keep some level of consistency (otherwise, the hardware seeds will be the same but the software ones will differ).

The regex doesn't seem to work as well because there are some tests that have redundant names (e.g. `xbar_access_same_device` and `xbar_access_same_device_slow_rsp`), so it will pick up the wrong seed (as opposed to taking the slice, which would fail in that scenario as it would pick a string and `long_seed.is_digit() = False`.

util/grade_regression_vcs.py
matutem: These very long functions are always problematic. Could you try to encapsulate the important steps here in different functions, and pass the objects needed as arguments? That will also illuminate the data flow here.

It is okay if you do it in a separate commit.

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,51 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
                 name = SEED_SUFFIX_RE.sub("", name)
+
+                if not name:
+                    log.warning(f"Skipping malformed entry: {snippet}")
+                    continue
+
                 counts[name] += 1
+
+                # Find seed passed by dvsim
+                if passing_tests_dir is not None and run_idx is not None:
```
matutem: This should really be an error, since you won't get the right seeds, and the generated output will be wrong.

util/grade_regression_vcs.py
matutem: I understand: this code is doing an indirection using the index and the test name into the `passing` directory to extract the full seed.

Imagine this code being in a function: the function name, signature, and doc should explain this to all future readers of this code.

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,51 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
-                name = INDEX_PREFIX_RE.sub("", name, 1)
+                snippet = token.rsplit("/", 1)[-1]
+
+                idx_m = INDEX_PREFIX_RE.match(snippet)
+                run_idx = idx_m.group(1) if idx_m else None
+
+                name = INDEX_PREFIX_RE.sub("", snippet, 1)
                 name = SEED_SUFFIX_RE.sub("", name)
+
+                if not name:
```
matutem:
```suggestion
                if not name or run_idx is None:
```

util/grade_regression_vcs.py
LouisTheLuis: Done, did it within the same commit for the sake of keeping it in one place.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Added an else statement with an error.

util/grade_regression_vcs.py
LouisTheLuis: Refactored the function to make it more clear.

util/grade_regression_vcs.py
matutem, line 309: We should also error or at least warn if `graded` is empty.

util/grade_regression_vcs.py
```diff
@@ -89,31 +89,51 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
+                    continue
+
                 counts[name] += 1
+
+                # Find seed passed by dvsim
+                if passing_tests_dir is not None and run_idx is not None:
+                    prefix = f"{run_idx}.{name}"
+                    for match in passing_tests_dir.glob(f"{prefix}*"):
+                        if match.is_dir():
+                            long_seed = match.name[len(prefix) + 1:]
```
matutem: Perhaps add a check for long_seed.bit_count()? We expect it to be close enough to 256, the probability of being much smaller drops rapidly.

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
     return proc.returncode, time.time() - start
 
 
-def parse_test_file(path: Path) -> Counter:
+def parse_test_file(path: Path, passing_tests_dir: Path) -> tuple[Counter, dict[str, list[str]]]:
```
matutem: Could you update the function docstring? Also, could you name the first arg `tests_file`?

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
     """
     Parses the corresponding graded file and counts the instances of a test
     within that file.
     """
-    INDEX_PREFIX_RE = re.compile(r"^\d+\.")  # leading run index
-    SEED_SUFFIX_RE = re.compile(r"\.\d+$")   # trailing seed
+    INDEX_PREFIX_RE = re.compile(r"^(\d+)\.")    # leading run index
+    SEED_SUFFIX_RE = re.compile(r"\.(\d+)$")     # VCS seed
     counts: Counter = Counter()
+    seeds: dict[str, list[str]] = {}
```
matutem: from collections also import defaultdict and change as

```suggestion
    seeds = defaultdict(list)
```

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
     if not path.is_file():
```
matutem: Why return garbage on an error? Could you also check that passing_tests_dir is_dir()?

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
+                    continue
+
                 counts[name] += 1
+
+                # Find seed passed by dvsim
+                seed = _find_long_seed(passing_tests_dir, run_idx, name)
+                if seed is None:
+                    log.error("Failed to obtain long seed for %s (run %s)", name, run_idx)
+                    sys.exit(1)
+                seeds.setdefault(name, []).append(seed)
```
matutem: If you followed the comment above about defaultdict this would be changed to

```suggestion
                seeds[name].append(seed)
```

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
     counts: Counter = Counter()
+    seeds: dict[str, list[str]] = {}
     if not path.is_file():
-        return counts
+        return counts, seeds
     try:
         with path.open() as fh:
             for raw in fh:
                 line = raw.strip()
                 if not line or line.startswith("#") or line.startswith("//"):
```
matutem, line 114: For each line we need to extract the index and the test name, and using the index we should succeed getting the full seed. Otherwise we should cause an error with enough info to understand what failed.

If we were sticky we could also get the short seed and check that it matches the tail of the long seed.

This suggests a single re, like r".*/(\d+)\.(\w+)\.(\d+)$", error (or warn and skip) if the match doesn't have 3 groups.

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
+    We can't use SEED_SUFFIX_RE here because test names can overlap
+    (e.g. xbar_access_same_device vs xbar_access_same_device_slow_rsp),
+    so we rely on .isdigit() to isolate the seed suffix unambiguously.
+    """
+    prefix = f"{run_idx}.{name}"
+
+    for entry in passing_tests_dir.glob(f"{prefix}*"):
+        if entry.is_dir():
+            long_seed = entry.name[len(prefix) + 1:]
+            if long_seed.isdigit():
```
matutem:
```suggestion
            return long_seed if long_seed.isdigit() else None
```

util/grade_regression_vcs.py
```diff
@@ -89,31 +92,64 @@ def run_urg_grade(unit: str, merged: Path, report: Path, dry_run: bool) -> tuple
...
+                    log.warning(f"Skipping malformed entry: {snippet}")
+                    continue
+
                 counts[name] += 1
+
+                # Find seed passed by dvsim
+                seed = _find_long_seed(passing_tests_dir, run_idx, name)
+                if seed is None:
+                    log.error("Failed to obtain long seed for %s (run %s)", name, run_idx)
+                    sys.exit(1)
```
matutem: Could you also add some check using int(seed).bit_count()? I think the probability of getting 256 - d bits drops exponentially with d. Perhaps warn id d > <maybe 16>

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: Actually, I added a stronger check:
```
def _find_long_seed(passing_tests_dir: Path, run_idx: str, 
                    name: str, short_seed: str) -> str | None:
    """Look up the 256-bit seed from the passed/ directory.
    We can't use SEED_SUFFIX_RE here because test names can overlap
    (e.g. xbar_access_same_device vs xbar_access_same_device_slow_rsp),
    so we rely on .isdigit() to isolate the seed suffix unambiguously.
    """
    prefix = f"{run_idx}.{name}"

    for entry in passing_tests_dir.glob(f"{prefix}*"):
        if entry.is_dir():
            long_seed = entry.name[len(prefix) + 1:]
            if long_seed.isdigit() and (int(long_seed) & 0xFFFFFFFF) == (int(short_seed) & 0xFFFFFFFF):
                return long_seed
    return None
```

util/grade_regression_vcs.py
LouisTheLuis: Done.

util/grade_regression_vcs.py
LouisTheLuis: This can be done with the `_find_long_seed` function erroring and then having this check:
```
                idx_m = INDEX_PREFIX_RE.match(snippet)
                seed_m = SEED_SUFFIX_RE.search(snippet)
                short_seed = seed_m.group(1) if seed_m else None
                run_idx = idx_m.group(1) if idx_m else None

                name = INDEX_PREFIX_RE.sub("", snippet, 1)
                name = SEED_SUFFIX_RE.sub("", name)

                if not name or run_idx is None or short_seed is None:
                    log.warning(f"Skipping malformed entry: {snippet}")
                    continue
```

In `_find_long_seed`:

```
    # If a seed is not found, error out
    candidate_names = [e.name for e in candidates] or ["(none)"]
    log.error(
        "Failed to obtain long seed for %s.%s.%s\n  Candidates:\n    %s",
        run_idx, name, short_seed, "\n    ".join(candidate_names)
    )
    sys.exit(1)
```

matutem (approved):
LGTM

PR #116, opened 2026-04-23
[doc] Add ACC threat model to docs

pqcfox:
Added threat model as written primarily covers SCA details.

### DIFF
--- a/doc/security/threat_model/acc_sca_threat_model.md
+++ b/doc/security/threat_model/acc_sca_threat_model.md
@@ -1,0 +1,81 @@
+# Asymmetric Cryptographic Coprocessor (ACC) Side Channel Analysis (SCA) Threat Model
+
+This document details the side-channel analysis (SCA) threat model for the Asymmetric Cryptographic Coprocessor (ACC), as utilized by the *cryptolib* cryptographic library to implement classical and post-quantum cryptography in the Pavona silicon root of trust (RoT) designs.
+
+## Context
+
+ACC overview: The Asymmetric Cryptographic Coprocessor is the programmable cryptographic coprocessor included in Pavona’s RoT designs.
+The ACC itself provides acceleration for operations including:
+
+* RSA-2048, RSA-3072, and RSA-4096 keygen
+* RSA-2048, RSA-3072, and RSA-4096 PKCS v1.5 signatures
+* RSA-2048, RSA-3072, and RSA-4096 PSS signatures
+* RSA-2048, RSA-3072, and RSA-4096 OAEP encryption
+* ECDSA with NIST P-256 and NIST P-384 curves
+* ECDH with NIST P-256 and NIST P-384 curves
+* Ed25519
+* X25519
+* ML-KEM-{512,768,1024}
+* ML-DSA-{44,65,87}
+
+At its core, the ACC consists of a highly customized lightweight processor containing a 32-bit general purpose register bank and a separate 256-bit wide register bank with corresponding bignum instructions.
+By reusing these wide registers as vector registers, a vectorized ISA is implemented without introducing an additional bank, allowing for performant and area-efficient implementation of various post-quantum cryptographic algorithms, especially ML-KEM and ML-DSA.
+
+ACC programs are loaded by the primary RoT core into an instruction memory (IMEM), and inputs and outputs are passed to/from the primary RoT core via the data memory (DMEM).
+A small section of DMEM is reserved as scratchpad memory for the DMEM, such that the primary RoT core cannot access anything placed in that section.
+
+To allow for hardware-backed keys which are never exposed to the primary RoT core, a sideloading datapath from the key manager to the ACC can be used.
+Moreover, to support fast ML-KEM and ML-DSA implementations with hardware-backed keys, a direct interface from the ACC to the fixed KMAC/SHA-3/SHAKE engine is provided.
+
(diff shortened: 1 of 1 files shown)


doc/security/threat_model/acc_sca_threat_model.md
```diff
@@ -0,0 +1,62 @@
+# Asymmetric Cryptographic Coprocessor (ACC) Side Channel Analysis (SCA) Threat Model
+
+This document details the side-channel analysis (SCA) threat model for the Asymmetric Cryptographic Coprocessor (ACC), as utilized by the *cryptolib* cryptographic library to implement classical and post-quantum cryptography in the Pavona silicon root of trust (RoT) designs.
+
+## Introduction
```
jadephilipoom: nit: since the content of this section is more background info about what ACC is than about the threat model itself, I'd call it `Context`.

doc/security/threat_model/acc_sca_threat_model.md
```diff
@@ -0,0 +1,62 @@
...
+* Ed25519
+* X25519
+* ML-KEM-{512,768,1024}
+* ML-DSA-{44,65,87}
+
+At its core, the ACC consists of a highly customized lightweight processor containing a 32-bit general purpose register bank and a separate 256-bit wide register bank with corresponding bignum instructions. By reusing these wide registers as vector registers, a vectorized ISA is implemented without introducing an additional bank, allowing for performant and area-efficient implementation of various post-quantum cryptographic algorithms, especially ML-KEM and ML-DSA.
+
+ACC programs are loaded by the primary RoT core into an instruction memory (IMEM), and inputs and outputs are passed to/from the primary RoT core via the data memory (DMEM). A small section of DMEM is reserved as scratchpad memory for the DMEM, such that the primary RoT core cannot access anything placed in that section.
+
+To allow for hardware-backed keys which are never exposed to the primary RoT core, a sideloading datapath from the key manager to the ACC can be used. Moreover, to support fast ML-KEM and ML-DSA implementations with hardware-backed keys, a direct interface from the ACC to the fixed KMAC/SHA-3/SHAKE engine is provided.
```
jadephilipoom: A common confusion that people have about the ACC threat model is the seeming contradiction between not trusting the main processor with keymgr keys, and simultaneously receiving the IMEM input (which can use the keys) from the main processor, which implies some level of trust. Sometimes people even think that perhaps the attacker could specify ACC programs without kernel compromise (which should not be the case). It might be worth adding something around here to clarify this. From my perspective, the reasoning is essentially that although we are not protected against an attacker who has _complete, fine-grained_ control over ACC or the kernel running on the main processor, we want to have some protection against attackers who have some _incomplete_ or _coarse-grained_ control. The assumption is that tricking or glitching the main processor into leaking a few bytes of a key somewhere in the midst of a computation requires less precision than glitching the main processor into loading a very specific malicious program onto ACC and pressing go.

doc/security/threat_model/acc_sca_threat_model.md
```diff
@@ -0,0 +1,62 @@
...
+
+Active side-channel mitigations: As noted above, a dual-core lockstep ACC implementation is the primary recommended approach for mitigating active attacks. This said, there are several defense-in-depth mechanisms also employed, including PRINCE scrambling for ACC memories, running hardware checksums for DMEM writes, and hardened runtime comparisons of ACC instruction counts to statically-determined runtime bounds.
+
+## Proposed Empirical SCA Methodology
+
+Planned methodology: For initial assessment and continuous integration (CI) evaluation of side-channel leakage, standard TVLA methodology (fixed vs. random Welch’s t-test) will be used with a ChipWhisperer CW310 Kintex K410T FPGA board as target, testing isolated routines (pointwise multiplication, NTT/INTT, message decoding, etc.) as well as full instantiations of ML-KEM, e.g. a plaintext-checking oracle. More extensive follow-up analyses will be performed using detailed manual inspection of traces, including evaluation of techniques from prominent recent attacks, e.g. [HNP+24](https://eprint.iacr.org/2024/060)’s practical attacks on high-order masked comparison gadgets.
```
jadephilipoom: Only ML-KEM?

jadephilipoom (approved):
Looks good, left a few comments but feel free to take them or leave them!

doc/security/threat_model/acc_sca_threat_model.md
pqcfox: Updated to be more general, thanks!

doc/security/threat_model/acc_sca_threat_model.md
pqcfox: Good call, changed the title of this section.

doc/security/threat_model/acc_sca_threat_model.md
pqcfox: This makes sense, adding something to this effect now :)

qmn (commented):
Hi, thanks for putting this together! My only comment is on presentation: could you un-flow to have one sentence per line? Thanks!

qmn:
Found some spare cycles to make it one sentence per line. Hope this looks fine!

pqcfox:
Thanks so much! Added last small fix, should be set to go.

qmn (approved):
LGTM, thanks! @pqcfox could you please (manually) squash and then merge this commit?

pqcfox:
Manually squashed as requested, will merge once the CI w/ lint checks etc. pass :)

PR #118, opened 2026-04-24
[ci] Remove GCP authentication from prepare-env

LouisTheLuis:


rtorok-zr (approved):
LGTM for now. I'm in favor of merging this now because it will unblock scheduled regressions, but we will likely be able to revert much of this once the cloud buckets are set up.

PR #122, opened 2026-04-27
[util,gen_system_list] Sort the tops

matutem:
The tops must be in sorted order, otherwise the resulting table is not stable since glob will return paths in any order.


vlpannel (approved):
LGTM; thanks for the catch!

PR #125, opened 2026-04-27
[dvsim] Change logging for cssutils

LouisTheLuis:
CSS logging should depend on the `--verbose` flag.

The `cssutils` within `premailer` marks certain CSS3 properties as errors through its validator. Examples such as the use of `rem` or the `overflow: clip`, create a stream of error messages within the `dvsim` flow.
The properties are included inline, for which reason the pages work fine.

As disabling the validator in `cssutils` would not be ideal (we still want to see errors if they exist), and we also do not want pages that compile properly to be treated as errors, this PR sets the `cssutils` messages under the `--verbose debug` flag.

### DIFF
--- a/util/dvsim/dvsim.py
+++ b/util/dvsim/dvsim.py
@@ -28,6 +28,7 @@ import shlex
 import subprocess
 import sys
 import textwrap
+import cssutils
 from pathlib import Path
 
 import Launcher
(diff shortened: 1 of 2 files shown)


matutem:
Does this means potentially more important css errors won't be visible? If so, could you create an issue so we can undo this change at some point?

matutem (approved):
I left a suggestion about filing an issue.

LGTM

PR #126, opened 2026-04-28
[ci] Reintegrate `Build Documentation` CI job with PR checks

rtorok-zr:
This PR migrates the `Build Documentation` CI job back to the standard set of per-PR checks, rather than running it as a separate workflow.


vlpannel (approved):
LGTM (once the commit lint error is fixed).

PR #127, opened 2026-04-29
[alert_handler,testutils] Fix parsing of alert info

matutem:
Fix two bug in alert_handler_testutils:
- Handle the size of the alert info being a multiple of 32: it was incorrectly declaring an error if the word index was equal to the dump_size, though that is okay when the bit index is 0. This happens to be the case for darjeeling, where the alert dump size is 288 bits (8 * 32).
- Fix the extraction of a word when its bits span two consecutive words.


qmn (approved):
Nice sleuthing! LGTM

Issue #44, opened 2026-03-25
[top_dragonfly,dv] Chip level dvsim tests don't have an existing SW image Bazel target


apinise:
There are tests in `hw/top_dragonfly/dv/chip_sim_cfg.hjson` that fail to build because the `sw_images` they define are not valid targets in `sw/device/tests/BUILD`. To prevent build issues these tests have been temporarily commented out, but are tracked below. Tests should be marked off from the list as they are fixed, or deemed unnecessary.

#### Affected dvsim tests from `hw/top_dragonfly/dv/chip_sim_cfg.hjson`

- [x] chip_sw_lc_ctrl_otp_hw_cfg
- [ ] chip_sw_otp_ctrl_nvm_cnt
- [ ] chip_sw_otp_ctrl_sw_parts
- [ ] chip_sw_alert_test
- [ ] chip_sw_alert_handler_lpg_sleep_mode_alerts
- [ ] chip_sw_sram_ctrl_scrambled_access_jitter_en
- [ ] chip_sw_coremark
- [ ] chip_sw_rv_dm_ndm_reset_req_when_cpu_halted
- [ ] chip_sw_rv_dm_access_after_wakeup

#### Affected dvsim tests from `hw/top_dragonfly/dv/chip_rom_tests.hjson`

- [ ] chip_sw_uart_smoketest_signed
- [ ] rom_e2e_smoke
- [ ] rom_e2e_shutdown_exception_c
- [ ] rom_e2e_shutdown_output
- [ ] rom_e2e_boot_policy_valid_a_good_b_good_test_unlocked0
- [ ] rom_e2e_boot_policy_valid_a_good_b_good_dev
- [ ] rom_e2e_boot_policy_valid_a_good_b_good_prod
- [ ] rom_e2e_boot_policy_valid_a_good_b_good_prod_end
- [ ] rom_e2e_boot_policy_valid_a_good_b_good_rma
- [ ] rom_e2e_boot_policy_valid_a_good_b_bad_test_unlocked0
- [ ] rom_e2e_boot_policy_valid_a_good_b_bad_dev
- [ ] rom_e2e_boot_policy_valid_a_good_b_bad_prod
- [ ] rom_e2e_boot_policy_valid_a_good_b_bad_prod_end
- [ ] rom_e2e_boot_policy_valid_a_good_b_bad_rma
- [ ] rom_e2e_boot_policy_valid_a_bad_b_good_test_unlocked0
[...]

PR #133, opened 2026-04-30
[top_darjeeling,dv] Remove rom_e2e_sigverify_mod_exp dvsim tests

apinise:
This PR removes the `rom_e2e_sigverify_mod_exp` tests from `hw/top_darjeeling/dv/chip_rom_tests.hjson`. These tests are no longer supported after the system software switched from RSA to ECDSA code signings and removed its collateral. The tests were originally removed in PR: https://github.com/lowRISC/opentitan/pull/22696

The tests were tracked in #44 and have been marked as resolved.

### DIFF
--- a/hw/top_darjeeling/dv/chip_rom_tests.hjson
+++ b/hw/top_darjeeling/dv/chip_rom_tests.hjson
@@ -752,146 +752,6 @@
       ]
       run_timeout_mins: 120
     }
-    {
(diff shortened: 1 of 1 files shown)


matutem (approved):
Evan, could you modify the PR description? It has this snippet: `after earlgrey switched from RSA`. Could you make it more generic? Perhaps saying `after the system software switched from RSA`.

Otherwise LGTM. Thanks.

PR #134, opened 2026-04-30
[ci] Refactor CI job orchestration

rtorok-zr:
- Migrates the documentation and bitstream uploads to a `merge-group` job that runs immediately when a Pull Request is merged, instead of waiting for the Nightly CI to do this. This partially reverts #126 by no longer running an artifact upload for the documentation (which was being redundantly run in `slow_lint`) on PR checks.
- Refactors the dispatch of the FPGA tests to avoid duplicating the bitstream builds and FPGA test jobs between PR/Nightly CI, while preserving the property that the Nightly FPGA tests run even if an on-demand check fails.
- Makes the FPGA test naming more consistent.


.github/workflows/ci.yml
```diff
@@ -318,59 +293,48 @@ jobs:
...
+  chip_earlgrey_cw310:
+    name: Earlgrey for CW310
     needs: quick_lint
-    if: github.event_name == 'pull_request'
     uses: ./.github/workflows/bitstream.yml
     secrets: inherit
     with:
       top_name: earlgrey
       design_suffix: cw310_hyperdebug
-      vivado_version: '2024.1'
```
vlpannel, line 330: I'm missing context here: why are we renaming the CW310-with-hyperdebug test to just CW310?

vlpannel: There are a couple more instances of this same change in this PR.

vlpannel (approved):
Just one simple question, but overall LGTM.

.github/workflows/ci.yml
rtorok-zr: There used to be separate bitstream builds for CW310 and CW310+HyperDebug, but the former was removed in upstream OT in mid-2025, with all the software tests migrated to use the HyperDebug bitstream. Since we now _always_ use the HyperDebug (this is the case for the CW340 as well), we do not need to specify it explicitly and I decided to make the naming more consistent with the other board(s).

vlpannel: This makes sense; now-unequivocal LGTM.

PR #135, opened 2026-04-30
[dvsim,ci] Allow for --publish-mode flag

LouisTheLuis:
This flag, `--publish-mode` defines what elements get published into the corresponding repository. It defaults to `public`, meaning, it only shows the generated reports and no proprietary data. With `private` mode, it will copy the reports, as well as the coverage and knowledge database for useful triaging.

A change is included to CI to allow for two publishes: one to a public-facing repository, and one to the internal repository. This is done in tandem with a change to `grade_regression_vcs.py` such that we can do regrading before running the `weekly`.


util/dvsim/FlowCfg.py
```diff
@@ -497,8 +497,16 @@ def _get_results_page_link(self, relative_to, link_text=''):
                                         relative_to)
         return "[%s](%s)" % (link_text, relative_link)
 
-    def publish_results(self, repository: str, test: str):
-        """Publish these results to a corresponding repository."""
+    def publish_results(self, repository: str, test: str, mode: str):
```
matutem, line 502: Please follow multi-line docstrigs guidelines: https://peps.python.org/pep-0257/#multi-line-docstrings: short one-line description, followed by space and then the details. The detailed description is aligned with the start of the triple quotes. For example, see https://github.com/pavona/pavona-staging/blob/main/util/ipgen/clkmgr_gen.py.

LouisTheLuis: Done.

LouisTheLuis:
_Important Note:_ I will remove the last commit ("tests") once the CI is up and running; then I will run the Weekly with the "reduced" version of the `all_sim_cfgs.hjson` to verify that it can work.

**Update:** Weekly#38 and Weekly#39 are passing. The last commit has been removed and it's ready for review.

util/dvsim/utils.py
```diff
@@ -619,6 +619,31 @@ def mk_symlink(path, link):
...
+    to handle this — depending on the situation, a missing source may be
+    fatal or merely warrant a warning.
+    '''
+    src = Path(src)
+    dest = Path(dest)
+
+    if not src.exists():
+        raise FileNotFoundError(f"Source path does not exist: {src}")
+
+    dest.parent.mkdir(parents=True, exist_ok=True)
```
matutem, line 640: I see, no need for try-except because with those args mkdir doesn't fail.

util/dvsim/SimCfg.py
```diff
@@ -554,6 +558,32 @@ def cov_unr(self):
...
+
+        seeds_dir = Path(self.scratch_path) / "seeds"
+        latest_dir = seeds_dir / "latest"
+
+        clean_odirs(odir=latest_dir, max_odirs=89)
+        mk_path(latest_dir)
+
+        manifest_path = latest_dir / "seeds.txt"
+        try:
+            with open(manifest_path, "w") as f:
```
matutem, line 579: This is where we can also separate per run status as available (pass, uvm fail, timeout, build failure, ...). Remember I was asking about the list of failed tests?

LouisTheLuis: Modified it a little bit to get the list per status available ("P", "F", "K").

util/dvsim/FlowCfg.py
```diff
@@ -701,5 +701,113 @@ def _write_index_html(self, repo_dir: str, title: str, filename: str, version: s
...
+        """
+        item_dest_root = os.path.join(repo_dir,
+                                      f"{self.name}_{test}_{self.flow}",
+                                      f"{self.name}_{test}_{self.flow}_{self.timestamp}",
+                                      os.path.basename(item.scratch_path))
+
+        cov_src = os.path.join(item.scratch_path, 'cov_report')
+        cov_dest = os.path.join(item_dest_root, 'cov_report')
+        try:
+            cp_path(cov_src, cov_dest)
```
qmn, line 719: With this and the three other uses of `cp_path`, is there a case where `cov_report`, `seeds/latest`, the merged_src, and kdb names change between being a file or directory from one run to the next?

It seems that `cp_path` is determining whether the input is a directory or a file where the appropriate choice between `shutil.copy2()` or `shutil.copytree()` should be sufficient.

util/dvsim/FlowCfg.py
```diff
@@ -701,5 +701,113 @@ def _write_index_html(self, repo_dir: str, title: str, filename: str, version: s
...
+                    # multiple matches (e.g. one per build_mode) don't collide.
+                    build_subdir = kdb_src.parent.parent.name
+                    kdb_dest = Path(cov_data_dest) / build_subdir / kdb_name
+                    cp_path(kdb_src, kdb_dest)
+                    log.info(f"Copied knowledge database "
+                             f"{build_subdir}/{kdb_name} for {item.name}")
+        else:
+            log.warning(f"No knowledge database for {item.name}. Current support only for VCS.")
+
+    def _strip_native_cov_link(self, html_path: Path):
```
qmn, line 771: This is kind of a lot of code just to get rid of a hyperlink (it would be better to prevent emitting the link in the first place) but given the current DVSim architecture, this is the best way to get the report up for now.

util/grade_regression_vcs.py
```diff
@@ -41,6 +41,8 @@
 LICENSE_HEADER = ("Copyright zeroRISC Inc.\n"
                   "Licensed under the Apache License, Version 2.0, see LICENSE for details.\n"
                   "SPDX-License-Identifier: Apache-2.0\n")
+INDEX_PREFIX_RE = re.compile(r"^(\d+)\.")    # leading run index
+SEED_SUFFIX_RE = re.compile(r"\.(\d+)$")     # VCS seed
```
qmn: If tests are named `[index].[name].[vcs seed]`, I think you could combine this into one regex:

```suggestion
RUN_RE = re.compile(r"^(\d+)\.(.*)\.(\d+)$")
```

run `RUN_RE.match()`, then use `.group(1)` to get the index prefix, `.group(2)` to get the test name, and `.group(3)` to get the VCS seed.

Or, well,

```
run_idx, test_name, long_seed = qual.partition(".")
```
with the appropriate try...except block.

qmn (commented):
Looks really good so far! Just a few questions.

util/dvsim/FlowCfg.py
LouisTheLuis: Nope, they should always be directories.

util/dvsim/FlowCfg.py
LouisTheLuis: Agreed; ideally, it would be done at runtime. The problem is, we want the flexibility of running `dvsim` once and publishing both privately and publicly. If done at runtime, this would make it so that we have to define the publishing mode with the run (which is not the plan).

qmn (approved):
Thanks! In the interest of getting this dashboard run going, and because I don't consider my comments blocking anyway, let's get this in. Please do consider them (either here, or in a following PR).

EDIT: looks like you also just pushed some updates. The usual caveats apply (fix lint; the other failures appear unrelated).

util/grade_regression_vcs.py
LouisTheLuis: Refactored using `qual.partition(".")` and using a `try...except` block.

.github/actions/dvsim/action.yml
```diff
@@ -97,32 +132,47 @@ runs:
           MP=${{ inputs.concurrent-jobs }}
         fi
         export FUSESOC_CORES=${{ inputs.fusesoc-core-path }}
-        DASH_REPO="${{ inputs.dashboard }}"
-        PUBLISH=""
+        DASH_PUBLIC="${{ inputs.dashboard-public }}"
```
matutem: I don't see where the decision is made between public and private. Is it just that the inputs ends up setting one of the DASH_ variables to true? Ah, you are probably publishing both public and private for each regression. And the same for the dashboards?

LouisTheLuis: There's no decision between public and private; it just publishes to both:
~~~
        util/dvsim/dvsim.py $COVERAGE \
          ${{ inputs.hjson }} \
          -t ${{ inputs.tool }} \
          -mp $MP \
          -i ${{ inputs.item }} ${{ inputs.extra-dvsim-flags }} \
          ${PUBLISH_PRIVATE} ${DASH_PRIVATE} \
          ${MODE} \
          ${RESEED} \
          || ! ${{ inputs.require-pass }}

        ssh-add -D
        printf '%s\n' "${{ inputs.public-upload-key }}" | ssh-add -

        util/dvsim/dvsim.py \
          ${{ inputs.hjson }} \
          -t ${{ inputs.tool }} \
          -i ${{ inputs.item }} \
          ${PUBLISH_PUBLIC} ${DASH_PUBLIC}
~~~

It's set up so that the `inputs.publish` variable is the only decision between publishing or not.

.github/workflows/dv-all.yml
```diff
@@ -69,7 +72,9 @@ jobs:
         item: ${{ inputs.item }}
         coverage: ${{ inputs.coverage }}
         concurrent-jobs: ${{ inputs.concurrent-jobs }}
-        dashboard: ${{ inputs.dashboard }}
+        dashboard-private: ${{ inputs.dashboard-private }}
+        dashboard-public: ${{ inputs.dashboard-public }}
         publish: ${{ inputs.publish }}
         reseed-source: ${{ inputs.reseed-source }}
         dashboard-upload-key: ${{ secrets.DASHBOARD_UPLOAD_KEY }}
+        public-upload-key: ${{ secrets.PUBLIC_UPLOAD_KEY }}
```
matutem, line 76: This is a secret key? Is it adequately protected, it should not show up in any log for example.

.github/actions/dvsim/action.yml
LouisTheLuis: Refactored to make it simpler: now we have a single bash variable for each:
~~~
    - name: Run DVSim
      id: dvsim
      if: ${{ inputs.run-type != 'quick' || steps.should_rerun.outputs.SHOULD_RERUN == '1' }}
      run: |
        COVERAGE=""
        if [ "${{ inputs.coverage }}" == "true" ]; then
          COVERAGE="--cov"
        fi
        MP=$(nproc)
        if [ "${{ inputs.concurrent-jobs }}" != "0" ]; then
          MP=${{ inputs.concurrent-jobs }}
        fi
        export FUSESOC_CORES=${{ inputs.fusesoc-core-path }}
        PUBLISH_PRIVATE=""
        PUBLISH_PUBLIC=""
        if [ -n "${{ inputs.dashboard-private }}" ]; then
          PUBLISH_PRIVATE="--publish ${{ inputs.dashboard-private }} --publish-mode private"
        fi
        if [ -n "${{ inputs.dashboard-public }}" ]; then
          PUBLISH_PUBLIC="--publish-prev ${{ inputs.dashboard-public }}"
        fi
        RESEED=""
        if [ -f ./grading_report.hjson ]; then
          RESEED="-rs ./grading_report.hjson"
        elif [ -n "${{ inputs.reseed-source }}" ]; then
          RESEED="-rs ${{ inputs.reseed-source }}"
        fi

        util/dvsim/dvsim.py $COVERAGE \
          ${{ inputs.hjson }} \
          -t ${{ inputs.tool }} \
          -mp $MP \
          -r 1 \
          -i ${{ inputs.item }} ${{ inputs.extra-dvsim-flags }} \
          ${PUBLISH_PRIVATE} \
          ${RESEED} \
          || ! ${{ inputs.require-pass }}

        if [ -n "$PUBLISH_PUBLIC" ]; then
          ssh-add -D
          printf '%s\n' "${{ inputs.public-upload-key }}" | ssh-add -

          util/dvsim/dvsim.py \
            ${{ inputs.hjson }} \
            -t ${{ inputs.tool }} \
            -i ${{ inputs.item }} \
            ${PUBLISH_PUBLIC}
        fi
      shell: bash
~~~

This also gets rids of the need for a `publish` input, as the dashboards inputs do that for us.

.github/workflows/dv-all.yml
LouisTheLuis: Yes, it's a secret key. GitHub scrubs it from the logs, fortunately.

util/dvsim/SimCfg.py
```diff
@@ -554,6 +554,40 @@ def cov_unr(self):
...
+        sections = [("P", "PASSED"), ("F", "FAILED"), ("K", "KILLED")]
+        manifest_path = latest_dir / "seeds.txt"
+        try:
+            with open(manifest_path, "w") as f:
+                f.write(f"# timestamp: {self.timestamp}\n")
+                f.write("# build_seed: {}\n".format(self.build_seed or "none"))
+                f.write("# status full_name.seed\n\n")
+                for code, label in sections:
+                    f.write(f"# {label}\n")
+                    for r in buckets[code]:
```
matutem, line 585: Could you sort these lists for ease of perusal.

LouisTheLuis: Sort by test name or by test iteration? They are already sorted by bucket.

qmn (approved):

PR #136, opened 2026-04-30
[ci] Move gitless check to weekly

vlpannel:
Saving on resources during normal CI checks because the gitless check is only relevant when making repo releases (from git archive).


qmn (approved):
LGTM

Issue #43, opened 2026-03-25
[top_dragonfly,dv] Chip level dvsim tests failing due to using egret specific software


apinise:
Many of the `chip_sw_*` tests in `hw/top_dragonfly/dv/chip_sim_cfg.hjson` fail to build due to the `sw_images` dependencies of the test only having `egret` execution environments in `sw/device/tests/sim_dv/BUILD`. Additionally, some of these Bazel targets rely specifically on `top_egret.h` or other `egret` specific files and do not use device tables (DT) in the software. This issue temporarily comments out the following tests from `hw/top_dragonfly/dv/chip_sim_cfg.hjson` for the aforementioned reasons, until they have been ported to `dragonfly`, or deemed not necessary. Tests can be marked off here as they are fixed or removed.

#### Affected dvsim tests
- [ ] chip_sw_example_manufacturer
- [ ] chip_sw_data_integrity_escalation
- [ ] chip_sw_sleep_pin_wake
- [ ] chip_sw_sleep_pin_retention
- [ ] chip_sw_uart_tx_rx
- [ ] chip_sw_uart_tx_rx_bootstrap
- [ ] chip_sw_inject_scramble_seed
- [ ] chip_sw_exit_test_unlocked_bootstrap
- [ ] chip_sw_uart_rand_baudrate
- [ ] chip_sw_uart_tx_rx_alt_clk_freq
- [ ] chip_sw_i2c_host_tx_rx
- [ ] chip_sw_i2c_device_tx_rx
- [ ] chip_sw_spi_device_tpm
- [ ] chip_sw_spi_host_tx_rx
- [ ] chip_sw_otp_ctrl_lc_signals_test_unlocked0
- [ ] chip_sw_otp_ctrl_lc_signals_dev
- [ ] chip_sw_otp_ctrl_lc_signals_prod
- [ ] chip_sw_otp_ctrl_lc_signals_rma
- [ ] chip_sw_otp_ctrl_vendor_test_csr_access
- [x] chip_sw_lc_ctrl_transition
- [x] chip_sw_lc_walkthrough_dev
- [x] chip_sw_lc_walkthrough_prod
- [x] chip_sw_lc_walkthrough_prodend
- [x] chip_sw_lc_walkthrough_rma
- [x] chip_sw_lc_walkthrough_testunlocks
- [ ] chip_sw_pwrmgr_main_power_glitch_reset
- [ ] chip_sw_pwrmgr_sleep_power_glitch_reset
[...]

PR #137, opened 2026-05-01
[top_level,tests] Fix data_integrity_escalation_reset test

matutem:
This has two commits:
- Make the test muitl-top ready.
  Have the SV side communicate the expected alert a via backdoor
  symbol overwrite.
- Update memories from which backdoor accesses are allowed for darjeeling.
  Only flash was allowed for reads.

Addresses #43


matutem:
Won't commit this in this repo at this point.

PR #139, opened 2026-05-06
[sw,hello_world] Split off gpio_demo from hello_world

qmn:
Make `hello_world` much simpler by having it just initialize the UART and just print "Hello World!". Rename the original `hello_world` to `gpio_demo`.

This is a recreation of #66 due to some repo shuffling.


rtorok-zr (approved):
LGTM

PR #140, opened 2026-05-06
[doc,mdbook] Fix lack of images in tooling docs

vlpannel:
Currently, the mdbook documentation fails to render images within the tooling docs because the `util/` directory is being ignored by mdbook. This PR fixes this.


qmn (approved):
LGTM, thanks!

PR #143, opened 2026-05-08
[util] Let grade_regression_vcs fail when no units have seeds.txt

LouisTheLuis:
This will make it possible to also stop the script from generating an empty grading report when no `seeds.txt` exist.

### DIFF
--- a/util/grade_regression_vcs.py
+++ b/util/grade_regression_vcs.py
@@ -455,6 +455,13 @@ def main() -> int:
     # -------- Phase 2: Aggregate --------
     rows, per_unit_summary, missing, hjson_entries = aggregate_units(args)
 
+    if not hjson_entries:
+        log.error("No units produced grade reports; aborting without writing output.")
+        if missing:
+            for u, why in missing:
+                log.error(f"  {u}: {why}")
+        sys.exit(1)
+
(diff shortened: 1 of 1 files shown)


util/grade_regression_vcs.py
```diff
@@ -455,6 +455,13 @@ def main() -> int:
     # -------- Phase 2: Aggregate --------
     rows, per_unit_summary, missing, hjson_entries = aggregate_units(args)
 
+    if not hjson_entries:
+        log.error("No units produced grade reports; aborting without writing output.")
+    if missing:
+        for u, why in missing:
+            log.error(f"  {u}: {why}")
+    sys.exit(1)
```
qmn: I think the `sys.exit(1)` is misplaced; should it be if there aren't any hjson_entries, or if there are missing items? It seems it should be only if there are no hjson_entries that you exit with an error.

qmn (commented):
Mostly good, just one note:

util/grade_regression_vcs.py
LouisTheLuis: Oop, that was a typo. Fixed.

qmn (approved):
LGTM

PR #144, opened 2026-05-11
[top,md] Use top generically and fix regression

matutem:
The bazel mechanism this implements is not specific to a ROT. Put back a regressed relative link.


vlpannel (approved):
LGTM.

PR #145, opened 2026-05-11
[doc] Update block diagrams

qmn:
Update the block diagrams by redrawing them on www.drawio.com. The accompanying *.drawio files serve as the source of truth for these block diagrams.

When exporting these diagrams:
* Set Zoom to 100%
* Set Border Width to 0
* Enable Transparent Background
* Disable Shadow
* Don't include a copy of the diagram
* Don't embed images
* Don't embed fonts


matutem:
The instructions for generating diagrams need to go in some contributing markdown. I assume they can be a recommendation, but insist diagrams be made with tools that enable modifying diagrams, and which can save the actual diagram source.

LouisTheLuis (approved):
LGTM with a nit: wouldn't it be ideal to have the clock domains shown explicitly as before (labeled as 1GHz, 250MHz, etc.)

qmn:
I made an artistic decision to leave clock domains out and focus on the relationship between blocks. Clocks change/have changed over time and are also different platform to platform (ASIC vs FPGA). The actual timing needs will depend on integrator-specific requirements, although it's still worth noting that we can close at ~GHz where that's possible.

PR #148, opened 2026-05-11
[dvsim] More changes to dvsim

LouisTheLuis:
This PR contains two commits. The first one removes the copying of KDBs as we no longer need them for analysis.
The second one creates a branch on the dashboard repository identical to the branch from which `dvsim` is being ran.

The second commit allows for the following (in my opinion, simpler) procedure for testing particular blocks. Say we want to make another IP block on a branch named `ip_block/dev` off of `main` and we want to test the coverage of that block separately from the rest with a dedicated dashboard for it. Instead of pushing to `main`, `--publish` now will push to `ip_block/dev` in the respective dashboard repository. 

The only caveat is that the branch must already exist on the dashboard repository.

### DIFF
--- a/hw/dv/tools/dvsim/vcs.hjson
+++ b/hw/dv/tools/dvsim/vcs.hjson
@@ -210,9 +210,6 @@
   cov_report_txt:   "{cov_report_dir}/dashboard.txt"
   cov_report_page:  "dashboard.html"
 
-  // Knowledge database
-  knowledge_db_dir_name: "kdb.elab++"
-
   // UNR related.
   // All code coverage, assert isn't supported
   cov_unr_metrics: "line+cond+fsm+tgl+branch"
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -565,8 +565,38 @@ class FlowCfg():
         subprocess.run(["git", "-C", repo_dir, "commit", "-m",
                         f"[dashboard] Update {self.name}_{test}_{self.flow} {self.timestamp}"],
                        check=True, env=env)
-        subprocess.run(["git", "-C", repo_dir, "push"], check=True, env=env)
+        subprocess.run(["git", "-C", repo_dir, "push", "origin", self.branch],
+                       check=True, env=env)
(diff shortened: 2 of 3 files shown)


matutem:
I assume the mode of operation enabled by the second commit is optional, right? How is it triggered? 

I hope it is not the default mode, meaning it should not always goes into an existing dashboard.

LouisTheLuis:
@matutem It's a mode only for publishing. The branch tracking is always there, however. That means that if you have, say, a personal dashboard `git@github.com:matutem/foo`, whenever you use `--publish` or `--publish-prev` it will push to the corresponding branch you are working from.

matutem:
I see, so how do we configure it all for a nightly or weekly regression to avoid overwriting stuff? Where is this documented?

LouisTheLuis:
@matutem We don't have to do anything; the `weekly.yml` and `nightly.yml` already work off the `main` branch, so they push to the `main` branches of both `internal-dashboard` and `public-dashboard`.

I ran an experiment with running a reduced `weekly.yml` from a separate branch (`pavona-dvsim-branch`), and it pushed to a `pavona-dvsim-branch` on both dashboards. That's how we keep them separate.

Therefore, the flow would be to:
- Allow CI to run jobs off of a branch.
- Write the weekly / nightly to run the jobs that we care about in that branch

As an example, if we were working with new blocks `ip_new_0`, `ip_new_1` and `ip_new_2` in the `new_ips` branch, we could make a `ip_new_sim_cfgs.hjson` that only contains those IPs. Then, modify `weekly.yml` and `nightly.yml` to only run `dv-all.yml` with that `ip_new_sim_cfgs.hjson`. With this, we will have a dashboard with a `main` branch that tracks the coverage of Pavona and a `new_ips` branch that tracks the coverage of the IPs in development.

There's no overwriting at play, it's all `git add` and `git push`, after all.

matutem:
Thanks for the explanation: this needs to be documented. Could you open an issue?

matutem (approved):
Please open an issue to document the whole publish feature, perhaps including how it interacts with branches.

LGTM

PR #152, opened 2026-05-14
[doc] Add link to pavona.org in header, rename introduction chapters in doc table of contents

vlpannel:
This PR makes a few minor changes to the documentation organization and styling.

* Have multiple chapters labeled in the table of contents as "Introduction" gives an inadequate level of detail, so they were renamed.
* Also add the Pavona logo (as SVG) into the mdbook header and linked it to the official Pavona website URL.
* Replace directly rendering templates with default values with rendering template description and IP configuration Hjsons to give more context about the parameters an IP template uses.

To view changes, checkout this branch and run `util/site/build-docs.sh serve` from the repo top, and it will be locally hosted at port 9000.

(Commits will be squashed into one.)


qmn (approved):
This looks good to me, just a fix and then a squash and then I think it LGTM:

Seems to be a stray  ./hw/top_earlgrey/ip_autogen/rstmgr/data/top_earlgrey_rstmgr.ipconfig.hjson not being picked up in the Build and test software CI jobs.

PR #161, opened 2026-05-19
[cryptolib] Sync cryptolib hardware from zerorisc/expo

qmn:
Due to the GitHub limitation on the "Rebase and merge" option of 100 commits, this PR merges the _second_ batch of 22 commits (with #162 having merged the _first_ 88).

PR #162, opened 2026-05-19
[crypto] Sync cryptolib software from zerorisc/expo

qmn:
Due to a GitHub limitation of 100 commits on the "Rebase and merge" option, this is a subset of the first 88 commits of #161. This needs to get merged before #161.


mkannwischer (approved):
Thanks @qmn. LGTM. 
I haven't reviewed every commit, but the set of commits and the touched files look reasonable to me.

PR #169, opened 2026-05-20
[dvsim] Fix dashboard design to look simpler

LouisTheLuis:
This commit removes a redundant column and makes the necessary CSS adjustments for that.


qmn (approved):
LGTM, thanks!

PR #172, opened 2026-05-21
 [dvsim] Improve index helper function

LouisTheLuis:
This commit regenerates the `index.html` based on the current directories present. Prunes old directories based on `max_entries`. Presents the dashboards in newest to oldest.


qmn (approved):
Looks good to me!

PR #168, opened 2026-05-20
[ci,gcp] Re-enable bazel and bitstream caches

rtorok-zr:
This PR configures the repository to use the Google Cloud buckets provided by the Pavona Project for caching bazel targets and FPGA bitstreams. 

### Bitstream Cache

Bitstreams built by CI are built and uploaded to the cache when a Pull Request is merged to the `main` branch. In a local checkout, you can enable the `post-checkout` hook to automatically fetch the latest matching bitstream each time you switch branches:

```cp util/git/hooks/post-checkout` .git/hooks/```

### Bazel Remote Cache

The Bazel remote cache stores pre-built artifacts from the `main` branch of the repository, making rebuilds faster. Downloading cached builds in a local checkout is configured automatically in `.bazelrc`.

PR #176, opened 2026-05-21
[ci,gcp] Fix project IDs for access to caches

rtorok-zr:
Fixes an issue with #168, where the project identifiers for authentication to Google Cloud via Workload Identity Federation were not updated correctly, causing a crash when attempting to authenticate to GCP from the `pavona` repository context.


qmn (approved):
LGTM

PR #177, opened 2026-05-21
[verilator] Mitigate risk of FuseSoC running local Verilator

rtorok-zr:
[Recreation of #166]

Refactors the implementation of `fusesoc_build` to eliminate `/usr/local/bin` from PATH when running FuseSoC with Verilator.

### TODO

`/usr/bin` still remains in the `PATH` when running FuseSoC with `Verilator` due to a dependency on the host `make`. The user is responsible for ensuring that `/usr/bin/verilator` does not contain an alternate undesired Verilator installation, e.g. if Verilator `4.x.X` is installed with `apt`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,7 +235,7 @@ jobs:
     needs: quick_lint
     runs-on:
       group: pavona-basic
-    timeout-minutes: 20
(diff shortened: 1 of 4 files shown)


rtorok-zr:
Obsoleted by #339. Closing.

PR #178, opened 2026-05-21
[README] Update readme, add logo lockups

qmn:
Update the README with some more information and add the logo lockup. ~The logo lockup doesn't seem to appear in the preview, but you can get a preview by visiting my version at https://github.com/qmn/pavona/tree/update-readme.~ (The logo lockup does appear in the rich diff after all.)


vlpannel (approved):
LGTM.

PR #179, opened 2026-05-22
[dj,fpga] Add slot for second ROM in FPGA bitstreams

rtorok-zr:
This PR updates the build tooling for Darjeeling FPGA bitstreams to make Vivado provision separate splice-able regions for `rom_ctrl0` and `rom_ctrl1`.

To retain the property that the bitstream with nothing spliced runs through the test ROM and then jumps to CTN RAM, this PR creates a `test_second_rom` as the default payload for `rom_ctrl1`. This binary performs no dedicated chip setup and is simply a passthrough to CTN.

On Xilinx FPGAs, the `rom1` BRAM region is built using a very strange bit order. For the lack of a better term, it is "crumb endian", i.e. in addition to swapping the nibbles within a byte as is done with `rom0`, we also have to swap the order of the bit-pairs within each nibble. In other words, this is equivalent to writing the bits in reverse order and then swapping the even and odd bits.

### DIFF
--- a/hw/top_darjeeling/util/vivado_hook_write_bitstream_pre.tcl
+++ b/hw/top_darjeeling/util/vivado_hook_write_bitstream_pre.tcl
@@ -1,1 +1,258 @@
-../../../hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
+# Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+send_msg "Designcheck 1-1" INFO "Checking design"
+
+# Ensure the design meets timing
+set slack_ns [get_property SLACK [get_timing_paths -delay_type min_max]]
+send_msg "Designcheck 1-2" INFO "Slack is ${slack_ns} ns."
+
+if [expr {$slack_ns < 0}] {
+  send_msg "Designcheck 1-3" ERROR "Timing failed. Slack is ${slack_ns} ns."
+}
+
+# Enable bitstream identification via USR_ACCESS register.
+set_property BITSTREAM.CONFIG.USR_ACCESS TIMESTAMP [current_design]
+
+# Generate an MMI file for the given BRAM cells.
+#
+# Args:
+#   filename:            Path to the output file.
+#   mem_info:            Dictionary of dictionaries with following properties for each (inner) value:
+#       brams:               A list of BRAM cells.
+#       mem_type_regex:      The BRAM type regex, dividing the mem type and site, e.g. {(RAMB\d+)_(\w+)}.
+#       fake_word_width:     If non-zero, pretend that $brams covers
+#                            `fake_word_width` bits. Influences the values of the
+#                            MMI's <AddressSpace> and <DataWidth> tags.
+#       addr_end_multiplier: A coefficient applied to the address space. Influences
+#                            the values of the MMI's <AddressSpace> and
+#                            <AddressRange> tags.
+#       schema:              Either "Processor" or "MemoryArray"
+#   designtask_count:    A number used for logging with `send_msg`.
+proc generate_mmi {filename mem_infos designtask_count} {
+    send_msg "${designtask_count}-1" INFO "Dumping MMI to ${filename}"
+
+    set workroot [file dirname [info script]]
+    set filepath "${workroot}/${filename}"
+    set fileout [open $filepath "w"]
+    set part [get_property PART [current_design]]
+    puts $fileout "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+    puts $fileout "<MemInfo Version=\"1\" Minor=\"1\">"
+
+    dict for {id mem_info} $mem_infos {
+        dict with mem_info {
+            if {[llen $brams] == 0} {
+                send_msg "${designtask_count}-1" INFO "Cannot make MMI for zero BRAMs"
+                continue
+            }
+
+            set fake_slice_width [expr $fake_word_width / [llen $brams]]
+
+            # Calculate the overall address space.
+            set space 0
+            set width 0
+            foreach inst [lsort -dictionary $brams] {
+                set slice_begin [get_property ram_slice_begin [get_cells $inst]]
+                set slice_end [get_property ram_slice_end [get_cells $inst]]
+                if {$slice_begin eq {} || $slice_end eq {}} {
+                    send_msg "${designtask_count}-2" ERROR "Extraction of ${filename} information failed."
+                }
+                set slice_width [expr {$slice_end - $slice_begin + 1}]
+                if {$slice_width < $fake_slice_width} {
+                    set slice_end [expr {$slice_begin + $fake_slice_width - 1}]
+                    set slice_width $fake_slice_width
+                }
+                set addr_begin [get_property ram_addr_begin [get_cells $inst]]
+                set addr_end [get_property ram_addr_end [get_cells $inst]]
+                if {$addr_begin eq {} || $addr_end eq {}} {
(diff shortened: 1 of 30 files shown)


hw/top_darjeeling/util/vivado_hook_write_bitstream_pre.tcl
rtorok-zr, line 1: This file is copied from `hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl`, except that `rom0` and `rom1` are divided (l. 228-232)

sw/device/lib/testing/test_second_rom/test_second_rom.ld
```diff
@@ -0,0 +1,169 @@
+/* Copyright lowRISC contributors (OpenTitan project). */
```
rtorok-zr, line 1: Nearly a duplicate of `sw/device/lib/testing/test_rom/test_rom.ld`.

sw/device/lib/testing/test_second_rom/test_second_rom_start.S
```diff
@@ -0,0 +1,76 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Test second ROM runtime initialization code.
+ */
+
+  // NOTE: The "ax" flag below is necessary to ensure that this section
+  // is allocated executable space in ROM by the linker.
+  .section .crt, "ax"
```
pqcfox, line 12: should the options from `test_rom_start.S` be mirrored here (e.g. `.option push` or `.option norvc`), or will those automatically be mirrored?

sw/device/lib/testing/test_rom/test_rom.c
```diff
@@ -72,7 +72,12 @@ static const dt_otp_ctrl_t kOtpCtrlDt = kDtOtpCtrl;
  */
 extern char _rom_ext_virtual_start_address[];
 extern char _rom_ext_virtual_size[];
+#ifndef HAS_ROM_CTRL1
 extern char _manifest_address[];
+# else
```
pqcfox: [nit] `#else`

hw/bitstream/vivado/BUILD
```diff
@@ -31,6 +31,10 @@ _CW340_TESTROM_PATH = _CW310_TESTROM_PATH
 
 _OTP_RMA_PATH = "{}/$(location {})".format(_PREFIX, _OTP_RMA)
 
+_EARLGREY_OTP_RMA_PATH = "{}/$(location {})".format(_PREFIX, _EARLGREY_OTP_RMA)
```
pqcfox: clarifying q: are these presently used?

sw/device/lib/testing/test_second_rom/test_second_rom.c
```diff
@@ -0,0 +1,124 @@
...
+static const dt_uart_t kUart0Dt = kDtUart0;
+static dif_uart_t uart0;
+
+/* These symbols are defined in
+ * `opentitan/sw/device/lib/testing/test_rom/test_rom.ld`, and describes the
+ * location of the flash header and manifest.
+ */
+extern char _rom_ext_virtual_start_address[];
+extern char _rom_ext_virtual_size[];
+extern char _manifest_address[];
```
pqcfox, line 37: looks like (correct me if I missed something) this isn't defined in test_rom.ld if OPENTITAN_IS_DARJEELING is defined; would this be an issue?

sw/device/lib/testing/test_second_rom/BUILD
```diff
@@ -0,0 +1,162 @@
...
+[
+    filegroup(
+        name = "test_second_rom_{}_hashfile".format(dev),
+        srcs = [":test_second_rom"],
+        output_group = "{}_hashfile".format(dev),
+    )
+    for dev in [
+        "sim_dv",
+        "sim_verilator",
+        "fpga_cw305",
```
pqcfox, line 68: [nit] could maybe exclude FPGA targets which the DJ bitstream won't fit on?

sw/device/lib/testing/test_second_rom/BUILD
```diff
@@ -0,0 +1,162 @@
...
+    ]
+]
+
+cc_library(
+    name = "darjeeling_test_second_rom_lib",
+    srcs = [
+        "darjeeling_fake_driver_funcs.c",
+    ],
+    deps = [
+        "//sw/device/lib/dif:rstmgr",
```
pqcfox: [nit] can maybe remove these deps?

sw/device/lib/testing/test_second_rom/BUILD
```diff
@@ -0,0 +1,162 @@
...
+
+pkg_files(
+    name = "package",
+    testonly = True,
+    srcs = [":pre_package"],
+    prefix = opentitan_select_top(
+        {
+            "darjeeling": "darjeeling/test_second_rom",
+        },
+        "UNKNOWN/test_second_rom",
```
pqcfox: clarifying q: what does the UNKNOWN prefix do here?

hw/bitstream/vivado/BUILD
rtorok-zr: This was leftover code, removing.

sw/device/lib/testing/test_second_rom/test_second_rom.c
rtorok-zr: The `test_second_rom` has these symbols declared in `test_second_rom.ld`. The `#ifdef` in `test_rom.ld` exists because in tops with `rom_ctrl1` we won't jump directly from the `test_rom` to ROM_EXT.

sw/device/lib/testing/test_second_rom/BUILD
rtorok-zr: I think it's better just to mirror `test_rom/BUILD` for now. This will need a larger refactor once more FPGA platforms are supported, possibly with smaller tops that fit on more boards.

sw/device/lib/testing/test_second_rom/BUILD
rtorok-zr: Nothing. I just wanted this to not be a top-specific target, and I needed a default name. I changed it to just `test_second_rom` instead.

sw/device/lib/testing/test_second_rom/test_second_rom_start.S
rtorok-zr: Those options are only applied in the `test_rom` to the `interrupt_vector` that is responsible for jumping to `_reset_start` on a reset, so this does not apply to `test_second_rom`.

sw/device/lib/testing/test_second_rom/test_second_rom.c
```diff
@@ -0,0 +1,124 @@
...
+#include "sw/device/silicon_creator/lib/drivers/retention_sram.h"
+#include "sw/device/silicon_creator/lib/manifest.h"
+
+static const dt_rv_core_ibex_t kRvCoreIbexDt = kDtRvCoreIbex;
+
+static const dt_uart_t kUart0Dt = kDtUart0;
+static dif_uart_t uart0;
+
+/* These symbols are defined in
+ * `opentitan/sw/device/lib/testing/test_rom/test_rom.ld`, and describes the
```
pqcfox: [nit] maybe should say `test_second_rom.ld`?

sw/device/lib/testing/test_second_rom/test_second_rom.c
pqcfox: Ah, perfect, thanks!

sw/device/lib/testing/test_second_rom/test_second_rom_start.S
pqcfox: Makes sense; resolved, thanks!

pqcfox (approved):

PR #187, opened 2026-05-23
[doc] Final updates

vlpannel:
Last passthrough through technical documentation before launch. Commits to be squashed into one.


qmn (approved):
LGTM, thanks for pushing through all these changes. Because the files changed are almost all docs and metadata (the one *.h file has changes only in comments), I'd just wait for the fast and slow lints to finish before merging. Please also squash your commits and remember to remove the helper script, too. Thanks!

PR #190, opened 2026-05-23
[dvsim] Merge lesser reseed when conflicting (when using --reseed-source)

qmn:
The list of reseeds passed to dvsim via `--reseed-source` can conflict with the seed provided by a mode (e.g. the mem_tests_mode of run_modes in mem_tests.hjson). In this case, select the _lesser_ of the two reseed values, as we're typically here because we're doing a graded run.


qmn:
Filed #191.

PR #192, opened 2026-05-24
[repo] Rename top levels

vlpannel:
The newly chosen names for the tops are `top_dragonfly` (integrated RoT top level) and `top_egret` (discrete chip RoT top level). This PR changes those names. Some technical things that were required to achieve a functional rename:
* adjusting third party BUILD files for tock and QEMU to accurately reflect their source repos' structures
* patching QEMU to correctly map the new top names to their former names

The experimental top level is likely going to be deprecated, so it has only a placeholder name for the sake of making renames easier (scripts were originally written with the intent of renaming all three); its rename commits are separated so that the placeholder name `top_scafi_deprecated` can easily be dropped.

(Renaming scripts will be removed from PR, commits will be squashed.)


rtorok-zr (approved):
LGTM for now after squashing the commits. Great to see this migration building on CI.

A few thoughts:
- I would remove the `find_names.sh` script. It was an impressive trick to automate renaming the tops, but I doubt it will be needed again after this PR.
  - Edit: I see you commented this above.
- We will likely want to flush the bitstream cache after this is merged and the `Post-Merge Artifacts` job runs, so we don't have stale cache entries with the old top names.
- ACK about Tock being pinned to the old top name. Let's leave it how you have it for now, but I will definitely write down that I should talk to Amit about supporting heterogeneous Pavona tops in a more modular way. Preferably, this looks like avoiding creating a separate driver crate for each top.

qmn (approved):
Looks good to me as well, thanks very much for this work! And, yes, as Ryan says, please remove the helper scripts, squash, and please proceed with merging.

PR #200, opened 2026-05-25
[dragonfly,chip] Fix chiplevel template, update system list

qmn:
A prior commit inadvertently brought back the pinmux configuration, so get rid of that (and an otherwise unused ast_init_done signal). This seems to get the dragonfly chip-level DV tests back up and compiling.

Also, while regenerating with `make -C hw top_and_cmdgen`, gen_system_list caught some old directories which caused an error in util/gen_system_list.py. It also seems that the detection of the verilator target is better done by simply checking for the existence of `chip_[top]_verilator.core`, so do that instead of searching for "sim_verilator" (searching for "verilator" alone is fraught because of the various Verilator-based linting targets).

### DIFF
--- a/doc/contributing/system_list.md
+++ b/doc/contributing/system_list.md
@@ -6,8 +6,8 @@ Click on the design name to get more information about the design.
 <!-- BEGIN CMDGEN util/gen_system_list.py -->
 | Design | Targets | Description |
 | --- | --- | --- |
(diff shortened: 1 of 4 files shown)


vlpannel (approved):
LGTM.

qmn:
Thanks for taking a look! The failing tests are known spotty and, in any case, affect Egret.

PR #208, opened 2026-05-26
[acc,doc] Updated documentation for ACC hardware configurations

apinise:
Added a table to explicitly define the memory sizes and key features of the classical and post-quantum hardware configurations. Made statements about IMEM/DMEM generic to the hardware configurations instead of stating the sizes directly.

### DIFF
--- a/hw/ip/acc/doc/theory_of_operation.md
+++ b/hw/ip/acc/doc/theory_of_operation.md
@@ -6,11 +6,36 @@
 
 ## Design Details
 
+### Configurations
+
+The ACC processor core supports two configurations: a classical mode supporting algorithms such as ECDSA, ECDH, RSA, alongside a post-quantum enabled mode that supports the classical operations in addition to ML-KEM and ML-DSA across all parameter sets.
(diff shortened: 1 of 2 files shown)


hw/ip/acc/doc/theory_of_operation.md
```diff
@@ -6,11 +6,36 @@
 
 ## Design Details
 
+### Configurations
+
+The ACC processor core supports two configurations: a classical mode supporting algorithms such as ECDSA, ECDH, RSA, alongside a post-quantum enabled mode that supports the classical operations in addition to ML-KEM and ML-DSA across all parameter sets.
+ACC requires different hardware configurations to meet these specifications, and is controlled at synthesis time through the `AccPQCEn` RTL parameter.
+The instruction and data memory sizes are configured separately from the rest of the RTL, but both are configured in the acc.hjson(../data/acc.hjson) file.
```
mkannwischer:
```suggestion
The instruction and data memory sizes are configured separately from the rest of the RTL, but both are configured in the [acc.hjson](../data/acc.hjson) file.
```

mkannwischer (approved):
Thanks @apinise. That is clearer to me.

PR #210, opened 2026-05-26
[util] Do not generate table entry for additional tops

LouisTheLuis:
This commit modifies `gen_system_list.py` such that new experimental tops or tops in development don't break `make -C hw all` due to a missing `.core` or `top_cfg`. These just get dropped from the table generator.

### DIFF
--- a/util/gen_system_list.py
+++ b/util/gen_system_list.py
@@ -61,6 +61,8 @@ def get_targets(topdir):
         top_data = hjson.loads(top_cfg.read_text())
         target_names = [t["name"] for t in top_data["targets"]]
         hw_targets = ", ".join(target_names)
+    else:
+        raise Exception(f"Could not find top_cfg within {topdir}.")
 
     return sim_targets + (", " * int(bool(sim_targets and hw_targets))) + hw_targets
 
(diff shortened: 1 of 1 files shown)


vlpannel (approved):
LGTM, though depending on whether this PR or #200 is merged first, there may be some minor conflicts to resolve.

util/gen_system_list.py
```diff
@@ -55,12 +55,16 @@ def get_targets(topdir):
     sim_targets = ""
     if verilator_file.is_file():
         sim_targets = "verilator"  # only one known sim target
+    else:
+        raise Exception(f"Could not find corefile within {topdir}.")
```
vlpannel: You might not want to raise an exception here: a top level may have other core files that are not for verilator targets but are for other hardware targets (like FPGAs, ASICs). This does not mean that the top is necessarily incomplete.

It may be enough to check just for the top level configuration to see if the top level is complete or not, or you may want to check to see if there are any core files within the toplevel directory.

LouisTheLuis: Noted. I will remove the verilator check and just leave the `top_cfg` one.

vlpannel (approved):
The adjustment makes sense; LGTM.

PR #213, opened 2026-05-27
[doc] Add sitemap.xml to doxygen output

fragglet:
This extends the current Doxyfile to generate a sitemap.xml file that references all generated .html files. Search engines will then be able to automatically discover all files on the docs site.


vlpannel (approved):
LGTM, thanks!

PR #214, opened 2026-05-28
[acc,mlkem] Make KYBER_K a runtime argument

mkannwischer:
This PR changes KYBER_K from a compiler-time constant to a run-time argument in all ML-KEM sources. This will allow to build a single binary that has support for all ML-KEM parameter sets without any unnecessary code duplication.

I left the incremental changes as separate commit to ease review (each commit passes tests), but we can likely squash into a single commit before merging.


sw/acc/crypto/mlkem/poly.s
```diff
@@ -188,30 +177,50 @@ poly_getnoise_eta_init:
 
 .globl poly_getnoise_eta_1
 poly_getnoise_eta_1:
+  /* eta1 = 3 for ML-KEM-512 (K=2), eta1 = eta2 = 2 for ML-KEM-768/1024
+   * (K=3,4). Dispatch on x14 to the matching CBD sampler. */
+  li  x28, 2
```
pqcfox, line 182: Possible consideration for later: would we potentially want to branch on hardened values for `KYBER_K` as defense-in-depth for branches such as the one directly below this? (not sure how significant of an impact this would have, especially since the actual `KYBER_K` value necessarily is used in offset arithmetic here, but thought I'd mention)

sw/acc/crypto/mlkem/poly.s
```diff
@@ -188,30 +177,50 @@ poly_getnoise_eta_init:
...
+  beq x14, x28, poly_getnoise_cbd3
+  beq x0, x0, poly_getnoise_cbd2
```
pqcfox: [super nit] tail-call with `jal x0, poly_getnoise_cbd2` might match a more common idiom w/ other ACC code? (may also indicate end-of-function more clearly to static analysis tooling--iirc it explicitly detects this tail-call idiom, but I'd need to check on that)

sw/acc/crypto/mlkem/pack_ciphertext.s
```diff
@@ -29,15 +24,13 @@
  * @param[in]  x11: dptr_input, dmem pointer to input polynomial
  * @param[out] x12: dptr_output, dmem pointer to output byte array
  * @param[in]  x13 (w3): const_80635
- * @param[in]  x14 (w6): modulus_bn
```
pqcfox, line 32: (clarification q, was this being used prior?)

sw/acc/crypto/mlkem/mlkem_decap.s
```diff
@@ -272,37 +194,27 @@ crypto_kem_dec:
...
   /* Set frame pointer */
   addi fp, sp, 0
-#if KYBER_K == 2
-    li  t0, -3296
-#elif KYBER_K == 3
-    li  t0, -4320
-#elif KYBER_K == 4
-    li  t0, -5344
-#endif
+  li   t0, -5344
```
pqcfox, line 203: Possible consideration for later: would it be worth evaluating the stack overhead on ML-KEM-512 from taking worst-case offsets? wondering if for very small DMEM configruations if this would an impact (not that this needs to be addressed in this PR, just if it's something to note for later).

sw/acc/crypto/mlkem/mlkem_indcpa_enc.s
```diff
@@ -338,13 +264,16 @@ indcpa_enc:
       la   x28, twiddles_basemul
       jal  x1, basemul_acc
       addi a1, a1, POLY /* points back to A[0][1] */
-    .endr
+2:
```
pqcfox: [super nit] labels here/below could be named to suggest that the loop skip is to prevent `LOOP` failing on a zero input

pqcfox (approved):
Looks fantastic, kudos on getting this done!

sw/acc/crypto/mlkem/poly.s
mkannwischer: It would be good to write this on a list to consider later, yes. One practical issue with that is that it's often used to control how often a loop runs. But in places where we branch on it, we should use hardened values.

sw/acc/crypto/mlkem/mlkem_decap.s
mkannwischer: Do I understand correctly that you wonder if we could fit ML-KEM-512 only into less DMEM? We should evaluate that. Probably 768 is more important - as far as I am aware no one is really using 512. 

I can imagine that we will at some point have a configuration where only one parameter set it supported. This can also be very useful to support smaller code size.

sw/acc/crypto/mlkem/pack_ciphertext.s
mkannwischer: That was just stale doc that I cleaned up on the way.

PR #217, opened 2026-06-01
[acc] Minor improvements to information-flow checker

mkannwischer:
This PR cherry-picks @jadephilipoom's PR from https://github.com/zerorisc/expo/pull/246. There are going to be two follow-up PRs.

Original PR description:

These were some small improvements I made along the way towards a bigger extension to the information-flow checker that I'll post as a follow-up PR; I separated these out for easier review. There's more detail in the commit messages, but in summary:

- print out control-flow nodes more nicely for debugging
- automatically resolve branches where both sides of the comparison are known constants
- allow specifying public values instead of secret values for constant-time checks, which makes it easier to ensure nothing has been forgotten


pqcfox (approved):
Looks perfect, thank you so much!

PR #223, opened 2026-06-02
[acc] Auto-discover in-tree LLVM toolchain in find_tool

mkannwischer:
It was reported by @phamhnh that when trying to run `acc_as.py` directly, one gets this unhelpful and wrong error message:
```
Unable to find 'riscv32-unknown-elf-gcc' in PATH or in '/tools/riscv/bin'. Set the 'RV32_TOOL_GCC' or TOOLCHAIN_PATH environment variable if you installed your RISC-V toolchain in an alternate location. (Hint: if you installed the toolchain via Bazel, this is likely 'bazel-pavona/external/lowrisc_rv32imcb_files/' )
```

The toolchain now actually lives under `bazel-pavona/external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/`, but I believe we can just auto-discover it.

This PR fixes the erorr message, and also makes `acc_as.py` auto-dicover the toolchain installed via bazel.

### DIFF
--- a/hw/ip/acc/util/shared/toolchain.py
+++ b/hw/ip/acc/util/shared/toolchain.py
@@ -1,8 +1,42 @@
 # Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+import glob
 import os
+from typing import Optional
+
+
+# GNU-style tool name -> the binary the in-tree LLVM toolchain provides.
+_LLVM_TOOL_NAMES = {
+    'gcc': 'clang',
+    'as': 'clang',
+    'ld': 'ld.lld',
+    'ar': 'llvm-ar',
+    'objcopy': 'llvm-objcopy',
+    'objdump': 'llvm-objdump',
+}
+
+
+def _find_bazel_llvm_tool(tool_name: str) -> Optional[str]:
+    '''Find tool_name in the Bazel LLVM toolchain under bazel-*/external/.'''
+    llvm_name = _LLVM_TOOL_NAMES.get(tool_name)
+    if llvm_name is None:
+        return None
+
+    repo_root = os.path.dirname(os.path.abspath(__file__))
+    while not os.path.exists(os.path.join(repo_root, 'MODULE.bazel')):
+        parent = os.path.dirname(repo_root)
+        if parent == repo_root:
+            return None
+        repo_root = parent
+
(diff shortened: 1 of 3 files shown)


hw/ip/acc/util/shared/toolchain.py
```diff
@@ -15,10 +49,11 @@ def find_tool(tool_name: str) -> str:
     2. Use the path set in $TOOLCHAIN_PATH/bin/riscv32-unknown-elf-<tool_name>.
     3. Look for riscv32-unknown-elf-<tool_name> in the system PATH.
     4. Look in the default toolchain install location, /tools/riscv/bin.
+    5. Fall back to the in-tree LLVM toolchain under bazel-*/external/.
```
mkannwischer, line 52: I'm unsure what order we want to try this in. I'd say the default for most users would be the toolchain installed via bazel, but I do see that one may want to overwrite it.

Thoughts @rtorok-zr?

mkannwischer: This order is probably not what we want because if someone has the old toolchain in their PATH or has set RV32_TOOL_XXX set to an old toolchain, we get very unexpected behavior.

rtorok-zr: The general practice is that this shouldn't be overridable. If the system PATH could override the binary, we start having machine-specific issues, which is what Bazel is trying to prevent. May take is that the in-tree LLVM toolchain should always be default.

util/acc_build.py
```diff
@@ -122,13 +124,12 @@ def call_acc_ld(src_files: List[Path], out_file: Path,
 
 
 def call_rv32_objcopy(args: List[str]):
-    rv32_tool_objcopy = os.environ.get('RV32_TOOL_OBJCOPY',
-                                       'riscv32-unknown-elf-objcopy')
+    rv32_tool_objcopy = os.environ.get('RV32_TOOL_OBJCOPY') or find_tool('objcopy')
```
rtorok-zr, line 127: In line with my comment above, I recommend just removing the `RV_TOOL_*` environment variables from this script. They existed prior to the LLVM migration because the build system was relying on an out-of-tree RISC-V toolchain.

hw/ip/acc/util/shared/toolchain.py
mkannwischer: I agree. I'm in favour of making that the only supported way. Then we can discard this PR and greatly simplify the toolchain.py.

Thinking about it more, I guess we also should discourage invoking acc_as.py directly. Maybe we can emit a warning nudging users to use the corresponding bazel target directly? And then eventually we turn that into an error. WDYT?

/cc @fragglet @adurbin-zr

PR #185, opened 2026-05-22
[doc] Update block diagrams

qmn:
Change their name, fix dark mode rendering quirks in the legend, and add fallback fonts.

PR #224, opened 2026-06-02
[doc,site] Use white background as default for SVGs

fish:
This adds a white background to SVGs by default when the dark theme is selected. This improves readability of many diagrams.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Pavona logo](./doc/pavona-logo-lockup.svg)
+![Pavona logo](./doc/pavona-logo-lockup.svg#no-bg)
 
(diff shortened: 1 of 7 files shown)


vlpannel (approved):
LGTM! Though I will note that for the `top_egret` and `top_darjeeling` block diagrams, having a white background will not change readability (fixed in #185).

fish:
> LGTM! Though I will note that for the `top_egret` and `top_darjeeling` block diagrams, having a white background will not change readability (fixed in #185).

Yep! Thanks, I added a `#no-bg` hash to those diagrams so that no background is placed in dark mode.

PR #225, opened 2026-06-03
[ci] Limit resources of weekly runs

LouisTheLuis:
Reduce concurrent jobs to 28.


rtorok-zr (approved):
Not sure how much effect lowering the job count from 32 to 28 will have, but let's go ahead and try.

PR #226, opened 2026-06-03
[ipgen] Include argument for removing top references

LouisTheLuis:
This commit includes a new flag `--no-top` that can be passed to ipgen to generate IP files without having references that are top-dependent (e.g. references in README.md files to other IPs that would exist in a top, like AST).


vlpannel (approved):
LGTM! Just remember to re-run topgen to fix the lint error on generated files.

PR #231, opened 2026-06-04
[ci,bitstream] Optimize resource usage for bitstream cache check

rtorok-zr:
The Vivado runner environment on CI has more limited throughput than the basic software environment, due to the long running time and high resource requirements of bitstream builds. This PR updates the CI script for building FPGA bitstreams to use a more lightweight runner environment for checking the bitstream cache, preventing workflows that would receive a cache hit from becoming blocked on other Vivado builds.


qmn (approved):
Very neat, thanks! LGTM (but let's give CI a go).

PR #232, opened 2026-06-04
[pentest] remove invalid ECDH keyblob zeroing

prusnak:
Remove the attempted zeroing of the second ECDH private-key share via memcpy(..., 0, ...). The second argument to memcpy is a source pointer, so passing 0 would read from the null address rather than write zeroes.

The full keyblob is already cleared with memset before copying the first share, so the second share remains zero-initialized without any additional operation.

### DIFF
--- a/sw/device/tests/penetrationtests/firmware/sca/cryptolib_sca_asym_impl.c
+++ b/sw/device/tests/penetrationtests/firmware/sca/cryptolib_sca_asym_impl.c
@@ -227,8 +227,6 @@ status_t cryptolib_sca_p256_ecdh_impl(
   uint32_t private_keyblob[kPentestP256MaskedPrivateKeyWords * 2];
   memset(private_keyblob, 0, sizeof(private_keyblob));
   memcpy(private_keyblob, uj_input.private_key, P256_CMD_BYTES);
-  memcpy(private_keyblob + kPentestP256MaskedPrivateKeyWords, 0,
-         P256_CMD_BYTES);
   otcrypto_blinded_key_t private_key = {
       .config =
           {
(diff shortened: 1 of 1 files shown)


mkannwischer (approved):
Thanks @prusnak for spotting and fixing this!

I agree with your assessment that this is wrong and the two lines are unnecessary anyway. The fix looks good to me!

We should really be catching such bugs automatically. I'll look into that.

Will you be able to sign the CLA? Let me know if we can help.

mkannwischer:
> Thanks @prusnak for spotting and fixing this!
> 
> I agree with your assessment that this is wrong and the two lines are unnecessary anyway. The fix looks good to me!
> 
> We should really be catching such bugs automatically. I'll look into that.
> 
> Will you be able to sign the CLA? Let me know if we can help.

I dug a bit deeper: The reason that this does not result in a NULL-pointer dereference is that the `memcpy` implementation checks for NULL pointers. This is thus a no-op:
https://github.com/pavona/pavona/blob/194493ec6536b5c902b831d038e924393288e80c/sw/device/lib/base/memory.c#L61-L66

I'm wondering if that NULL check there serves any purpose. According to the C11 standard (which is the semantics that our `memcpy` claims to implement), passing NULL pointers is undefined behaviour. I'm inclined to remove the NULL check and mark input pointers to `memcpy` and other memory operations as non-null. Then the compiler would warn you if you try to pass a NULL pointer. I'll open a PR for that shortly as it flagged another similar bug. 
This will need broader discussion, though, as I may be missing something here.

prusnak:
> Will you be able to sign the CLA? Let me know if we can help.

I already did yesterday.

mkannwischer:
> > Will you be able to sign the CLA? Let me know if we can help.
> 
> I already did yesterday.

Cool, thanks!
@rtorok-zr can you get that setup, so we can get this PR merged?

mkannwischer:
recheck

pqcfox (approved):
Thank you so much for catching this!

mkannwischer:
The CI failures are unrelated to this change. Thanks again @prusnak!

PR #233, opened 2026-06-05
[base] mark mem* operations as nonnull 

mkannwischer:
- Depends on #232 as this would have flagged that bug

Annotate memcpy/memset/memcmp/memrcmp/memchr/memrchr with
`__attribute__((nonnull))` via a new `OT_NONNULL` macro, so the compiler
rejects null pointer arguments at compile time under
-Wnonnull (already on via -Wall -Werror). This catches bugs like the
invalid ECDH keyblob zeroing fixed in the previous commits.

Also remove memcpy's "if (dest == NULL || src == NULL)
return dest;" guard. It conflicted with the nonnull attribute.
That check hides bugs instead of serving a real purpose.

### DIFF
--- a/sw/device/tests/penetrationtests/firmware/fi/cryptolib_fi_asym_impl.c
+++ b/sw/device/tests/penetrationtests/firmware/fi/cryptolib_fi_asym_impl.c
@@ -667,8 +667,6 @@ status_t cryptolib_fi_p256_ecdh_impl(
   uint32_t private_keyblob[kPentestP256MaskedPrivateKeyWords * 2];
   memset(private_keyblob, 0, sizeof(private_keyblob));
   memcpy(private_keyblob, uj_input.private_key, P256_CMD_BYTES);
-  memcpy(private_keyblob + kPentestP256MaskedPrivateKeyWords, 0,
-         P256_CMD_BYTES);
   otcrypto_blinded_key_t private_key = {
       .config =
           {
@@ -893,8 +891,6 @@ status_t cryptolib_fi_p384_ecdh_impl(
   uint32_t private_keyblob[kPentestP384MaskedPrivateKeyWords * 2];
   memset(private_keyblob, 0, sizeof(private_keyblob));
   memcpy(private_keyblob, uj_input.private_key, P384_CMD_BYTES);
-  memcpy(private_keyblob + kPentestP384MaskedPrivateKeyWords, 0,
-         P384_CMD_BYTES);
   otcrypto_blinded_key_t private_key = {
       .config =
           {
--- a/sw/device/lib/base/macros.h
+++ b/sw/device/lib/base/macros.h
@@ -448,6 +448,12 @@ extern "C++" {
 #define OT_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
 
 /**
+ * Marks all pointer parameters as non-null so the compiler warns (`-Wnonnull`)
+ * when a NULL/0 argument is passed.
+ */
+#define OT_NONNULL __attribute__((nonnull))
+
+/**
  * Attribute for weak functions that can be overridden, e.g., ISRs.
  */
 #define OT_WEAK __attribute__((weak))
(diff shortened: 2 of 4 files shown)


mkannwischer:
@qmn @pqcfox @fragglet, I'd appreciate your thoughts on this change. Does anyone see a reason to keep the NULL-check in `memcpy`? Keeping it could hide similar bugs as the one spotted here.

Note that you cannot have `__attribute__((nonnull))` AND check for NULL, because the former tells the compiler that the pointers cannot be NULL so it would optimize away the NULL-check (and it warns about it via `-Wtautological-compare`).

fragglet:
> @qmn @pqcfox @fragglet, I'd appreciate your thoughts on this change. Does anyone see a reason to keep the NULL-check in `memcpy`? Keeping it could hide similar bugs as the one spotted here.

To the best of my knowledge, it's undefined behavior what happens if you pass a `NULL` pointer to `memcpy`, `memmove`, etc. I'm not sure why there's this special case in the current implementation - it seems counterproductive. Better to allow the compiler to detect cases like these.

One minor detail you might want to consider with this PR is the one of `memcpy(x, NULL, 0)`. Sometimes you can have a case like this:
```c
uint8_t *mybuf = NULL;
size_t mybuf_len = 0;

if (something) {
   mybuf_len = 1234;
   mybuf = malloc(mybuf_len);
   ...
}

// Sometimes we have something to add, but if not, mybuf might be NULL
memcpy(dest, mybuf, mybuf_len);
```
[A recent proposal for the next C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf) makes it valid to pass `NULL` in this case but presumably this PR could lead to spurious warnings.

mkannwischer:
> > @qmn @pqcfox @fragglet, I'd appreciate your thoughts on this change. Does anyone see a reason to keep the NULL-check in `memcpy`? Keeping it could hide similar bugs as the one spotted here.
> 
> To the best of my knowledge, it's undefined behavior what happens if you pass a `NULL` pointer to `memcpy`, `memmove`, etc. I'm not sure why there's this special case in the current implementation - it seems counterproductive. Better to allow the compiler to detect cases like these.
> 
> One minor detail you might want to consider with this PR is the one of `memcpy(x, NULL, 0)`. Sometimes you can have a case like this:
> 
> ```c
> uint8_t *mybuf = NULL;
> size_t mybuf_len = 0;
> 
> if (something) {
>    mybuf_len = 1234;
>    mybuf = malloc(mybuf_len);
>    ...
> }
> 
> // Sometimes we have something to add, but if not, mybuf might be NULL
> memcpy(dest, mybuf, mybuf_len);
> ```
> 
> [A recent proposal for the next C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf) makes it valid to pass `NULL` in this case but presumably this PR could lead to spurious warnings.

I don't think this will lead to spurious warnings as it will only warn if it can statically determine that the pointer is `NULL`. [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=string/string.h;h=743395b3e3b2dd52194b1a6b8c34df4f8b56d8a1;hb=HEAD#l47) also uses `nonnull` , so they would have the same problem?

fragglet:
Yep, to be clear I'm happy with this PR as it stands - just wanted to note a few details.

fragglet (approved):

sw/device/tests/penetrationtests/firmware/fi/cryptolib_fi_asym_impl.c
```diff
@@ -667,8 +667,6 @@ status_t cryptolib_fi_p256_ecdh_impl(
   uint32_t private_keyblob[kPentestP256MaskedPrivateKeyWords * 2];
   memset(private_keyblob, 0, sizeof(private_keyblob));
   memcpy(private_keyblob, uj_input.private_key, P256_CMD_BYTES);
-  memcpy(private_keyblob + kPentestP256MaskedPrivateKeyWords, 0,
```
fragglet, line 670: Are these supposed to be `memset`s?

prusnak: No. Memset which zeroes the whole structure is 2 lines above.

fragglet: Quite so, thanks.

PR #235, opened 2026-06-08
[acc,util] Add endloop pseudo-op for loop-body size checks

mkannwischer:
acc_as gains an optional, assembler-only endloop psuedo-instruction for the end of a LOOP/LOOPI body.
When present, acc_as emits an as check that the
body is exactly bodysize instructions long (comparing the byte distance to the loop's declared bodysize), so a miscounted loop fails to assemble. A stray endloop with no open loop is reported by acc_as directly.

endloop is opt-in: existing code that does not use it assembles unchanged. The check is done by the assembler on the final byte layout, so it remains correct even when using macros or assembly directives like .rept.

Mismatching body counts are printed as
```
error: loop: loop body size does not match declared bodysize (2)
```

### DIFF
--- a/hw/ip/acc/data/base-insns.yml
+++ b/hw/ip/acc/data/base-insns.yml
@@ -636,6 +636,16 @@
     - A `LOOP` error if `iterations` is zero.
     - *loop-at-end
 
+- mnemonic: endloop
+  synopsis: Mark end of a loop body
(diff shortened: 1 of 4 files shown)


pqcfox (approved):
Looks excellent, thanks so much for adding this!

PR #237, opened 2026-06-09
[docker] add arm64 support

### LINKED ISSUE #229: [docker] arm64 support
### Description

I tried to update `util/container/Dockerfile` to support arm64. This makes development possible on Apple Silicon devices.

I have the changes ready - see https://github.com/prusnak/pavona/commit/70a74e345eab292811a15f5778cbff36d5f487a4, however there is one outstanding issue:
- Dockerfile installs Verilator from `https://storage.googleapis.com/verilator-builds/verilator-v4.210.tar.gz` which is a x86_64 build

All other components support both archs (x86_64 and arm64).

Is there a verilator build in that storage bucket for arm64?
Does the build contain some pavona-specific patches?

If not, we could bump the container to Ubuntu 24.04 LTS which contains Verilator 5.x in its repository and we could switch to the official Ubuntu build - see https://github.com/prusnak/pavona/commit/fef32e2596e97bbf3bf9e4e2a8f2ae4178e2f827

prusnak:
fixes https://github.com/pavona/pavona/issues/229

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -102,6 +102,24 @@ rust.repository_set(
     versions = ["1.90.0"],
 )
 rust.repository_set(
+    name = "rust_host_aarch64",
+    edition = "2024",
+    exec_triple = "aarch64-unknown-linux-gnu",
+    sha256s = {
+        "rustc-1.90.0-aarch64-unknown-linux-gnu.tar.xz": "4e1a9987a11d7d91f0d5afbf5333feb62f44172e4a31f33ce7246549003217f2",
+        "clippy-1.90.0-aarch64-unknown-linux-gnu.tar.xz": "1a7f969c5712c17d960d6fba8235b33f264a9484b369dc388766a540f1ba464f",
+        "cargo-1.90.0-aarch64-unknown-linux-gnu.tar.xz": "bd8d1da6fe88ea7e29338f24277c22156267447adbfc47d690467ad32d02c2a7",
+        "llvm-tools-1.90.0-aarch64-unknown-linux-gnu.tar.xz": "3f559ee41cf201a818630b1877e53ee0b9cd32793ee99c297ef56225a8aeafd6",
+        "rust-std-1.90.0-aarch64-unknown-linux-gnu.tar.xz": "4952abb7d9d3ed7cea4f7ea44dcb23dc67631fae4ac44a5f059b90a4b5e9223f",
+    },
+    target_compatible_with = [
+        "@platforms//cpu:aarch64",
+        "@platforms//os:linux",
+    ],
+    target_triple = "aarch64-unknown-linux-gnu",
+    versions = ["1.90.0"],
+)
+rust.repository_set(
     name = "rust_tock",
     edition = "2021",
     exec_triple = "x86_64-unknown-linux-gnu",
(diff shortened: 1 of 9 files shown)


qmn:
Hi Pavol, thanks so much for getting in arm64 support! It looks like lint failure is due to an out-of-date MODULE.bazel.lock. Could you please run `$ ./bazelisk.sh mod deps`, and re-push? That should fix the "Lint (quick)" job.

Would you be able to comment on the changes to gpiodpi.c as well as status.rs? Did implementing arm64 support expose the need for these changes?

prusnak:
> Would you be able to comment on the changes to gpiodpi.c as well as status.rs?

These changes were done by @mkannwischer - can you please comment on your portion of changes? I will then add your description in the commit message.

mkannwischer:
Both changes were needed, yes. I have tested it in a docker container on MacOS.

status.rs assumed that char is signed which true on x86_64, but its unsigned on aarch64.


The change in gpiodpa.c adds missing initialization. It seems we got lucky on x86 it was 0 anyway - that wasn't the case on arm64.

Would be nice to move those changes into a separate commit each and describe the problem they are fixing. Just in case someone needs to understand this in a few years from now. @prusnak, could you do that, please? Otherwise I can do it, but I will likely only get to it next week.

util/container/Dockerfile
```diff
@@ -17,12 +16,17 @@ ARG CLANG_VERSION=16
 
 # Main container image.
 FROM ubuntu:22.04 AS pavona
```
mkannwischer, line 18: Lets switch to 24.04 while we are at it?

mkannwischer (changes requested):

mkannwischer:
I think it would be good to also add Arm support to `./bazelisk.sh`.
It would be nice if the steps in the [Getting started](https://docs.pavona.org/book/doc/getting_started/index.html) just work on a aarch64-linux sysem (say Ubuntu 24).

mkannwischer:
I've continued on this a little bit, and I'm testing it here: https://github.com/mkannwischer/pavona/pull/1

MODULE.bazel
```diff
@@ -398,8 +447,6 @@ use_repo(
 
 register_toolchains("//rules/pavona:localtools")
 
-register_toolchains("//toolchain:cc_toolchain_opentitan")
```
rtorok-zr, line 401: Removing this will cause the device software builds to break. I think the fix you made in `toolchain/BUILD` should address the issue with building from ARM64 hosts, so you should be able to add this back.

hw/dv/dpi/gpiodpi/gpiodpi.c
```diff
@@ -227,6 +227,8 @@ svBitVecVal gpiodpi_host_to_device_tick(void *ctx_void, svBitVecVal *gpio_oe,
   struct gpiodpi_ctx *ctx = (struct gpiodpi_ctx *)ctx_void;
   assert(ctx);
 
+  *req_exit = 0;
```
rtorok-zr, line 230: This does not seem related to ARM64 support. Was this intended to be in a different PR?

rules/fusesoc.bzl
```diff
@@ -116,7 +116,9 @@ def _fusesoc_build_impl(ctx):
                 "HOME": home_dir,
                 # Obtain the non-hermetic binary path and append Bazel's default PATH.
                 "PATH": (BIN_PATHS["vivado"] + ":" if ctx.attr.target == "synth" else "") + "/bin:/usr/bin:/usr/local/bin",
-                "LD_PRELOAD": "/lib/x86_64-linux-gnu/libudev.so.1",
+                "LD_PRELOAD": "/lib/{}/libudev.so.1".format(
```
rtorok-zr, line 119: This is nice to have, but is probably not needed, because this `LD_PRELOAD` is specific to Vivado, which only supports x86 platforms.

hw/dv/dpi/gpiodpi/gpiodpi.c
rtorok-zr: Edit: please disregard. I see @mkannwischer 's comment about this above.

prusnak:
> I've continued on this a little bit, and I'm testing it here: [mkannwischer#1](https://github.com/mkannwischer/pavona/pull/1)

Feel free to take over. I will close this PR once a PR superseding it will be opened.

prusnak:
superseded by https://github.com/pavona/pavona/pull/248

PR #239, opened 2026-06-09
[crypto] Add endloop markers to Ed25519 implementation

pqcfox:
This PR adds the `endloop` pseudo-instruction (see #235) to the ends of each loop in the Ed25519 ACC implementation, ensuring that the loops indeed end where suggested by indentation in the code.

PR #240, opened 2026-06-10
[acc,util] Emit .option exact to stop assembler branch relaxation

mkannwischer:
clang's integrated assembler treats RISC-V conditional branches as relaxable: any beq/bne/... whose target it cannot prove to be within the +-4 KiB branch range (in particular, branches to symbols defined in another object) is silently rewritten into an inverted branch plus a jal.

This rewriting changes the emitted instruction count behind our back, which is a real hazard for ACC hardware loops: loop/loopi encode the body size in instructions, so if a conditional branch inside a loop body is expanded to two instructions, the hardware's loop-end address no longer matches the body the programmer wrote and execution silently goes wrong. It also makes any conditional branch a variable-size fragment until layout, so expressions like `. - label` spanning a branch no longer evaluate as absolute at parse time; this showed up in https://github.com/pavona/pavona/pull/239, where the loop-body size check that the endloop marker emits could not be placed after a loop body containing a beq. Finally, rsa.s, rsa_keygen.s, p256_verify.s, p256_ecdsa_sca.s, p384_verify.s and p384_ecdsa_sca.s all already contained such silently doubled branches (to external symbols).

Emit `.option exact` in the preamble of every translated file to disable the relaxation: branches are now emitted exactly as written.

`.option exact` is supported since LLVM 21.

### DIFF
--- a/hw/ip/acc/util/acc_as.py
+++ b/hw/ip/acc/util/acc_as.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


pqcfox (approved):
Looks great!

PR #241, opened 2026-06-10
[acc,mlkem] Minor fixes for ML-KEM assembly (DMEM initialization, layout assumptions)

mkannwischer:
In the process of testing the ML-KEM ACC implementation on FPGA, I hit a few minor issues that I fixed here.

- Depends on #214


pqcfox (approved):
Looks great! Thanks for catching these!

PR #244, opened 2026-06-10
[dv] Fix test failures

matutem:
This PR has several independent commits that fix several test failures, for both top-level and block level tests.
Please refer to the individual commits for details.


ysekhar (approved):
There are a few verible lint issues to resolve. Otherwise LGTM.

PR #188, opened 2026-05-23
[dj,sw] Enable Darjeeling FPGA bootstrapping

rtorok-zr:
This PR enables Darjeeling execution on FPGA platforms:
- Adds support for splicing `rom1` and OTP images using `updatemem`.
- Adds a new implementation of the bootstrapping interface in the host-side tooling to enable loading software images into CTN SRAM using JTAG/OpenOCD.
- Adds build tooling for building Darjeeling OTP images with ROM_EXT signing keys.

Dependent on #186

PR #245, opened 2026-06-11
[df,fpga] Enable FPGA software execution for integrated top-level designs

rtorok-zr:
This PR enables software execution for integrated top-level designs on Xilinx FPGA platforms:
- Adds support for splicing `rom1` and OTP images using `updatemem`.
- Adds a new implementation of the bootstrapping interface in the host-side tooling to enable loading software images into CTN SRAM using JTAG/OpenOCD.
- Adds build tooling for building `top_dragonfly` OTP images with ROM_EXT signing keys.

Replaces #188


rules/pavona/exec_env.bzl
```diff
@@ -393,20 +400,29 @@ def common_test_setup(ctx, exec_env, firmware):
...
         openocd = exec_env.openocd
         jtag_data = [openocd]
-        jtag_test_cmd = '''
-            --openocd="$(rootpath {})"
-        '''.format(openocd.label)
+        openocd_arg = '--openocd="$(rootpath {})"'.format(openocd.label)
+        param["openocd"] = openocd_arg
+        jtag_test_cmd = """
+            {}
+        """.format(openocd_arg)
```
qmn: Seems like this could simply be
```suggestion
        jtag_test_cmd = openocd_arg
```
?

rules/pavona/splice.bzl
```diff
@@ -14,16 +14,20 @@ load("@nonhermetic//:env.bzl", "BIN_PATHS", "ENV")
 
 DEFAULTS = struct(
     rom = "//hw/bitstream/universal:none",
+    second_rom = "//hw/bitstream/universal:none",
     otp = "//hw/bitstream/universal:none",
     env = "//hw/bitstream/universal:none",
 )
 
-def gen_vivado_mem_file(ctx, name, src, tool, otp_size, swap_nibbles = True):
+def gen_vivado_mem_file(ctx, name, src, tool, otp_size = 1024, otp_padding = 0, swap_nibbles = True, swap_crumbs = False):
```
qmn, line 22: Not sure that exactly this line is the best place for it, but is there an explanation somewhere about what `swap_crumbs` does? It was introduced in b2a54ef43968c007f413a40380a8944e1e9d3151:

https://github.com/pavona/pavona/blob/b2a54ef43968c007f413a40380a8944e1e9d3151/hw/ip/rom_ctrl/util/gen_vivado_mem_image.py#L194-L197

In plain English this seems to be: a nibble (4 contiguous bits) is subdivided into two crumbs (2 bits each). To swap these crumbs is to swap the upper two bits with the lower two bits.

So for example, if I have a nibble with bits `W X Y Z` then the crumb-swapped version is `Y Z W X`.

Would that be right? It would be nice to have an explanation for what it does, in the code, and why it's necessary to do at all.

qmn (commented):
Thanks so much for the crucial work. One small suggestion and a request for some explanation of "crumbs".

rules/pavona/splice.bzl
rtorok-zr: Your understanding is correct. I added an explanation of this in `splice.bzl`.

qmn (approved):
LGTM, thanks for the fixes!

PR #248, opened 2026-06-12
AArch64 support

### LINKED ISSUE #229: [docker] arm64 support

mkannwischer:
This pull request adds AArch64 support for bazelisk.sh and the toolchains for verilated tests, device software, etc. It also updates the Dockerfile accordingly (which allows running on AArch64 MacOS without emulating x86_64). Two source changes were needed to make the verilated tests work on AArch64 (see the first two commits).

This PR also adds CI to test the setup natively on Ubuntu 22, 24, 26 (x86_64 + AArch64) and in Docker (x86_64 + AArch64). That required a few changes to how we install dependencies in CI.
The libtinfo5 (Ubuntu 24+26) and libxml2 (Ubuntu 26) are somewhat nasty and we should try to find a better solution soon (or at the very least document the workarounds somewhere so users can find them). 

I did intentionally not add anything to the docs yet regarding aarch64 as I wanted to discuss that first. Currently we do not say anything about supported architectures. I think it would be nice to describe which parts are expected to work/not work on AArch64. 

Continuation of #237 
- Resolves #229


mkannwischer:
Coincidentally Github has made Ubuntu 26 Action runners available today: https://github.com/actions/runner-images/issues/14226. I've added CI tests against it as well.

mkannwischer:
@prusnak could you try this out please (I removed the toolchain from the Dockerfile) if this works for you?

prusnak:
> @prusnak could you try this out please (I removed the toolchain from the Dockerfile) if this works for you?

I successfully built the container and run bazel smoke test which (among other things) downloaded the llvm toolchain.

Idea: currently when i use the container, it always downloads bazel and llvm toolchain again and again. Can we tweak the dockerfile so this is pre-cached already?

mkannwischer:
> > @prusnak could you try this out please (I removed the toolchain from the Dockerfile) if this works for you?
> 
> I successfully built the container and run bazel smoke test which (among other things) downloaded the llvm toolchain.

Nice! Thanks for trying.

> 
> Idea: currently when i use the container, it always downloads bazel and llvm toolchain again and again. Can we tweak the dockerfile so this is pre-cached already?

I think you should be able to map the cache to some persistent location outside of the container, then it should only do it once. But I agree it would be much more convenient if this happens during building the container and the user does not have to worry about it. As this PR is already massive, I would like to do that as a follow-up.

qmn (approved):
It's indeed the case that we haven't specified the full compatibility, but I think arm64 support is a good direction. Thanks @prusnak and @mkannwischer for your work here.

This all LGTM, I'm okay with the hacking around libtinfo5 and libxml2 for now. I'll open issues for these.

mkannwischer:
Thanks for the review @qmn!
I've rebased on top of main and resolved the bazel lockfile conflict.

dop-amin (approved):
Looks good to me!

I've tested this using Docker version 25.0.2, build 29cf629 on macOS 26.4.1 (25E253). 
Setup was smooth, all tests I've tried work. 

However, there was one warning emitted upon running bazel initially:
```
./bazelisk.sh build sw/device/examples/hello_world:hello_world
fatal: detected dubious ownership in repository at '/home/dev/src'
To add an exception for this directory, call:

    git config --global --add safe.directory /home/dev/src
```

I recall the same warning from before when using Docker on x86_64, so I suppose this is not a regression of this PR.

mkannwischer:
> Looks good to me!
> 
> I've tested this using Docker version 25.0.2, build 29cf629 on macOS 26.4.1 (25E253). Setup was smooth, all tests I've tried work.
> 
> However, there was one warning emitted upon running bazel initially:
> 

Thanks @dop-amin for trying it out. I have opened an issue for the warning you got: https://github.com/pavona/pavona/issues/276.

mkannwischer:
> > @prusnak could you try this out please (I removed the toolchain from the Dockerfile) if this works for you?
> 
> I successfully built the container and run bazel smoke test which (among other things) downloaded the llvm toolchain.
> 
> Idea: currently when i use the container, it always downloads bazel and llvm toolchain again and again. Can we tweak the dockerfile so this is pre-cached already?

Opened an issue for this too: https://github.com/pavona/pavona/issues/277

PR #252, opened 2026-06-12
[csrng,cov] Exclude aes_cipher_core coverage

matutem:
The aes_cipher_core is more extensively exercised by the aes block level DV, and the aes embedded in csrng is constrained to use a subset of aes functionality.

This exclusion is already implemented for xcelium in file hw/ip/csrng/dv/cov/csrng_v2s_exclusions.vRefine, and this just provides a similar exclusion for VCS.


ysekhar (approved):
LGTM

PR #254, opened 2026-06-15
[acc,doc] Clean up and correct ACC vector instruction docs.

phamhnh:
Tidy up the documentation and encoding for the ACC instructions:

- Standardize operand descriptions (e.g. "Name of the destination WDR" -> "Destination WDR").
- Lowercase all type-suffix enums (`.8S`/`.16H` -> `.8s`/`.16h`, etc.) so the documented syntax matches the assembler.
- Rewrite the vector instruction docs (`bn.addv`, `bn.subv`, `bn.shv`, `bn.trn`, `bn.mulv`, `bn.mulv.l`) to describe their actual behaviour, removing stale duplicated "If <vec> is set to 1" blocks and TODOs.
- Correct `bn.trn` to describe the NEON-style `TRN1/TRN2` transpose instead of the old lane-write text, and fill in the missing `bn.mulv/bn.mulv.l` descriptions.
- Fix assorted typos (`wrd'`, `bn.wsrw` "read"->"write", duplicate operands line) and add missing empty errs lists.
- Drop the unused `shift_arith` operand from `bn.shv` (the shift is logical only) and update `enc-schemes.yml` to mark bit 15 fixed. `insn.py` is also updated to remove `shift_arith` argument.

### DIFF
--- a/hw/ip/acc/data/bignum-insns.yml
+++ b/hw/ip/acc/data/bignum-insns.yml
@@ -17,11 +17,11 @@
   synopsis: Add
   operands: &bn-add-operands
     - name: wrd
-      doc: Name of the destination WDR
+      doc: Destination WDR
     - name: wrs1
-      doc: Name of the first source WDR
+      doc: First source WDR
     - name: wrs2
-      doc: Name of the second source WDR
+      doc: Second source WDR
     - &bn-shift-type-operand
       name: shift_type
       abbrev: st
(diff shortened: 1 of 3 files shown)


apinise (approved):
LGTM! Thanks for the updates @phamhnh, it's much more concise/readable with the removal of old text and clean up.

mkannwischer (approved):

PR #255, opened 2026-06-15
[ci] Don't depend on lsb_release

rtorok-zr:
Newer versions of RHEL/AlmaLinux do not offer `lsb_release` in a `dnf` package, so we use the more portable `/etc/os-release` to tag Bazel cache entries.


qmn (approved):
I tried this on Ubuntu and AlmaLinux both and so LGTM.

PR #256, opened 2026-06-15
[sw,tests] Fix all_escalation_resets test

matutem:
Instruct OTTF to skip configuring alerts, since they can get in the way of the alert configuration needed for this test.


LouisTheLuis (approved):
LGTM, merge after fixing lint.

PR #259, opened 2026-06-16
[acc,util] Use numeric local labels for endloop body markers

mkannwischer:
If a loop/loopi was placed within a .rept or repeatedely called .macro, the labels enabling the `endloop` check introduced in 8459b12ca2132b670f14e5d2ae3acd3659b6ffe1
would result in "symbol already defined" errors.
Switch to numeric labels instead as those may be redefined.


pqcfox (approved):
Nice fix on this!

PR #260, opened 2026-06-16
[acc,kmac] Fixes bugs and optimizes the eager KMAC refresh in ACC

apinise:
This PR addresses issues with the eager refresh ACC uses during transactions with KMAC, along with miscellaneous bugs.

#### Previous Eager Refresh
ACC previously requested the next digest from KMAC during XOF operations as soon as the previous digest had been read. This prevented stalls incurred by `bn.wsrr` instructions reading from the `DigestShare0` or `DigestShare1` registers. There was an issue identified where the eager refresh would occur after the previous read exhausted the Keccak state.

This forced KMAC into a manual run and an additional ~96 cycle delay. During this time if ACC ended the transaction and attempted to start a new one, the `hold` signal was dropped and reasserted before KMAC could finish its manual run and return to a wait state.

#### New Eager Refresh
ACC now tracks the Keccak state and output buffer to determine if an eager refresh will lead to a manual run or new permutation. When a `next` signal on the interface triggers a manual run, ACC will not assert it until there is an explicit attempt to read the digest using a `bn.wsrr` instruction. All other scenarios in the interface remain the same. KMAC has introduced an Eager Wait state as a safeguard, transitioning anytime `hold` is dropped during an ACC <-> KMAC transaction, forcing ACC to stall before starting the next transaction.

#### Bug Fix
Fixed an issue with undersized message error reporting due to unconditioned sampling of the ISPR address as `DigestShare0` or `DigestShare1`. Sampling is now only valid while `ispr_rd_en_i` is `1`.

### DIFF
--- a/hw/ip/acc/dv/accsim/sim/kmac.py
+++ b/hw/ip/acc/dv/accsim/sim/kmac.py
@@ -129,6 +129,7 @@ class KmacBlock:
         self._digest1_ready_next = False
         self._digest0_read = False
         self._digest1_read = False
+        self._pending_read = False
         self._unmasked_digest = bytes()
         self._apply_mask = False
         self._finished_digest0 = False
@@ -137,7 +138,6 @@ class KmacBlock:
         self._shift_digest_reg = False
         self._new_permutation = False
         self._digest_request_ctr = 0
-        self._skip_digest_shift_cycle = False
         self._kmac_undersized_err = False
         self._kmac_oversized_err = False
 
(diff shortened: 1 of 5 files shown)


mkannwischer (approved):
Thanks @apinise. I have tried out this patch and confirmed that this allowed me to match the ACVP and Wycheproof testvectors for the ML-KEM and ML-DSA implementations on the CW340.

PR #263, opened 2026-06-16
[top_dragonfly,rom1] Add second_rom arg for dragonfly BUILD

matutem:
- Add "second_rom" attribute in exec_env_common_attrs in exec_env.bzl.
- Add test_second_rom image at slot 7 in the `sw_test_mode_test_rom` mode, so any dragonfly sim_dv test that uses that mode will also use test_second_rom.
- Changed some comments in chip_dragonfly_sim_cfg.hjson.


rtorok-zr (approved):
LGTM

PR #264, opened 2026-06-16
[yum] Fix yum-requirements for AlmaLinux 10

qmn:
Install epel-release first (in prepare-env) to make libftdi and libftdi-devel available to install in the "Install individual packages" step. Then, additionally install `bc` and `time`, which are used by VCS in the dvsim regressions.


LouisTheLuis (approved):
LGTM

PR #267, opened 2026-06-18
[pwrmgr,dv] Properly connect up pwrmgr_sec_cm_checker_assert

matutem:
I think the author of https://github.com/lowRISC/opentitan/pull/28474 hadn't noticed that the esc_timeout port gets supplied with esc_timeout_lc_q in the bind statement in pwrmgr_bind.sv

Connecting it up properly will silence warnings from e.g. Xcelium that we are not supplying all ports in the bind.

This *might* cause test failures: the timeout can now actually be asserted. But any test that was previously working because of this was only working because it wasn't seeing the error.


LouisTheLuis (approved):
Seems like just renames. LGTM.

matutem:
And the `esc_timeout_lc_q` port is removed, which was previously unconnected so was resulting in unknown ('x or 'z) values.

PR #269, opened 2026-06-22
[acc,mlkem] Add ACC-backed ML-KEM to the cryptolib

mkannwischer:
- Depends on #241 

This PR wires up the ML-KEM ACC implementation into the cryptolib. It adds a new bazel config option `acc_has_pqc` that defaults to false. When it's set the ACC implementation is used instead of mlkem-native. Currently this has to be manually set, but ultimately this should be derived from the top configuration. Since there is currently no pre-built bitstream with PQC-enabled ACC available, the ACC path is not currently exercised in CI. The main purpose of this PR is to enable running full ACVP + Wycheproof testvectors for the ACC ML-KEM implementation.

Below are instructions to test both paths in Verilator and on the CW340:

mlkem-native:
```
# verilated functional test 
./bazelisk.sh test //sw/device/tests/crypto:mlkem_functest_sim_verilator --test_output=streamed

# ACVP + Wycheproof tests on the CW340 (pre-built bitstream)
./bazelisk.sh test //sw/device/tests/crypto/cryptotest:mlkem_kat_fpga_cw340_test_rom --test_arg=--usb-serial=<usb-serial> --test_output=streamed
```

ACC:
```
# set AccPQCEn=1
git fetch https://github.com/mkannwischer/pavona acc-pqcen
git cherry-pick 4ac8abbfd696408e9c16c6def644335257d3ff38

# verilated functional test (with acc_has_pqc=true)
./bazelisk.sh test //sw/device/tests/crypto:mlkem_functest_sim_verilator --test_output=streamed  --define=acc_has_pqc=true

# ACVP + Wycheproof tests on the CW340 (with acc_has_pqc=true)
./bazelisk.sh test //sw/device/tests/crypto/cryptotest:mlkem_kat_fpga_cw340_test_rom  --test_arg=--usb-serial=2055355B3236 --test_output=streamed --define bitstream=vivado --define=acc_has_pqc=true
```

`<usb-serial>` should be 2055355B3236 in case you are running this on the shared workstation.


mkannwischer:
@qmn, @apinise, may I please ask you for a glimpse at the `acc_has_pqc` config option that switches between the ACC backend and the mlkem-native backend (see https://github.com/pavona/pavona/pull/269/commits/1fd0a2f2b72af52fac1d2c1c2fb2ad5f42d0db4e and [BUILD](https://github.com/mkannwischer/pavona/blob/5f35b805bb091326ebe0b25d5bc9e1677874a55f/sw/device/lib/crypto/impl/BUILD)). Is this approach compatible with how we want this to ultimately look like when it's derived from the top config?

sw/device/lib/crypto/impl/mlkem.c
```diff
@@ -79,18 +85,24 @@ otcrypto_status_t otcrypto_mlkem512_keygen_derand(
   uint32_t *sk_share0;
   uint32_t *sk_share1;
   HARDENED_TRY(keyblob_to_shares(secret_key, &sk_share0, &sk_share1));
-  memset(sk_share1, 0, MLKEM512_SECRETKEYBYTES);
+  memset(sk_share1, 0, kOtcryptoMlkem512SecretKeyBytes);
 
+#ifdef ACC_HAS_PQC
+  (void)work;
+  HARDENED_TRY(mlkem_acc_512_keygen(randomness.data, (uint8_t *)public_key->key,
+                                    (uint8_t *)sk_share0));
```
mkannwischer: The APIs should be changed to uint32_t *, so that no cast is needed here.

mkannwischer: Fixed by using `word32_buf_t` for the ML-KEM API (in line with P256, etc).
Note that this is possible in ML-KEM because all sizes neatly divide by 4. For ML-DSA this doesn't work because sizes are not divisible by 4).

sw/device/lib/crypto/include/mlkem.h
```diff
@@ -73,16 +109,16 @@ otcrypto_status_t otcrypto_mlkem512_keygen(
  * blob for the private key should have a length of 2x
  * ceil(kOtCryptoMlkem512SecretKeybytes / sizeof(uint32_t)) = 816 words.
  *
- * @param randomness seed for keygen, len `kOtcryptoMlkem512KeygenSeedBytes`.
+ * @param randomness seed for keygen, len 16 words.
```
pqcfox: (more curiosity) how come this was changed to a concrete value here + below?

pqcfox (approved):

sw/host/cryptotest/testvectors/parsers/nist_acvp_mlkem_parser.py
```diff
@@ -16,12 +16,8 @@
...
 
 ACVP encapsulationKeyCheck and decapsulationKeyCheck tests verify that
-the implementation rejects invalid keys. The cryptolib does not expose
-dedicated key validation functions as of now, but encapsulate_derand
-and decapsulate perform internal checks (FIPS 203 modulus check,
-hash consistency) that reject invalid keys. We map these tests to
-regular encaps/decaps operations with dummy randomness/ciphertext:
-valid keys succeed, invalid keys cause the operation to fail.
+the implementation rejects invalid keys. These are skipped for now since
+the ACC path does not implement key validation yet.
```
pqcfox, line 20: (could be also worth adding issues to track the KAT skips)

sw/device/lib/crypto/include/mlkem.h
mkannwischer: "`kOtcryptoMlkem512KeygenSeedWords` words" overflowed into the next line when formatted which looked ugly. But maybe just dropping the `words` is better. Let me change that.

mkannwischer: done.

PR #273, opened 2026-06-23
[dvsim] Support variants in --select-cfgs

LouisTheLuis:
This PR is composed of three commits:
- The first one includes a change that makes it possible for the `dvsim` `--select-cfgs` flag to differentiate between two variants of the same block. Before this change, the `--select-cfgs` option automatically selected all blocks with the same name and ran them (e.g. `--select-cfgs pwrmgr` would run both `pwrmgr-top-egret` and `pwrmgr-top-dragonfly`). Now by passing the variant, we can specify which one needs to be ran.
- The second one includes the `variant` field for `xbar` sim_cfgs. This is done so that the first change can also support choosing the specific `xbar` the user wants to be ran with the `--select-cfgs` flag.
- The third one fixes the `sim_cfg` check in CI to allow `xbar`s to have variants.


util/dvsim/SimCfg.py
```diff
@@ -244,6 +242,16 @@ def _purge(self):
         log.info("Purging scratch path %s", self.scratch_path)
         rm_path(self.scratch_path)
 
+    def _is_selected_cfg(self):
+        '''Returns whether this cfg is selected via --select-cfgs'''
+        if not self.select_cfgs:
```
adurbin-zr, line 247: What is the logic behind this being True? select_cfgs was not set so 'this cfg is selected view --select-cfgs'?  Could you please add a comment for this branch?

util/dvsim/FlowCfg.py
```diff
@@ -377,7 +377,17 @@ def prune_selected_cfgs(self):
             sys.exit(1)
 
         # Filter configurations
-        self.cfgs = [c for c in self.cfgs if c.name in self.select_cfgs]
+        filtered_cfgs = []
+        for c in self.cfgs:
+            for s in self.select_cfgs:
+                if s == c.name:
```
adurbin-zr, line 383: select_cfgs can both be a string and a collection?

LouisTheLuis: It's always a string, but the string can take the shape of a name (e.g. `pwrmgr`) or a name and a variant (e.g. `pwrmgr_dragonfly`).

util/dvsim/SimCfg.py
LouisTheLuis: If it was not set, `dvsim` still needs to create the object for it; that's the gist. Added a comment.

util/dvsim/FlowCfg.py
adurbin-zr: I see. A comment might be helpful for a future reader.

adurbin-zr: also, if that's the case, shouldn't we be more defensive to ensure the string is at the beginning if it's dealing with a variant?

adurbin-zr: Looks like the latter is handled below.

adurbin-zr (approved):

PR #275, opened 2026-06-23
[ci] Fix dvsim run branch name

LouisTheLuis:
The branch name in `dvsim.py` runs substitutes backlashes for dashes for filesystem purposes (so that the generated scratch directory is valid). The CI should follow this scheme, as it is the one followed by `dvsim.py`.


vlpannel (approved):
LGTM.

PR #279, opened 2026-06-24
[topgen] Break topgen outputs into different scripts

vlpannel:
Currently the topgen tool is bogged down by a large number of argument flags that make it more difficult to use and comprehend. This PR reduces `util/topgen.py` to only generating the complete top configuration, which other tools may then use to produce further outputs.

Additionally, new scripts that can adapt the complete top configuration into new outputs are:
* `util/gen_top_ral.py`: generates the top level's RAL model
* `util/top_cm_and_blocks.py`: depending on subcommand used, either checks countermeasures of all a top's IPs or lists which blocks are present within a top level
  * This script's list of blocks can be fed to regtool to generate the registers of all top-specific IPs
* `util/gen_top_ipconfigs.py`: creates IP configuration Hjson files that can be used to generate templated IPs or crossbars from top level information
  * In addition, ipgen and tlgen have been revised to be able to use the output of this script
* `util/gen_top_docs.py`: generates top level's pinmux and target documentation (placed under the top's pinmux subdirectory)
* `util/gen_top_sw.py`: generates the software (C, Rust, Bazel) files for a top level
* `util/gen_top_sv.py`: generates the SystemVerilog files associated with a top level design

The code in these files is mostly cut and pasted from the original topgen script.

The makefile used for running the ACE flow as a whole (`hw/Makefile`) has also been adjusted accordingly. In a future PR, this makefile should be reorganized for readability, but that is out-of-scope for this PR.


hw/dv/tools/ralgen/ralgen.py
```diff
@@ -56,16 +55,12 @@ def main():
         if alias_hjson:
             args += ["--alias", root_dir / alias_hjson]
     else:
-        ral_spec = root_dir / top_hjson
-        # Extract top_name from ral spec path and create _seed.testing.hjson
-        ral_path = Path(ral_spec)
-        top_name = ral_path.stem
-        seed_path = ral_path.with_name(f"{top_name}_seed.testing.hjson")
```
matutem, line 63: I assume we don't need the seed file since that is only needed by topgen?

hw/dv/tools/ralgen/ralgen.py
```diff
@@ -56,16 +55,12 @@ def main():
...
+        original_spec = root_dir / top_hjson
+        top_name = Path(original_spec).stem
+        ral_spec = original_spec.parent / "autogen" / f"{top_name}.gen.hjson"
 
-        cmd = util_path / "topgen.py"
-        args = [cmd, "-r", "-o", os.getcwd(), "-t", ral_spec, "-s", seed_path]
-        if hjson_path:
-            args += ["--hjson-path", root_dir / hjson_path]
+        cmd = util_path / "gen_top_ral.py"
+        args = [cmd, "-o", os.getcwd(), "-t", ral_spec]
```
matutem: os.getcwd() -> Path('.') everywhere and remove `os` import?

hw/dv/tools/ralgen/ralgen.py
vlpannel: Yes.

util/topgen/c_test.py
```diff
@@ -111,7 +111,7 @@ def _get_irq_peripherals(self, rv_plic, addr_space):
         # There may be multiple PLICs, for example, in one address space.
         # Or devices with core # interfaces in address spaces that are different
         # from the CPU's and PLIC's.
-        rv_plic_addr_spaces = rv_plic['base_addrs'][None]
+        rv_plic_addr_spaces = rv_plic['base_addrs'].get(None) or rv_plic["base_addrs"].get("null")
```
matutem: Please use single quotes for strings, as the pre-existing code does. Here and everywhere else.

matutem: Could we trace the use of 'null' and be consistent? Either use None or 'null'.

util/topgen/c_test.py
```diff
@@ -200,7 +200,7 @@ def _get_irq_peripherals(self, rv_plic, addr_space):
             status_default_mask = 0
             for irq in self.top["interrupt"]:
                 if irq["module_name"] == inst_name:
-                    if irq["intr_type"] == IntrType.Status:
+                    if irq["intr_type"] == IntrType.Status or irq["intr_type"] == "IntrType.Status":
```
matutem: Weird, why the string 'IntrType.Status'? How is that sneaking in?

util/topgen/clocks.py
```diff
@@ -265,6 +265,8 @@ def __init__(self, raw: Dict[str, object]):
             assert isinstance(raw_grp, dict)
             grp = Group(raw_grp, f'clocks.groups[{idx}]')
             self.groups[grp.name] = grp
+            for group_clk, gc_src_name in raw_grp.get("clocks", dict()).items():
```
matutem: Why this change? Was this buggy before this change? Seems unrelated.

util/topgen/README.md
```diff
@@ -128,12 +135,9 @@ It is possible to restrict what the tool should generate.
 $ util/topgen.py --help
 usage: topgen [-h] --topcfg TOPCFG --seedcfg SEEDCFG [--outdir OUTDIR]
               [--hjson-path HJSON_PATH] [--verbose]
-              [--version-stamp VERSION_STAMP] [--no-top] [--no-xbar]
-              [--no-plic] [--no-rust] [--top-only] [--check-cm] [--xbar-only]
-              [--plic-only] [--alert-handler-only] [--rust-only] [--top_ral]
+              [--version-stamp VERSION_STAMP] [--no-plic] [--plic-only]
```
matutem: Why are the plic and alert_handler flags still needed?

util/topgen/c_test.py
vlpannel: The problem is translation between Python and Hjson.

* Translating `None` to Hjson casts it as `null` in JSON/Hjson
* However, JSON object properties (equivalent to keys in Python dicts) must always be strings
* So reading an Hjson back into Python, the dict key is automatically made into `'null'` (str) rather than `None`

Because of this weird `None` (Python) --> `null` (JSON) --> `'null'` (Python) translation, I've chosen to manually search for both values for now.

util/topgen/c_test.py
vlpannel: I assume it's the same thing as my comment on casting above. Because we're now using this to read in the complete topcfg, probably an object using the `IntrType` enum is putting the value in the complete cfg Hjson as a string; I will try to fix this in the `IntrType` class.

util/topgen/README.md
vlpannel: I can discard these flags too. They tend to just stop the hardware generation early, and I don't think that complete config generation is so cumbersome that there is a good enough argument for only generating rv_plic or only generating the alert handler.

util/topgen/clocks.py
vlpannel: This was not buggy before the change, but this change is required for using the complete config.

* Original topgen: resolve the clocking in Python so that the `Clocks` object already knows the mapping, populate complete config with the clocks in each group.
* New: read in the complete config and derive the group clocks from that metadata (needs to go deeper into the dict to find those values).

Compare `hw/top_dragonfly/data/top_dragonfly.hjson` and `hw/top_dragonfly/data/autogen/top_dragonfly.gen.hjson` (specifically the clocks.groups field).

util/topgen/c_test.py
vlpannel: Update: it's because of the function `for_json` in `util/reggen/access.py`, where the IntrType enum gets cast as a string when put into the Hjson.

I fixed it by instead making the `for_json` function return the enum value instead. This reduces the readability of the complete config Hjson, but it also does not require the discussed code kludge.

util/topgen/c_test.py
matutem: This suggests we should avoid `None` as a key in python. I think the default name for unnamed register blocks could be 'regs'. Perhaps this should be an issue? It may need changes in both topgen and reggen?

util/topgen/clocks.py
matutem: Perhaps this could be done in the constructor (__init__) of class `Group` ? It currently simply initializes self.clocks to an empty map, but if it finds an existing map for self.clocks it could just initialize from it (see clocks/py, line 116).

util/gen_top_docs.py
```diff
@@ -0,0 +1,278 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
```
matutem: copyright zeroRISC for any new files.

util/gen_top_docs.py
```diff
@@ -0,0 +1,278 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+r"""Top Module Documentation Generator
```
matutem, line 5: Could you expand this file docstring? Remember it can be displayed in the usage or help message. It would  be useful to mention which files are generated, and what they contain.

util/gen_top_ipconfigs.py
```diff
@@ -0,0 +1,620 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Generate ipgen IP configuration files from a complete top configuration.
```
matutem, line 6: Also please expand this docstring to explain what this generates.

util/gen_top_ipconfigs.py
```diff
@@ -0,0 +1,620 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+// ------------------- W A R N I N G: A U T O - G E N E R A T E D   C O D E !! -------------------//
+// PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED WITH THE FOLLOWING COMMAND:
+// util/top_ipconfigs.py {} -o {}
+
+"""
+
+
+def _get_alert_handler_params(top: ConfigT, name: str) -> ParamsT:
```
matutem: Is this a duplicate of what topgen.py has? If so, this code may need a better location, like in some file under ipgen?

util/gen_top_ral.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Generate top RAL model for DV from a complete top configuration."""
```
matutem: Expand this as explained above: generated files, etc. Anything you would like from the help message (other than the args) of a tool you'd like to use 🙂

util/topgen/c_test.py
vlpannel: I think this is a pretty pervasive issue that would introduce too many changes for this PR.

Filed issue #336.

hw/top_dragonfly/data/autogen/top_dragonfly.gen.hjson
```diff
@@ -18382,7 +18382,7 @@
       type: interrupt
       module_name: uart0
       desc: uart0 rx_watermark interrupt
-      intr_type: IntrType.Status
```
matutem, line 18385: It seems unfortunate we lost mentioning Event and Status. Is there a way to associate a symbolic name to 1 and 2?

hw/top_dragonfly/rtl/autogen/testing/top_dragonfly_rnd_cnst_pkg.vbl
```diff
@@ -5,11 +5,11 @@
 # ------------------- W A R N I N G: A U T O - G E N E R A T E D   C O D E !! -------------------#
 # PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED WITH THE FOLLOWING COMMAND:
 #
-# util/topgen.py -t hw/top_dragonfly/data/top_dragonfly.hjson
-#                -o hw/top_dragonfly/
+# util/gen_top_sv.py --completecfg hw/top_dragonfly/data/autogen/top_dragonfly.gen.hjson
+#                    --seedcfg hw/top_dragonfly/data/autogen/top_dragonfly.secrets.testing.gen.hjson
```
matutem, line 9: By the way, how is the secrets gen hjson file generated? I don't see it as modified.

util/reggen/access.py
```diff
@@ -6,13 +6,14 @@
 """
 
 from enum import Enum
+from typing import Any
```
matutem: I imagine this is to handle the pesky `None`s?

hw/top_egret/ip_autogen/flash_ctrl/rtl/flash_ctrl_mem_reg_top.sv
```diff
@@ -0,0 +1,42 @@
+// Copyright lowRISC contributors (OpenTitan project).
```
matutem: Where this this come from? It seems to be a simple encapsulation for the integrity generator, which I assume was embedded? Is it actually used?

hw/Makefile
```diff
@@ -70,14 +70,29 @@ ips_reg        = $(addsuffix _reg, $(local_ips))
...
 top_hjson = $(PRJ_DIR)/hw/$*/$(dir_hjson)/$*.hjson
+top_complete_hjson = $(PRJ_DIR)/hw/$*/$(dir_hjson)/autogen/$*.gen.hjson
 top_seed_hjson = $(PRJ_DIR)/hw/$*/$(dir_hjson)/$*_seed.$(SEED_MODE).hjson
+top_complete_seed_hjson = $(PRJ_DIR)/hw/$*/$(dir_hjson)/autogen/$*.secrets.$(SEED_MODE).gen.hjson
 top_chip_hjson = $(PRJ_DIR)/hw/$(TOPS)/$(dir_hjson)/$(TOPS).hjson
 
 top: $(tops_gen)
 $(tops_gen): %_gen:
 	${PRJ_DIR}/util/topgen.py -t $(top_hjson) -s $(top_seed_hjson) -o ${PRJ_DIR}/hw/$*/ ${toolflags}
+	(for ipgen in $$(${PRJ_DIR}/util/top_cm_and_blocks.py -t $(top_complete_hjson) --filter ipgen get_blocks); do \
```
matutem, line 84: Could you create an issue for the Makefile to be dependency driven? This means creating more targets with proper dependencies. Most of the new stuff depends on the completed config, and on the generators.

The issue could also propose using bazel for this, on the assumption that it can be more sophisticated about dependencies for python scripts, which drive most of the generation.

util/reggen/gen_rtl.py
```diff
@@ -120,12 +121,22 @@ def gen_rtl(block: IpBlock, outdir: str) -> int:
     # (writing to <block>_reg_top.sv). In any other case, we also need the
     # interface name, giving <block>_<ifname>_reg_top.
     lblock = block.name.lower()
-    for if_name, rb in block.reg_blocks.items():
+    for if_name, rb in list(block.reg_blocks.items()) + list(block.memories.items()):
```
matutem: I see, this is how the flash_ctrl mem top was generated. Do we actually need this now? I assume it could be a separate PR that goes together with instantiating those blocks: no point of generating them if they are unused.

util/tlgen/generate.py
```diff
@@ -28,12 +28,18 @@ def generate(xbar: Xbar, library_name: str = "ip") -> List[Tuple[str, Any]]:
...
+        filename=str(importlib.resources.files('tlgen') / 'xbar.BUILD.tpl'))
+    xbar_databuild_tpl = Template(
+        filename=str(importlib.resources.files('tlgen') / 'xbar.data_BUILD.tpl'))
     try:
         out_rtl = xbar_rtl_tpl.render(xbar=xbar)
         out_pkg = xbar_pkg_tpl.render(xbar=xbar)
         out_core = xbar_core_tpl.render(xbar=xbar,
                                         library_name=library_name)
         out_hjson = xbar_hjson_tpl.render(xbar=xbar)
+        out_build = xbar_build_tpl.render(xbar=xbar, outpath=outpath)
```
matutem, line 42: How come these use `outpath` but not the other `render` calls above?

util/tlgen/xbar.BUILD.tpl
```diff
@@ -0,0 +1,19 @@
+# Copyright lowRISC contributors (OpenTitan project).
```
matutem, line 1: copyright to zeroRISC

util/tlgen/xbar.data_BUILD.tpl
```diff
@@ -0,0 +1,10 @@
+# Copyright lowRISC contributors (OpenTitan project).
```
matutem, line 1: copyright to zeroRISC

util/tlgen/xbar.sim_cfg.hjson.tpl
```diff
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
```
matutem, line 3: Add zeroRISC copyright. In fact, make sure any new file is zeroRISC, and any modified adds zeroRISC, if the modification are important or are due to dependencies that are also important.

util/tlgen/xbar.sim_cfg.hjson.tpl
```diff
@@ -2,7 +2,9 @@
...
 // SPDX-License-Identifier: Apache-2.0
 //
-// xbar_${xbar.name}_sim_cfg.hjson file generated by `tlgen.py` tool
+// xbar_${xbar.name}_sim_cfg.hjson file generated by `tlgen.py` tool<%
```
matutem: What is the `<%` at the EOL?

util/topgen/c_test.py
```diff
@@ -272,8 +273,8 @@ def _get_alert_peripherals(self, alert_handler, addr_space):
                     if region is not None:
                         break
             if region is None:
-                log.error(f"Could not find supported reg interface for alerts "
-                          f"coming from module {name}")
+                log.error(f'Could not find supported reg interface for alerts '
```
matutem: The first string has no argument interpolation, so it doesn't need `f`.

util/gen_top_docs.py
matutem: copyright zeroRISC

util/gen_top_docs.py
```diff
@@ -0,0 +1,278 @@
...
+def gen_memory_map_docs(top, c_helper, out_path):
+    """Generate memory map documentation for all address spaces.
+
+    Creates a markdown file showing the memory layout for all address spaces
+    defined in the top configuration.
+    """
+    doc_path = out_path / 'doc'
+    doc_path.mkdir(parents=True, exist_ok=True)
+
+    gencmd = ("util/topgen.py -t hw/top_{topname}/data/top_{topname}.hjson "
```
matutem: use f-strings

util/gen_top_docs.py
```diff
@@ -0,0 +1,278 @@
...
+    template = Template(filename=str(template_path))
+
+    memory_map_content = template.render(
+        top=top,
+        helper=c_helper,
+        gencmd=gencmd
+    )
+
+    memory_map_path = doc_path / "memory_map.md"
+    memory_map_path.write_text(memory_map_content)
```
matutem: Add try-except since a file write can fail.

util/gen_top_docs.py
```diff
@@ -0,0 +1,278 @@
...
+
+    name_to_block = {}
+    for module in completecfg["module"]:
+        block_type = module["type"]
+        ip_hjson = get_ip_hjson_path(block_type, completecfg, REPO_TOP)
+        name_to_block[block_type] = IpBlock.from_path(str(ip_hjson), [])
+    c_helper = TopGenCTest(completecfg, name_to_block)
+    outdir = args.outdir or args.completecfg.parents[2].resolve()
+
+    gen_top_docs(completecfg, c_helper, outdir)
```
matutem, line 297: Add the standard

```python
if __name__ == "__main__":
    main()
```
at the end of all new scripts.

util/tlgen.py
```diff
@@ -50,9 +58,9 @@ def main():
         sys.stdout.write(tlgen.selfdoc(heading=3, cmd='tlgen.py --doc'))
         return
 
-    # Check if topcfg defined
-    if not args.topcfg:
-        log.error("--topcfg option is mandatory to generate codes.")
+    # Check if tl_xbar_cfg defined
```
matutem: Better if you add "required=True" around line 26, and remove this check.

util/gen_top_ipconfigs.py
```diff
@@ -0,0 +1,620 @@
...
+    seedcfg = load_cfg(args.seedcfg) if args.seedcfg is not None else {}
+
+    # can use either seed cfg or generated secrets cfg
+    seed_value = (seedcfg.get("topgen_seed")
+                  or seedcfg.get("seed", {}).get("topgen_seed", {}).get("value"))
+    if seed_value is None:
+        log.warning("no top level seed value found (may be required by some IPs)")
+
+    extract_clocks(completecfg)
+    amend_resets(completecfg, {}, allow_missing_blocks=True)  # name_to_block not needed
```
matutem: This call should not be necessary: the complete config should have the definitive list of resets. You can verify by comparing the list of resets before and after this call: they must be equal, indicating the call should be removed. If they are not the same we have a bug in the complete config generation.

util/gen_top_ipconfigs.py
```diff
@@ -0,0 +1,620 @@
...
+
+    ipgen_modules = [module for module in completecfg["module"]
+                     if is_ipgen(module) and ipgen_is_selected(module, args.ips)]
+
+    for module in ipgen_modules:
+        log.info(f"generating ipconfig for {module['name']}")
+        if seed_value is not None:
+            SecurePrngFactory.create("topgen", seed_value)  # reset random seed every time
+
+        params = get_params(completecfg, module, args.completecfg.parents[2])
```
matutem: The last arg is the path to hw/top_*, right? Maybe create a well-named variable assigned to the current last arg and use that instead.

matutem (commented):
A few more comments, suggestions, and things to fix. I think this is getting near the end-line though.

matutem:
How do you verify this works as intended? I assume you can compare the generated files and filter the obvious changes. Can you elaborate, perhaps in the topgen README?

Also how do you verify the various scripts generate all files that the old topgen generated? This may be trickier to verify.

And lastly, could you open an issue to turn the file docu-string of the new scripts into a more complete description of what they do, and add that docustring to the usage message so the script's --help command will be more informative. Perhaps you already opened such an issue...

util/gen_top_docs.py
vlpannel: As far as I can tell, almost all of this code is simply cut and pasted from the original topgen script (which has this copyright), but I can update to include both copyright headers.

util/tlgen/xbar.sim_cfg.hjson.tpl
vlpannel: This is the start of the Python block. I probably should have just placed it on the next line; it causes the resulting file to have an extra newline, but that's fine.

hw/top_dragonfly/data/autogen/top_dragonfly.gen.hjson
vlpannel: > Update: it's because of the function for_json in util/reggen/access.py, where the IntrType enum gets cast as a string when put into the Hjson.

[Earlier](https://github.com/pavona/pavona/pull/279#discussion_r3493578107), we were discussing how the `intr_type` being read back into Python from Hjson was requiring us to search for strings.

The code can either store a string in the Hjson and read it back out as a string or store the enum's int value in Hjson and read it back out directly to enum. Either one is fine, and I am happy to restore it to the string version if that's better.

hw/top_dragonfly/rtl/autogen/testing/top_dragonfly_rnd_cnst_pkg.vbl
vlpannel: The secrets file is generated at the same time as the complete config. This is because a lot of the elaboration done in the process for the complete top config is also useful for generating the secrets file. Can break that file out on its own too, but this currently seems redundant to me.

util/reggen/access.py
vlpannel: This is actually a mistake on my part (thank you for the catch!). I will delete this line and instead more strictly type the `for_json` function on line 15 (should be outputting either a string or int based on which version of the function we choose; see [this](https://github.com/pavona/pavona/pull/279#discussion_r3683391719) previous comment).

util/tlgen/xbar.sim_cfg.hjson.tpl
vlpannel: Thanks for the clarification; while I didn't consider the modifications all that important, I did not realize that changes in other code/dependencies also warrants this.

hw/top_egret/ip_autogen/flash_ctrl/rtl/flash_ctrl_mem_reg_top.sv
vlpannel: Conversation continues [below](https://github.com/pavona/pavona/pull/279#discussion_r3677674384).

util/reggen/gen_rtl.py
vlpannel: I can remove this for now and put this in later! It makes sense not to generate unused things.

For context, I saw that `rv_dm` has a mem_reg_top file and assumed that memories may want to be partially instantiated as well.

hw/Makefile
vlpannel: I had the same thought as I was editing the file. Now have made issue #343. If we are switching to Bazel, then it would not be worth overhauling this Makefile, and instead a new one should be built from scratch.

util/tlgen/generate.py
vlpannel: The original tlgen code has the other outpaths hard coded. Not terribly relevant to this PR, so I created an issue for this: #344.

vlpannel:
Thank you for the review! I'll make the fixes suggested.

For the questions about verifying this code: yes, it was mostly just the check for "did autogenerated code change". I will check on the list of files which are affected by the new makefile and compare that with what is changed on `main`.

As for the docstrings: apologies on my part! I have a local commit fixing them that I must've forgotten to push. Should be pushed now. Will not file an issue for now unless there is a good reason to.

util/reggen/gen_rtl.py
vlpannel: That and `soc_proxy_ctn_reg_top.sv`.

vlpannel:
@matutem I've made all requested changes; PTAL. A few notes:
* I checked to see all the files the original topgen updated, and edited the code so that the new Makefile generates every file that the original one did except for `hw/top_scafi_deprecated/data/autogen outgoing_alerts_scafi_deprecated.hjson` (but we do not really care about this file).
* I decided to ultimately use the string value of the `IntrType` in the Hjson--after all, Hjson is about readability, and it's not too much of a hassle to check for a string instead of an int in the Python code.
* The secrets Hjson is still generated alongside the complete top config; let me know if this should change.
* I'll be fixing up most of the new change commits before merge.

Note: the lint failure is currently due to bitstream cache issues, which seems to be happening on `main` as well.

matutem (approved):
This was a big one! Thanks for your patience and responsiveness.
LGTM

PR #280, opened 2026-06-25
[acc,mlkem] Test ACC ML-KEM on a PQCEn=1 CW340 bitstream in CI

mkannwischer:
- Depends on #269 

Builds a PQC-enabled CW340 bitstream (`chip_egret_cw340_pqc`, synthesized with `AccPQCEn=1`) and runs the ACC ML-KEM KAT against it in CI, so the ACC's PQC datapath is exercised on hardware.

- `AccPQCEn` is forwarded through the egret chip wrapper as a chip-level parameter (default 0). The PQC bitstream is a separate cached design.
- It's built and cached like the others, so it's reusable locally without a local Vivado build.
- `mlkem_kat` opts into the `fpga_cw340_pqc_test_rom` exec_env and runs with `--define=acc_has_pqc=true` in the PR and nightly CW340 jobs, selected by the `cw340_pqc_test_rom` tag.

Test on CW340:
```
./bazelisk.sh test \
  //sw/device/tests/crypto/cryptotest:mlkem_kat_fpga_cw340_pqc_test_rom \
  --define=acc_has_pqc=true --define=bitstream=vivado \
  --test_arg=--usb-serial=<serial> --test_output=streamed
```

### DIFF
--- a/.github/workflows/egret-cw340-nightly.yml
+++ b/.github/workflows/egret-cw340-nightly.yml
@@ -96,6 +96,21 @@ jobs:
       timeout_filters: 'long,eternal'
       timeout: 300
 
+  execute_long_pqc_fpga_tests_cw340:
+    name: CW340 PQCEn=1 Long Tests
+    uses: ./.github/workflows/fpga.yml
+    secrets: inherit
+    with:
+      name: CW340 PQCEn=1 Long Tests
+      job_name: execute_long_pqc_fpga_tests_cw340
+      bitstream: chip_egret_cw340_pqc
+      board: cw340
+      interface: cw340
+      tag_filters: cw340_pqc_test_rom
+      timeout_filters: 'long,eternal'
+      timeout: 300
+      extra_bazel_flags: --define=acc_has_pqc=true
+
   verify_fpga_jobs:
     name: Verify FPGA jobs
     uses: ./.github/workflows/verify-fpga.yml
@@ -110,4 +125,5 @@ jobs:
       - execute_long_sival_fpga_tests_cw340
       - execute_long_sival_rom_ext_fpga_tests_cw340
       - execute_long_manuf_fpga_tests_cw340
+      - execute_long_pqc_fpga_tests_cw340
     if: success() || failure()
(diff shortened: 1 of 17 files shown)


.github/workflows/egret-cw340-nightly.yml
```diff
@@ -96,6 +96,21 @@ jobs:
...
+    with:
+      name: CW340 PQCEn=1 Long Tests
+      job_name: execute_long_pqc_fpga_tests_cw340
+      bitstream: chip_egret_cw340_pqc
+      board: cw340
+      interface: cw340
+      tag_filters: cw340_pqc_test_rom
+      timeout_filters: 'long,eternal'
+      timeout: 300
+      extra_bazel_flags: --define=acc_has_pqc=true
```
pqcfox, line 112: not necessarily for this PR, but longer term, is there be a way to define `acc_has_pqc=true` internally set when using the cw340_pqc execution environments?

pqcfox (approved):

.github/workflows/egret-cw340-nightly.yml
mkannwischer: I could not find a nice way so far. Could you give it a try?
Would be good if one could still overwrite it because of course you can also run the ibex implementations on the cw340_pqc.

sw/device/tests/crypto/cryptotest/BUILD
```diff
@@ -90,6 +90,9 @@ MLKEM_TESTVECTOR_ARGS = " ".join([
 
 cryptotest(
     name = "mlkem_kat",
+    extra_exec_envs = {
+        "//hw/top_egret:fpga_cw340_pqc_test_rom": "fpga_cw340",
```
rtorok-zr, line 94: Would it be simpler to just add `fpga_cw340_pqc_test_rom` to `CRYPTOTEST_EXEC_ENVS`? There shouldn't be any harm to adding that as an extra exec_env on the other cryptotests, even non-PQ ones.

rtorok-zr (approved):
LGTM, let's get the CW340 PQC bitstream build + tests to pass before merging.

sw/device/tests/crypto/cryptotest/BUILD
mkannwischer: Then ALL tests are run on that env eating more CW340 time and doesn't add much value over the other CW340 tests.

mkannwischer:
> LGTM, let's get the CW340 PQC bitstream build + tests to pass before merging.

Thanks! It had passed before I rebased. Not sure what happened there. I'll restart it once the other job is done.

PR #284, opened 2026-06-29
[acc,mldsa] Add ACC-backed ML-DSA to the cryptolib

mkannwischer:
- Depends on #269 
- Depends on #247 
- Depends on #258

This PR wires up the ML-DSA ACC implementation into the cryptolib in the same way that #269 did for ML-KEM. This confirms that the ACC implementation passes ACVP and Wycheproof testvectors.
The `acc_has_pqc` configuration option is used to select between the mldsa-native and the ACC code path.
Below are instructions to test both paths in Verilator and on the CW340:

mldsa-native:
```
# verilated functional test 
./bazelisk.sh test //sw/device/tests/crypto:mldsa_functest_sim_verilator --test_output=streamed

# ACVP + Wycheproof tests on the CW340 (pre-built bitstream)
./bazelisk.sh test //sw/device/tests/crypto/cryptotest:mldsa_kat_fpga_cw340_test_rom --test_arg=--usb-serial=<usb-serial> --test_output=streamed
```

ACC:
```
# set AccPQCEn=1
git fetch https://github.com/mkannwischer/pavona acc-pqcen
git cherry-pick 4ac8abbfd696408e9c16c6def644335257d3ff38

# verilated functional test (with acc_has_pqc=true)
./bazelisk.sh test //sw/device/tests/crypto:mldsa_functest_sim_verilator --test_output=streamed  --define=acc_has_pqc=true

# ACVP + Wycheproof tests on the CW340 (with acc_has_pqc=true)
./bazelisk.sh test //sw/device/tests/crypto/cryptotest:mldsa_kat_fpga_cw340_test_rom  --test_arg=--usb-serial=2055355B3236 --test_output=streamed --define bitstream=vivado --define=acc_has_pqc=true
```

`<usb-serial>` should be 2055355B3236 in case you are running this on the shared workstation.

PR #258, opened 2026-06-16
[acc,mldsa] Small bug fixes and preparation for cryptolib integration

mkannwischer:
- Depends on #247 

In preparation of integration the ML-DSA ACC code into the cryptolib, I had to make a few adjustments to the ML-DSA ACC assembly. It also required a change to the flow script.
The actual cryptolib integration will follow in another pull request.

PR #247, opened 2026-06-12
[acc,mldsa] Turn ML-DSA parameter set into a runtime parameter

mkannwischer:
In preparation of wiring the ML-DSA ACC implementation up into the cryptolib, this PR changes the implementation so that the parameter set is a runtime argument. This allows building a single binary supporting all parameter sets and operation which enabled code sharing between them.

PR #285, opened 2026-06-29
[acc,mldsa] Test ACC ML-DSA on a PQCEn=1 CW340 bitstream in CI 

mkannwischer:
- Depends on #269 
- Depends on #247 
- Depends on #258
- Depends on #280 
- Depends on #284 

Doing this in a separate PR to not introduce the #280 dependency into #284.


pqcfox (approved):
Fantastic, excited this is in!

PR #286, opened 2026-06-29
[ci] Reduce concurrent jobs

LouisTheLuis:
In order to support more internal jobs, we reduce the number of licenses used by the CI on weeklies.


qmn (approved):
LGTM, thanks!

PR #287, opened 2026-06-29
[doc] Reduce default verilator threads to one

### LINKED ISSUE #283: [doc] Getting started guide doesn't work on machines with <4 cores
The [Getting Started](https://docs.pavona.org/book/doc/getting_started/) guide on the website instructs the user to run the hello world verilator test with this command:
```
$ ./bazelisk.sh test sw/device/examples/hello_world:hello_world_sim_verilator --test_output=streamed
```
However, our Verilator build options are configured to build with four cores:

https://github.com/pavona/pavona/blob/3f893f49e8f2170b8472d44fed4e591e1c9e85a2/hw/BUILD#L35-L40

As a result, if the command is run on a machine with less than four cores, it produces an error that looks like this:
```
VerilatedContext has 2 threads but model 'Vchip_sim_tb' (instantiated as 'TOP') was Verilated with --threads 4."
```
This can be worked around by adding `--//hw:verilator_options=--threads,$(nproc)` to the bazelisk command line; we should probably add this to the getting started guide.

fragglet:
Our build hard-codes the compiled verilator binaries to use four cores. This is fine on most modern machines, but if you're building on eg. a virtual machine without that many cores, it won't work.

Rather than complicate our documentation, this sets the default to only assume a single core, which will work on any machine. However, we include a note in the getting started guide for how to improve performance. This also updates the Github Actions configuration to make full use of all available cores, so that we don't make the CI runs slower.

Fixes #283.


qmn:
Hi Simon, thanks so much for the fix! While it's certainly wrong to just put `--threads 4`, is there a way we can fix it from `hw/BUILD` to use `$(nproc)` by default? That would not only save the extra typing on Getting Started but also fix it everywhere else.

Edit: I've done some digging and I haven't found any clean way of passing the number of CPUs to through to Bazel. This makes sense. In any case, it's not always the right move to set `--threads` to the number of processors, as Verilator performs a coarse-grain partitioning of the design across threads, and you get diminishing returns as the number of threads grows. [`--threads 1`](https://veripool.org/guide/latest/exe_verilator.html#cmdoption-threads) is perhaps the safest option, despite the performance hit. Ideally, we'd pass the minimum of, say, 4, and `$(nproc)` to `--threads`.

fragglet:
> Hi Simon, thanks so much for the fix! While it's certainly wrong to just put `--threads 4`, is there a way we can fix it from `hw/BUILD` to use `$(nproc)` by default? 

I considered this but I think it's not a good idea because it would potentially affect reproducibility of builds - ie. the compiled verilator binary really shouldn't depend on the number of cores in the host machine that's doing the compilation.

> `--threads 1` despite the performance hit.

Yes, fair enough I suppose, though the performance hit is unfortunate, it's going to be the safer option.

fragglet:
Now updated to use `--threads 1` as a safe default, though I added a brief note that it can be removed for better performance.

qmn:
Thanks so much (and pardon the delay)! I think it's best to keep the instructions as concise as possible so that we expose new users to just the right amount of new information at a time. Would it be possible to make `--threads 1` the default globally, but add a note to `doc/getting_started/setup_verilator.md` for better performance?

fragglet:
I have reworked this as suggested - this PR now changes the default number of threads to 1, but includes a note in the getting started guide about how to configure Bazel/Verilator to use more threads.

doc/getting_started/README.md
```diff
@@ -168,6 +168,14 @@ INFO: Build completed successfully, 49 total actions
 The output (with some trimming) shows the files that were built.
 You can examine the RISC-V (dis)assembly in the `bazel-bin/sw/device/examples/hello_world/hello_world_sim_verilator.dis` file.
 
+Note: by default, verilator only uses a single core for simulation.
+Modern machines have many cores, and you can make use of them by using the `--//hw:verilator_options` flag.
+Create a file named `.bazelrc-site` containing a line like the following:
+```
+common --//hw:verilator_options=--threads,8
+```
+The above example would appropriate be for a machine with 8 cores (you can determine the number of cores in your CPU by running the `nproc` command).
```
mkannwischer:
```suggestion
The above example would be appropriate for a machine with 8 cores (you can determine the number of cores in your CPU by running the `nproc` command).
```

hw/BUILD
mkannwischer: Outdated comment.

rtorok-zr (approved):

mkannwischer (approved):
Thanks @fragglet!

PR #230, opened 2026-06-04
[acc,masking] Add masked ML-KEM decaps and keygen, plus masking gadgets

pqcfox:
This PR adds a fully-masked ML-KEM implementation for Pavona's Asymmetric Cryptography Coprocessor, as written by Hoang Nguyen Hien Pham (@phamhnh).

This implementation of ML-KEM decapsulation and keypair generation was designed to support arbitrary masking order. For first-order masking, it allows direct use with the masked KMAC hardware already present in Pavona.

A full explanation of the effort, including an overview of what masking is, why side-channel hardening is important, and how this implementation is designed is available in Hien's blog post here: https://www.zerorisc.com/blog/hardened-pqc-on-pavona-masking-ml-kem.

This draft PR includes:

- Functional implementations of various masking gadgets evaluated as part of the masking effort (bitsliced and non-bitsliced)
- A fully-masked ML-KEM decapsulation implementation
- A fully-masked ML-KEM keypair generation implementation

The implementation in this work is ready for evaluation (see the draft PR for details) but there is still some further work we'd like to do.

In particular, we're working to:

- Remove the evaluated masking gadgets (see the blog post for what these are) not used in this branch, retaining them in an archived branch for academic review
- Coalesce the remaining gadgets into a single file
- Documentation each remaining gadget
- Update the clobbered register lists for each function changed
- Evaluate and possibly tweak our register whitening approach (see the blog post for an explanation of this as well) throughout the implementation based on side-channel testing
- Size down stacks for each ACC program to what is actually needed
- Lightly clean up the test code added as part of this PR

We would like to invite feedback from the community while we work on incorporating these changes, and expect the implementation to be ready for full review in the coming weeks.

PR #288, opened 2026-06-30
[acc] Extend information-flow tools to track DMEM

mkannwischer:
This PR ports https://github.com/zerorisc/expo/pull/247. 

It also makes a few further adjustments to enable running it on the ML-KEM masking gadgets from #230.

### DIFF
--- a/hw/ip/acc/data/base-insns.yml
+++ b/hw/ip/acc/data/base-insns.yml
@@ -289,7 +289,7 @@
     bytes: 4
   iflow:
     - to: [grd]
-      from: [dmem]
+      from: [dmem-grs1-4]
 
 - mnemonic: sw
   rv32i: true
@@ -320,7 +320,7 @@
     target: [offset, grs1]
     bytes: 4
   iflow:
-    - to: [dmem]
+    - to: [dmem-grs1-4]
       from: [grs2]
 
 - mnemonic: beq
@@ -444,41 +444,13 @@
     target: [csr]
   iflow:
     - to: [grd]
+      from: [csr]
+    - to: [csr]
+      from: [csr,grs1]
+    - test:
+        - csr == 0x7d9
+      to: [kmac_core]
       from: []
-    - test:
-        - csr == 0x7c0
-      to: [fg0-all]
-      from: [fg0-all, grs1]
-    - test:
-        - csr == 0x7c0
-      to: [grd]
-      from: [fg0-all]
-    - test:
-        - csr == 0x7c1
-      to: [fg1-all]
-      from: [fg1-all, grs1]
-    - test:
-        - csr == 0x7c1
-      to: [grd]
-      from: [fg1-all]
-    - test:
-        - csr == 0x7c8
-      to: [fg0-all, fg1-all]
-      from: [fg0-all, fg1-all, grs1]
-    - test:
-        - csr == 0x7c8
-      to: [grd]
-      from: [fg0-all, fg1-all]
-    - test:
-        - csr >= 0x7d0
-        - csr <= 0x7d8
-      to: [mod]
-      from: [mod, grs1]
-    - test:
-        - csr >= 0x7d0
-        - csr <= 0x7d8
-      to: [grd]
-      from: [mod]
 
 - mnemonic: csrrw
   rv32i: true
@@ -509,49 +481,13 @@
     target: [csr]
   iflow:
     - to: [grd]
+      from: [csr]
+    - to: [csr]
+      from: [grs1]
+    - test:
+        - csr == 0x7d9
+      to: [kmac_core]
       from: []
-    - test:
-        - grd != 0
-        - csr == 0x7c0
-      to: [fg0-all]
-      from: [grs1]
-    - test:
-        - grd != 0
-        - csr == 0x7c0
-      to: [grd]
-      from: [fg0-all]
-    - test:
-        - grd != 0
-        - csr == 0x7c1
-      to: [fg1-all]
-      from: [grs1]
-    - test:
-        - grd != 0
-        - csr == 0x7c1
-      to: [grd]
-      from: [fg1-all]
-    - test:
-        - grd != 0
-        - csr == 0x7c8
-      to: [fg0-all, fg1-all]
-      from: [grs1]
-    - test:
-        - grd != 0
-        - csr == 0x7c8
-      to: [grd]
-      from: [fg0-all, fg1-all]
-    - test:
-        - grd != 0
-        - csr >= 0x7d0
-        - csr <= 0x7d8
-      to: [mod]
-      from: [mod, grs1]
-    - test:
-        - grd != 0
-        - csr >= 0x7d0
-        - csr <= 0x7d8
-      to: [grd]
-      from: [mod]
 
 - mnemonic: ecall
   rv32i: true
--- a/hw/ip/acc/data/bignum-insns.yml
+++ b/hw/ip/acc/data/bignum-insns.yml
@@ -1328,6 +1328,10 @@
         - wsr == 0x9
       to: [wsr]
       from: [wsr]
+    - test:
+        - wsr == 0xc
+      to: [wsr]
+      from: [wsr]
   encoding:
     scheme: wcsr
     mapping:
--- a/hw/ip/acc/data/bignum-insns.yml
+++ b/hw/ip/acc/data/bignum-insns.yml
@@ -1134,7 +1134,7 @@
     - *data-addr
   iflow:
     - to: [wrd]
-      from: [dmem]
+      from: [dmem-grs-32]
     - test:
         - grs_inc == 1
       to: [grs]
@@ -1191,7 +1191,7 @@
     target: [offset, grs]
     bytes: 32
   iflow:
-    - to: [dmem]
+    - to: [dmem-grs-32]
       from: [wrs]
     - test:
         - grs_inc == 1
--- a/hw/ip/acc/data/bignum-insns.yml
+++ b/hw/ip/acc/data/bignum-insns.yml
@@ -996,7 +996,7 @@
     - &data-addr A `BAD_DATA_ADDR` error if the computed address is not a valid DMEM address aligned to WLEN bits.
   iflow:
     - to: [wref-grd]
-      from: [dmem]
+      from: [dmem-grs1-32]
     - test:
         - grd_inc == 1
       to: [grd]
@@ -1069,7 +1069,7 @@
     target: [offset, grs1]
     bytes: 32
   iflow:
-    - to: [dmem]
+    - to: [dmem-grs1-32]
       from: [wref-grs2]
     - test:
         - grs1_inc == 1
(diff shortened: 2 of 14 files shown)


hw/ip/acc/data/base-insns.yml
```diff
@@ -444,41 +444,13 @@
     target: [csr]
   iflow:
     - to: [grd]
-      from: []
-    - test:
-        - csr == 0x7c0
-      to: [fg0-all]
```
pqcfox, line 450: Clarification question: how come the iflow rules for the flags and mod CSRs were removed? (Is this handled by the SPECIAL_REGS mechanism?)

hw/ip/acc/data/bignum-insns.yml
```diff
@@ -1293,15 +1293,7 @@
     - &bad-wsr An `ILLEGAL_INSN` error if `wsr` doesn't name a valid WSR.
   iflow:
     - to: [wrd]
-      from: []
```
pqcfox, line 1296: (ditto to the clarification question for the WSR logic here)

hw/ip/acc/data/bignum-insns.yml
```diff
@@ -1326,16 +1318,16 @@
...
       from: [wrs]
     - test:
-        - wsr == 0x3
-      to: [acc]
-      from: [wrs]
+        - wsr == 0x8
+      to: [kmac_core]
+      from: []
+    - test:
+        - wsr == 0x9
```
pqcfox, line 1328: Clarification question: it looks like this corresponds to `KMAC_DATA_S1`, and 0x8 prior corresponds to `KMAC_DATA_S0`; shouldn't these both be to `kmac_core`?

pqcfox:
Have started on a review for this, which I should have completed tomorrow. Thanks!

pqcfox:
(Apologies, will need a bit more time--thanks!)

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -72,56 +221,87 @@ def nonexistent() -> 'InformationFlowGraph':
...
+            out |= self.flow[s]
+        return out
 
-    def sinks(self, source: str) -> Set[str]:
+    def sinks(self, source: str) -> Set[InformationFlowNode]:
         '''Returns all sinks for the given source.'''
         out = set()
         for sink in self.flow:
-            if source in self.flow[sink]:
+            if [s for s in self.flow[sink] if s.name == source]:
```
pqcfox: [nit] maybe more idiomatic could be
```suggestion
            if any([s.name == source for s in self.flow[sink]]):
```

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -72,56 +221,87 @@ def nonexistent() -> 'InformationFlowGraph':
...
                 out.add(sink)
-        if source not in self.flow:
+        if [s for s in self.flow if s.name == source]:
```
pqcfox: [nit] possibly simliarly
```suggestion
        if any([s.name == source for s in self.flow]):
```

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -72,56 +221,87 @@ def nonexistent() -> 'InformationFlowGraph':
...
 
         If the node is not a sink in the graph, does nothing.
         '''
-        self.flow.pop(node, None)
+        sink_nodes = [n for n in self.flow if n.name == node]
+        assert len(sink_nodes) <= 1
+        if sink_nodes:
+            self.flow.pop(sink_nodes[0], None)
+
+    def simplify(self) -> None:
```
pqcfox, line 288: This seems at first glance like it might have results dependent on the iteration order of `self.flow.keys()`; in particular, if we have (in iteration order) 

`sources = [<DMEM range 0x0:0x10>, <DMEM range 0x20:0x30>, <DMEM range 0x10:0x20>]`

then it seems like 

* `sources[0]` would _not_ merge with `sources[1]`, since not adjacent
* `sources[0]` would merge with `sources[2]`
* `sources[2]` would be deleted

and then `sources[1]` is the last source in the list, so we're left with `sources[0].union(sources[2])` and `sources[1]` even though all three should be coalesced together.

Am I possibly missing something I haven't seen this behavior at all when working from this commit before, so I'm thinking I may be misreading what's happening here and that it works as is.

hw/ip/acc/util/shared/information_flow.py
pqcfox: [nit] (possibly ditto in `combine()` as well)

pqcfox (commented):
Got through almost everything; will cover `information_flow_analysis.py` first thing tomorrow, here's the review so far.

hw/ip/acc/util/shared/information_flow_analysis.py
```diff
@@ -288,9 +293,28 @@ def _get_iflow_update_state(
     return iflow.seq(rec_return_iflow)
 
 
+def simplify_control_deps(
```
pqcfox, line 306: This appears like it might run into the same issue as `simplify()` prior, where the order of `control_deps.keys()` may affect the result.

hw/ip/acc/util/shared/information_flow_analysis.py
```diff
@@ -525,14 +555,14 @@ def _get_iflow(program: ACCProgram, graph: ControlGraph, start_pc: int,
         # the cycle)
         stable_constants = ConstantContext.empty()
         for k, v in start_constants.values.items():
-            if k not in cycle_iflow.all_sinks():
+            if not any([s for s in cycle_iflow.all_sinks() if s.name == k]):
```
pqcfox: [nit] for possible clarity
```suggestion
            if not any([s.name == k for s in cycle_iflow.all_sinks()]):
```

hw/ip/acc/util/shared/information_flow_analysis.py
```diff
@@ -676,34 +725,61 @@ def stringify_control_deps(program: ACCProgram,
...
+    symbols = get_dmem_symbols(program)
+    # if any overlapping nodes affect control flow at the same PCs, unify them
+    dedup_control_deps = {}
     for node, pcs in control_deps.items():
+        if any([n.overlaps(node) and pcs == dedup_control_deps[n] for n in dedup_control_deps]):
+            # there is already an overlapping node with the same PCs, meaning
+            # we've already processed this node
+            continue
+        for other in control_deps:
+            if pcs == control_deps[other] and node.overlaps(other):
```
pqcfox, line 755: I think this may (similar to `simplify()`) assume an ordering on how `control_deps` is iterated through.

pqcfox:
This looks excellent (and fantastic catch on the non-termination issue with cycle traversal). 

Many of the comments above are just nits and small clarification questions; I think the primary details of interest are the notes re: iteration order in functions like `simplify_control_deps()` where it appears as though iteration order through a dictionary could potentially affect results (as .`union()` on DMEM information flow nodes presently returns `None` instead of disjoint unions, meaning different orderings of `control_deps` could seemingly reveal different results). 

I'll run some trials later today to double-check that this could actually be an issue; in the meantime, let me know if there's something obvious I'm missing. Thanks!

(Alternatively, I can make a follow-up PR to address the potential ordering issues as needed; this putative issue doesn't seem to cause any issues in practice on this PR in my experience, and fixing it shouldn't take long.)

pqcfox (changes requested):

pqcfox:
Confirmed that the issue with `InformationFlowGraph.simplify()` (and thus presumably the other two issues) are actual issues that can show up in practice:

```shell
$ cd hw/ip/acc/util
$ PYTHONPATH=/home/kat/internal-pavona/util python3
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from shared.information_flow import *
>>> sink = DmemInformationFlowNode(0x0, 0x100)
>>> sources = [DmemInformationFlowNode(0x0, 0x100), DmemInformationFlowNode(0x200, 0x300), DmemInformationFlowNode(0x100, 0x200)]
>>> graph = InformationFlowGraph({sink: sources})  # calls self.simplify() internally
>>> graph.flow
{<shared.information_flow.DmemInformationFlowNode object at 0x784642fffbc0>: {<shared.information_flow.DmemInformationFlowNode object at 0x784642a30470>, <shared.information_flow.DmemInformationFlowNode object at 0x784642b2cf80>}}
```

(in the above, `graph.flow`'s lone key should have a singleton value with lone element being the DMEM range `0x0 - 0x300`)

Will make a quick patch and add it to this PR.

pqcfox:
Added a draft fix for the `simplify()` and related on my fork at `kat/iflow-dmem-backend-simplify-fix`, will make sure it works and add to here tomorrow.

pqcfox:
From chatting today, will split out the `simplify()` related suggested changes into a separate issue; took a bit more time to determine how to revise `simplify_control_deps()` than expected, and it may be a change large enough to be worthy of its own review.

hw/ip/acc/data/base-insns.yml
mkannwischer: They are now covered ` ISR_REMAPPING/SPECIAL_REG_NAMES`, yes.

hw/ip/acc/data/bignum-insns.yml
mkannwischer: same answer :-)

hw/ip/acc/util/shared/information_flow.py
mkannwischer: Fixed.

hw/ip/acc/util/shared/information_flow.py
mkannwischer: Fixed.

hw/ip/acc/util/shared/information_flow.py
mkannwischer: as discussed offline, let's look at this in a follow-up.

hw/ip/acc/util/shared/information_flow_analysis.py
mkannwischer: as discussed offline, let's look at this in a follow-up.

hw/ip/acc/util/shared/information_flow_analysis.py
mkannwischer: Fixed.

hw/ip/acc/util/shared/information_flow_analysis.py
mkannwischer: as discussed offline, let's look at this in a follow-up.

hw/ip/acc/data/bignum-insns.yml
mkannwischer: 0x8 is kmac_cfg, 0x9 is kmac_msg. 
wsr in this case is the kmac_core.
for kmac_cfg, KMAC is reset losing its state.
for kmac_msg, it writes the wsr, so the KMAC state depends on its previous state plus the wdr.
This does look correct to me. I think you could also write kmac_core instead of wsr.

What this is missing is 0xc (kmac_msg1) which only exists for the masked path. Very good that we looked at this again. I've added it now.

mkannwischer:
> From chatting today, will split out the `simplify()` related suggested changes into a separate issue; took a bit more time to determine how to revise `simplify_control_deps()` than expected, and it may be a change large enough to be worthy of its own review.

Thanks. Please open a follow-up PR when it's ready. I've addressed all other comments. Good catches!

hw/ip/acc/data/bignum-insns.yml
pqcfox: Ooh, good catch, and thank you for the explanation!

pqcfox (approved):
Looks fantastic! Thanks for the detailed replies on the above!

PR #297, opened 2026-07-03
[dma] Fix path to RAL spec in sim config

Razer6:
There is no `rvscs` path in Pavona ;-)


adurbin-zr (approved):

Razer6:
The failing FPGA builds are unrelated to this change.

PR #298, opened 2026-07-03
[hw,ac_range_check,rtl] Fix deny logging for unmatched address ranges

Razer6:
When no range matches an access, deny_mask is all zero and prim_leading_one_ppc defaults to deny_index 0. The logging gate checked log_enable_mask[0], so unmatched denies were only logged if range 0 happened to have logging enabled.

Add a no_range signal to bypass the per-range log enable check when deny_mask is empty. Apply the same fix in the DV predictor.


LouisTheLuis (approved):
LGTM.

adurbin-zr (commented):
in commit description:

s/signal tp bypass/signal to bypass/

adurbin-zr (approved):

PR #299, opened 2026-07-03
[hw,prim_flop_2sync] Model CDC race-through in 2-flop synchronizers

Razer6:
This updates the CDC random delay instrumentation to model three possible per-bit outcomes when a signal crosses into a destination clock domain:

- normal capture, where FF1 captures the new value and the synchronizer has the usual 2-cycle delay

- missed capture, where FF1 retains the previous value for one cycle and the synchronizer delay extends to 3 cycles

- race-through, where the input races through FF1 during a setup/hold violation window and FF2 can capture it after only 1 cycle

To support the race-through case, prim_cdc_rand_delay now emits a race_through_o mask alongside the delayed data, and prim_flop_2sync uses that mask to bypass FF1 on a per-bit basis in simulation. Non-simulation behavior remains unchanged.

The LC and MuBi synchronizer assertions are updated to account for the new 1-3 cycle synchronization window. For MuBi synchronizers with stability checking enabled, the expected output delay range is adjusted to include the additional stability flop on top of the randomized synchronizer delay.

This also updates the MuBi sync template so regenerated MuBi synchronizer variants keep the same assertion behavior.

### DIFF
--- a/hw/ip/prim/rtl/prim_cdc_rand_delay.sv
+++ b/hw/ip/prim/rtl/prim_cdc_rand_delay.sv
@@ -6,13 +6,24 @@
 // and allows DV to model real CDC delays within simulations, especially useful at the chip level
 // or in IPs that communicate across clock domains.
 //
-// The instrumentation is very simple: when this is enabled the input into the first
-// synchronizer flop has a mux where the select is randomly set. One of the mux inputs is the input
-// of this module, and the other is the output of the first flop: selecting the latter models the
-// effect of the first flop missing the input transition.
+// Models three physical outcomes per bit when a signal crosses clock domains:
 //
-// Notice the delay should cause the input to be skipped by at most a single cycle. As a perhaps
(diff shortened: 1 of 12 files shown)


matutem (dismissed):
This is a good improvement.

Please don't edit prim_mubi*_sync.sv file; instead edit util/design/gen-mubi.py and files under util/design/mubi or util/design/data.

Razer6:
@matutem The commit edits the corresponding template and re-generates the SV files using the generator script you mentioned. 

One request: could you run that commit through the regressions on your side? I have limited access.

matutem:
@Razer6 sorry about the delay: the hangup was because I have not run regressions. I just started some for spi_host and the alert_receiver_driver flags some errors: the logic tolerates some skew for asynchronous alerts, but something is going amiss. I am afraid this PR needs to be held until this problem is resolved.

For the record, it took some time for the original CDC instrumentation to be ready.

PR #301, opened 2026-07-06
 [acc] Fix ACCSim profiler per-function cycle counts

### LINKED ISSUE #296: [sw,acc] Simulator profiler reports incorrect per-function cycle counts
### Description

The ACC simulator's function profiler (`hw/ip/acc/dv/accsim/sim/stats.py`) reports nonsensical "functions" named `$d` and `$x` that absorb the vast majority of cycles, while real functions (`ntt`, `basemul`, `crypto_kem_keypair`,...) show near-zero counts. The per-function breakdown is therefore unusable.

### How to reproduce the error
- Running any SW tests on the ACC with `stats=True` will produce the error in the `test.log`.
- For the output below:  `bazel test //sw/acc/crypto/tests/mlkem:mlkem512_keypair_test0`
- Test run on commit [388dc77b33dbee58fb350e2ea83509ae17b37419](https://github.com/pavona/pavona/commit/f2bc21ca1e4c0373e940bd2a267ff13e9e09d12d).

### Output
```bash
Function cycle counts
-----------------------
function                [instr., stall]
----------------------  -----------------
$d                      [19119, 1007]
$x                      [5818, 4251]
ntt                     [68, 0]
cbd3                    [32, 0]
basemul                 [32, 0]
basemul_acc             [32, 0]
crypto_kem_keypair      [16, 3]
poly_getnoise_eta_init  [16, 0]
poly_gen_matrix_init    [16, 0]
poly_getnoise_eta_1     [8, 4]
poly_getnoise_cbd3      [8, 0]
poly_gen_matrix         [8, 0]
keccak_send_message     [6, 3]
pack_sk                 [5, 0]
pack_pk                 [5, 0]
poly_tomont             [2, 0]
poly_add                [2, 0]
_imem_start             [1, 1]
```

mkannwischer:
The profiler built its address->symbol map from every symbol in the
imem section, sweeping in the LLVM's $x/$d mapping symbols and
.L* labels. These outnumber real labels, so most instructions were
attributed to the nearest preceding mapping symbol.

One could filter such labels, but this commit instead takes the
more robust route and marks function labels using
`.type xxx, @function`, such that the profiler can filter for
such labels.

Additionally, this commit adds linting infrastructure for ACC asm
in ci/scripts/check-acc-asm.py currently consisting of a single
check: Every non-local (not .L or _ prefixed) needs to have a
corresponding .type annotation.

A few non-function labels get a _ prefix to pass the linting.

- Fixes https://github.com/pavona/pavona/issues/296.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,8 @@ jobs:
         run: ruff check
       - name: Validate testplans with schema
         run: ./ci/scripts/validate_testplans.sh
+      - name: ACC assembly lint
+        run: ./ci/scripts/check-acc-asm.py
       - name: C/C++ formatting
         run: ./bazelisk.sh test //quality:clang_format_check
       - name: Rust formatting
--- a/ci/jobs/quick-lint.sh
+++ b/ci/jobs/quick-lint.sh
@@ -52,6 +52,9 @@ ci/scripts/mypy.sh $tgt_branch
 echo -e "\n### Validate testplans with schema."
 ci/scripts/validate_testplans.sh
 
+echo -e "\n### Check ACC assembly lint"
+ci/scripts/check-acc-asm.py
+
 echo -e "\n### Use clang-format to check C/C++ coding style"
 ci/scripts/clang-format.sh $tgt_branch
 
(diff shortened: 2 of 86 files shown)


mkannwischer:
@phamhnh, could you give this a try please?

phamhnh:
> @phamhnh, could you give this a try please?

I ran the same test with this PR and my local repo which doesn't have this issue. Except for the `main` function is now stated as `<no function>` in the stats since it's not annotated with `.type xxx, @function` yet, everything else looks good to me. But as the stats of `main` is not needed for profiling, I don't think it needs to be annotated either. Thank you!

pqcfox (approved):
LGTM, thanks for the excellent care in fixing this :)

PR #300, opened 2026-07-06
[ci] Stop installing libtinfo5 in prepare-env

mkannwischer:
libtinfo5 is only required by Vivado. The installation for CI has been moved into the the runner image.

The following steps work without pinning a specific version:
```
  echo 'deb http://archive.ubuntu.com/ubuntu jammy universe' \
    | sudo tee /etc/apt/sources.list.d/jammy.list >/dev/null
  sudo apt-get update
  sudo apt-get download libtinfo5
  sudo dpkg -i libtinfo5_*.deb
  sudo rm -f libtinfo5_*.deb /etc/apt/sources.list.d/jammy.list
  sudo apt-get update
```

(Tested on Ubuntu 24.04).

 - Fixes https://github.com/pavona/pavona/issues/271
 
I've also updated the documentation to to include the setup for Vivado. However, that section wants a larger overhaul.

PR #302, opened 2026-07-06
[ci] Bump libtinfo5 to 6.3-2ubuntu0.2

mkannwischer:
The Ubuntu pool only keeps the latest version, so the 6.3-2ubuntu0.1 URL now returns 404. Also correct the TODO comment: the libtinfo5 dependency comes from Vivado not the RISC-V toolchain.

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -59,15 +59,15 @@ runs:
           sudo apt update --fix-missing
           grep '^[^#]' apt-requirements.txt | xargs sudo apt install -y
           sudo apt install wget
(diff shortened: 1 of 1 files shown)


mkannwischer:
We could merge this as a temporary workaround, until we got #300 to work.

pqcfox (approved):

pqcfox:
Looks great, thanks for handling this!

mkannwischer:
The lint job and the setup tests have passed, so this clearly works better than main.
merging to unblock other PRs.
We should work on a proper fix to drop that workaround though.

PR #303, opened 2026-07-06
[RFC] Propose DMA 2.0 enhancements

Razer6:
Add RFC 2026-04 describing the proposed DMA 2.0 enhancements:

- Generic ASID-indexed host ports for reuse across SoC tops
- Throughput improvements through address setup optimization and read/write overlap
- Inline AES-CTR/GCM support for encrypted DMA transfers
- Dedicated memset operation using the source address register as the fill value


domrizz0:
Thanks for proposing this @Razer6. You may also want to send a note to pavona@pavona.org letting people know. 

Added @pqcfox for vis on the AES-GCM dependency.

PR #304, opened 2026-07-07
[dvsim,util] Fix reseed naming issue

LouisTheLuis:
Due to the way that the `--reseed-source` flag for `dvsim` is set up, the `chip_*` tests get treated as `name=chip` and `variant=*`, which results in none of the chip tests being regraded.

This commit solves that by introducing a function that does a simple check on the `name` and `variant` to make sure that they are properly matched. Additionally, it uses a more robust way to determine the name and the variant of the grading report.

Also fixes the template for `rstmgr_cnsty_chk`.

### DIFF
--- a/hw/dv/grading/grading_report.hjson
+++ b/hw/dv/grading/grading_report.hjson
@@ -3218,7 +3218,7 @@
     }
     {
       test: chip_plic_all_irqs_0
-      name: chip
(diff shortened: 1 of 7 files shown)


qmn (approved):
Thanks for the fix! Looks good to me.

PR #305, opened 2026-07-08
[sw,rom] Merge discrete and integrated ROM implementations

rtorok-zr:
This PR performs a major refactor of the ROM code, which removes the separate `base_rom` used for the integrated (Dragonfly) design, and folds the integrated-specific features into the original (discrete/Egret) ROM target, with top-specific features controlled via `#ifdef` macros.
    
This refactor also involved a significant retooling of the ePMP setup, with a new "waterfall" slot allocation design that enables the ROM_EXT and later stages to perform the same ePMP setup regardless of whether the single- or dual-ROM boot sequence is used.

~Dependent on #245~


sw/device/silicon_creator/rom/e2e/rom_e2e_c_init_test.c
```diff
@@ -19,18 +22,18 @@
...
+  return dt_pinmux_reg_block(kDtPinmuxAon, kDtPinmuxRegBlockCore);
+}
+
+static inline uint32_t sram_main_mem_base(void) {
+  return dt_sram_ctrl_memory_base(kDtSramCtrlMain, kDtSramCtrlMemoryRam);
+}
+
+static inline uint32_t uart0_reg_base(void) {
+  return dt_uart_reg_block(kDtUart0, kDtUartRegBlockCore);
+}
```
qmn: Was there any reason you couldn't simply call dt_uart_reg_block(kDtUart0, kDtUartRegBlockCore) wherever you called uart0_reg_base()?

sw/device/silicon_creator/rom/e2e/rom_e2e_static_critical_test.c
```diff
@@ -45,8 +47,7 @@ void boot_measurements_test(void) {
   CHECK_WORD_ALIGNED(signed_region_size);
 
   dif_hmac_t hmac;
-  CHECK_DIF_OK(
-      dif_hmac_init(mmio_region_from_addr(TOP_EGRET_HMAC_BASE_ADDR), &hmac));
+  CHECK_DIF_OK(dif_hmac_init(mmio_region_from_addr(hmac_reg_base()), &hmac));
```
qmn, line 50:
```suggestion
  CHECK_DIF_OK(dif_hmac_init_from_dt(kDtHmac, &hmac));
```
Include `sw/device/lib/dif/autogen/dif_hmac_autogen.h` and then you should be able to get rid of hmac_reg_base entirely.

sw/device/silicon_creator/rom/e2e/watchdog/BUILD
```diff
@@ -50,7 +51,13 @@ otp_json(
...
     otp_image(
         name = "otp_img_watchdog_enable_{}".format(lc_state),
-        src = "//hw/top_egret/data/otp:otp_json_{}".format(lc_state),
+        src = pavona_select_top(
+            {
+                "egret": "//hw/top_egret/data/otp:otp_json_" + lc_state,
+                "dragonfly": "//hw/top_dragonfly/data/otp:otp_json_" + lc_state,
+            },
+            None,
+        ),
```
qmn, line 60: In some calls to `pavona_select_top()`, the default case is whatever applies to Egret. In this case it's `None`. Is there a general practice for establishing the default? Should this plainly fail if we're asked to create an OTP image for a top not appearing in this list, e.g. scafi_deprecated?

sw/device/silicon_creator/rom/boot_policy.c
```diff
@@ -6,15 +6,19 @@
 
 #include "sw/device/lib/base/hardened.h"
 #include "sw/device/silicon_creator/lib/base/chip.h"
-#include "sw/device/silicon_creator/lib/boot_data.h"
 #include "sw/device/silicon_creator/lib/drivers/lifecycle.h"
 #include "sw/device/silicon_creator/lib/error.h"
 #include "sw/device/silicon_creator/lib/shutdown.h"
 #include "sw/device/silicon_creator/rom/boot_policy_ptrs.h"
 
+#ifdef HAS_FLASH_CTRL
```
qmn, line 14: We'll want to rethink the implication of `HAS_FLASH_CTRL`, (if only the name) because the connection of having a flash controller to the presence of two slots is not immediately intuitive. (In this case, having a flash controller implies you have a flash, and with nonvolatile storage, you're interested in doing firmware upgrades, and so you might want to keep two firmware slots just in case the new firmware fails to verify.)

One not-very-developed idea is that you change use of `HAS_FLASH_CTRL` here to simply `TWO_BOOT_SLOTS`, and then have `HAS_FLASH_CTRL` also set `TWO_BOOT_SLOTS` by implication. That way, down the line, it's easier to have the presence of other non-volatile storage also imply having multiple boot slots.

sw/device/silicon_creator/rom/BUILD
```diff
@@ -44,16 +54,32 @@ dual_cc_library(
...
+    ) + pavona_select_top(
+        {
+            "egret": ["PAVONA_IS_EGRET"],
+            "dragonfly": ["PAVONA_IS_DRAGONFLY"],
+        },
+        [],
+    ),
     deps = dual_inputs(
+        # On designs with onboard flash, ROM_EXT is held there; otherwise it
+        # lives in CTN memory, which is goverened by the soc_proxy.
```
qmn:
```suggestion
        # lives in CTN memory, which is governed by the soc_proxy.
```

sw/device/silicon_creator/rom/BUILD
```diff
@@ -64,47 +90,160 @@ cc_library(
...
+    srcs = ["boot_rom_ext.c"],
+    hdrs = ["boot_rom_ext.h"],
+    defines = pavona_if_ip(
+        "keymgr_dpe",
+        ["HAS_KEYMGR_DPE"],
+        ["HAS_KEYMGR"],
+    ) + pavona_select_top(
+        {
+            "egret": ["DISCRETE_OTP_MAP"],
+            "dragonfly": ["INTEGRATED_OTP_MAP"],
```
qmn, line 187: The choice of name ("discrete" and "integrated" OTP maps) invites the question of whether the OTP map is a property of the top or a property of its implementation. The choice here suggests the latter, but I imagine it's actually a combination of more things: not just the intended implementation but also peripherals, other IP blocks, certain software needs. That said, it's the intended implementation (discrete or integrated) that most influences which OTP map you'll use, so for now these can stay.

sw/device/silicon_creator/lib/drivers/keymgr_dpe.h
```diff
@@ -88,9 +88,9 @@ typedef enum sc_keymgr_dpe_dest {
  * ```
  */
 enum {
-  kScKeymgrDpeSecMmioEntropyReseedIntervalSet = 1,
-  kScKeymgrDpeSecMmioSwBindingSet = 17,
-  kScKeymgrDpeSecMmioMaxVerSet = 2,
+  kScKeymgrSecMmioEntropyReseedIntervalSet = 1,
+  kScKeymgrSecMmioSwBindingSet = 9,
+  kScKeymgrSecMmioMaxVerSet = 2,
```
qmn, line 93: Could you clarify where these magic values come from?

sw/device/silicon_creator/lib/drivers/keymgr_dpe.c
qmn, line 1: Correct me if I'm wrong but this seems to be a simple deletion of "DPE" from the names of symbols in this file

sw/device/silicon_creator/second_rom/second_rom_start.S
```diff
@@ -106,7 +106,6 @@ _rom_interrupt_vector_c:
    * ROM shadow stack.
    */
   .section .bss
-  .balignl 4, UNIMP
```
qmn, line 109: Could you comment on this deletion?

sw/device/silicon_creator/second_rom/second_rom_start.S
```diff
@@ -178,17 +177,6 @@ _second_rom_start_boot:
   li x30, 0x0
   li x31, 0x0
 
-  // 2nd stage PMP setting
-  // Reset ROM0 slots
-  li   t0, 0
```
qmn, line 183: I know this is being deleted but RISC-V has a `zero` register

sw/device/silicon_creator/second_rom/second_rom_start.S
```diff
@@ -223,6 +211,10 @@ _second_rom_start_boot:
   la   t0, (_rom_interrupt_vector_c + 1)
   csrw mtvec, t0
 
+  // Advance ePMP state to lock access to ROM0.
+  .extern epmp_advance_boot_stage
+  call epmp_advance_boot_stage
```
qmn: Mostly for my own understanding: is it more idiomatic to advance the boot stage right before a jump? In other words, was there a conscious choice to advance boot stage here, rather than where ePMP slots 0-2 were cleared above?

sw/device/silicon_creator/rom/rom_epmp_test.c
```diff
@@ -236,12 +245,76 @@ static bool passed = false;
...
+/**
+ * Returns the size of the (first) ROM.
+ */
+static uint32_t rom_ctrl0_rom_size(void) {
+#ifdef HAS_ROM_CTRL1
+  return dt_rom_ctrl_memory_size(kDtRomCtrl0, kDtRomCtrlMemoryRom);
+#else
+  return dt_rom_ctrl_memory_size(kDtRomCtrl, kDtRomCtrlMemoryRom);
+#endif
+}
```
qmn: For later: It's hard to imagine a reason to have more than two ROMs but hopefully we can have an indirection on which rom_ctrl we're talking to.

sw/device/silicon_creator/rom/rom_epmp_init.S
```diff
@@ -1,10 +1,19 @@
...
-#include "hw/top_egret/sw/autogen/top_egret_memory.h"
+#if defined(PAVONA_IS_EGRET)
+ #include "hw/top_egret/sw/autogen/top_egret_memory.h"
+#define MMIO_BASE_ADDR TOP_EGRET_MMIO_BASE_ADDR
+#define MMIO_SIZE_BYTES TOP_EGRET_MMIO_SIZE_BYTES
+#elif defined(PAVONA_IS_DRAGONFLY)
+#include "hw/top_dragonfly/sw/autogen/top_dragonfly_memory.h"
+#define MMIO_BASE_ADDR TOP_DRAGONFLY_MMIO_BASE_ADDR
+#define MMIO_SIZE_BYTES TOP_DRAGONFLY_MMIO_SIZE_BYTES
+#endif
```
qmn, line 18: A separate PR, but this should be made top-agnostic (topgen should be able to help out)

sw/device/silicon_creator/rom/rom_epmp_init.S
```diff
@@ -22,12 +31,6 @@
  */
 #define TOR(addr) ((addr) >> 2)
 
-/**
- * Encode address and length for Naturally-Aligned-Power-Of-Two addressing
- * mode.
- */
-#define NAPOT(addr, len) (((addr) >> 2) | (((len) - 1) >> 3))
-
```
qmn, line 30: It might be helpful to keep this unless there's a reason it's going away (other than being unused)?

sw/device/silicon_creator/rom/rom_epmp.c
```diff
@@ -4,81 +4,58 @@
...
-      HARDENED_CHECK_EQ(lc_state, kLcStateProdEnd);
-      debug_rom_access = kEpmpPermLockedNoAccess;
-      break;
-    case kLcStateRma:
-      HARDENED_CHECK_EQ(lc_state, kLcStateRma);
-      debug_rom_access = kEpmpPermLockedReadWriteExecute;
-      break;
-    default:
-      HARDENED_TRAP();
-  }
```
qmn, line 81: Did I remember correctly that debug rom has unconditionally open access now, and so this section's not necessary?

sw/device/silicon_creator/rom/rom_start.S
```diff
@@ -143,64 +147,83 @@ _rom_shadow_stack:
...
   // Check if we should halt here.
-  li   a0, (TOP_EGRET_OTP_CTRL_CORE_BASE_ADDR + \
-            OTP_CTRL_SW_CFG_WINDOW_REG_OFFSET)
-  lw   t0, OTP_CTRL_PARAM_CREATOR_SW_CFG_ROM_EXEC_EN_OFFSET(a0)
+  call otp_sw_cfg_base
+  // Save the OTP SwCfg window address (no previous value to save; we are the
+  // entry point).
+  mv s0, a0
+#ifdef DISCRETE_OTP_MAP
+  lw   t0, OTP_CTRL_PARAM_CREATOR_SW_CFG_ROM_EXEC_EN_OFFSET(s0)
```
qmn: The test here should really be for whether the exec enable OTP field exists, as this block concerns whether ROM execution is enabled, not intrinsically whether we are in a discrete chip.

sw/device/silicon_creator/rom/rom_start.S
```diff
@@ -143,64 +147,83 @@ _rom_shadow_stack:
...
   // Set up the global pointer and re-enable linker relaxations.
   la gp, __global_pointer$
   .option pop
 
 LABEL_FOR_TEST(kRomStartBootMaybeHalt)
   // Check if we should halt here.
-  li   a0, (TOP_EGRET_OTP_CTRL_CORE_BASE_ADDR + \
-            OTP_CTRL_SW_CFG_WINDOW_REG_OFFSET)
-  lw   t0, OTP_CTRL_PARAM_CREATOR_SW_CFG_ROM_EXEC_EN_OFFSET(a0)
+  call otp_sw_cfg_base
```
qmn: I don't have too much experience working deeply in the ROM here, but seeing calls to C code this early give me a little pause. I know that `otp_sw_cfg_base()` is simply a read of the base address of the OTP controller register block, but I can imagine that some subsequent change to `otp_sw_cfg_base()` inadvertently adds stack operations, and there's not a story in the ROM so far about the value of `sp`. Is there no way to know this value at compile time to render the `call` unnecessary?

sw/device/silicon_creator/rom/rom_start.S
```diff
@@ -278,16 +306,16 @@ LABEL_FOR_TEST(kRomStartWatchdogEnabled)
   // |------------------+----------------+----------------+------------------|
   // | Configuration:   | Input          | Input          | Input            |
   // |                  | Pull-down      | Pull-down      | Pull-up          |
```
qmn, line 311: Last I recall, straps are only in the discrete part.

sw/device/silicon_creator/rom/rom_start.S
```diff
@@ -315,8 +343,8 @@ LABEL_FOR_TEST(kRomStartWatchdogEnabled)
 
   // Double-check the GPIO strap value.
   li   t6, HARDENED_BOOL_TRUE
-  li   a1, TOP_EGRET_GPIO_BASE_ADDR
-  lw   t1, GPIO_DATA_IN_REG_OFFSET(a1)
+  call gpio_reg_base
+  lw   t1, GPIO_DATA_IN_REG_OFFSET(a0)
```
qmn: Wonder if there's any significance to putting the GPIO base addr in a1 in the second check as opposed to a0 again.

sw/device/silicon_creator/rom/rom_start.S
qmn: The section named was removed in f15aff5ca5c658033e4587a186719e23707b71a7 without replacement.

sw/device/silicon_creator/rom/rom_start.S
```diff
@@ -462,6 +497,20 @@ LABEL_FOR_TEST(kRomStartWatchdogEnabled)
...
+  // call these functions from within `rom_epmp_init` because nested function
+  // calls access the stack, which we cannot use until we finish setting up
+  // ePMP.
+  call sram_ctrl_ram_napot
+  mv a2, a0
+#ifdef HAS_FLASH_CTRL
+  call flash_ctrl_mem_napot
+#else
+  call soc_proxy_ctn_napot
+#endif
```
qmn: This is another example of why I think `HAS_FLASH_CTRL` needs renaming: it's not clear why we'd use soc_proxy if we don't have a flash_ctrl.

sw/device/silicon_creator/rom/rom_start.S
qmn, line 515: Okay, so at last, the stack pointer is initialized here. It's hard for me to stomach seeing uses of `call` that jump to C code (no matter how much one assures me that the C code doesn't call into other code, doesn't use the stack, etc). Is there any way we can use a preprocessor macro that helps us initialize values from DT without having to go to C code?

sw/device/silicon_creator/rom/rom.c
```diff
@@ -298,249 +328,124 @@ static rom_error_t rom_init(void) {
...
 
-  // Check cached boot data.
-  rom_error_t boot_data_ok = boot_data_check(&boot_data);
-  if (launder32(boot_data_ok) != kErrorOk) {
-    HARDENED_TRAP();
-  }
-  HARDENED_CHECK_EQ(boot_data_ok, kErrorOk);
-  CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomPreBootCheck, 4);
+  CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomBoot, 4);
+  uintptr_t _entry_point = ((uintptr_t)_second_rom_boot_address) + 0x180;
```
qmn: Could you explain the magic `0x180`?

sw/device/silicon_creator/rom/rom.c
```diff
@@ -553,95 +458,32 @@ OT_WARN_UNUSED_RESULT
 static rom_error_t rom_boot(const manifest_t *manifest,
                             uintptr_t imm_section_entry_point,
                             uint32_t flash_exec) {
+#ifdef HAS_ROM_CTRL1
```
qmn: This might be the GitHub PR review interface but I believe this is a second level of #ifdef in the code. This harks back to my comment about using preprocessor macros in general, but this second level is prone to confusion. Can we break out all two-ROM code into another file?

sw/device/silicon_creator/rom/rom.c
```diff
@@ -751,17 +528,27 @@ OT_WARN_UNUSED_RESULT
...
+  CFI_FUNC_COUNTER_PREPCALL(rom_counters, kCfiRomTryBoot, 2, kCfiRomVerify);
+  rom_error_t error =
+      rom_verify(manifests.ordered[0], boot_data, lc_state, &sigverify_ctx,
+                 &flash_exec, &rom_verify_increment_cfi);
+#else
   CFI_FUNC_COUNTER_PREPCALL(rom_counters, kCfiRomTryBoot, 2, kCfiRomVerify);
-  rom_error_t error = rom_verify(manifests.ordered[0], &flash_exec);
+  rom_error_t error = rom_verify(manifests.ordered[0], lc_state, &sigverify_ctx,
+                                 &flash_exec, &rom_verify_increment_cfi);
+#endif
```
qmn: It might be better to not have function prototypes change depending on presence of flash_ctrl

qmn (approved):
Firstly, great work on unifying (and yet factoring out) these two ROM implementations!

The full implications of all the changes are admittedly challenging to fully grok in one pass. Much of the proof of this ROM is in the pudding, that is, the more robust check is that it passes CI. Of course, downstream integrators are going to want to modify this ROM, and the changes here (especially with ePMP management) are welcome and remove a lot of redundant changes. Still, it's nice to have a common ROM that supports both discrete and integrated with a minimum of duplication.

Some fast comments about what I've seen:
* Not for now, but in dealing with the boot policy, `HAS_FLASH_CTRL` could be better named something like `TWO_BOOT_SLOTS`.
* One suggestion on use of DT.
* The use of `call` in early ROM code gives me pause.

One stylistic comment on the ROM, for a future PR, is that there is a lot of preprocessor use in rom.c. There weren't nearly as many #ifdefs in the original ROM. Also, there are ifdefs in function prototypes which really shouldn't be a thing. It would be much easier to read if we can hide away the preprocessor macros behind function calls that no-op if the corresponding IP block doesn't exist.

To review this PR, I first went through the Bazel files and metadata (and linker scripts). Then, I looked at the ePMP code. Then using the ROM (previously known as base ROM or mask ROM), I looked over the overall boot flow: rom_start, rom, second_rom, rom_ext. I looked at header files for sanity, then skimmed over the ROM contents. Most _new_ code was in ePMP.

I would generally say that this code all looks good to me. I'm going to leave my approval here, but I'd like someone better versed in ROM to give their approval too.

sw/device/silicon_creator/rom/e2e/rom_e2e_c_init_test.c
rtorok-zr: No, and what you said is admittedly easier. I was just used to doing it this way because of the (somewhat cursed) method of bringing their return values into assembly, which we discussed workarounds for last week.

sw/device/silicon_creator/rom/e2e/watchdog/BUILD
rtorok-zr: I wanted this to fail-fast if you're using an unsupported top, so I did it this way.

sw/device/silicon_creator/rom/boot_policy.c
rtorok-zr: I was able to simplify this to always have `boot_policy` always interpret as having 2 boot slots, but `boot_policy_ptrs` just returns the same slot twice, which `boot_policy` adds a short-circuit for. Ended up being much cleaner.

sw/device/silicon_creator/rom/BUILD
rtorok-zr: Makes sense. FYI, these will be removed in a follow-up PR that unifies the OTP layout.

sw/device/silicon_creator/lib/drivers/keymgr_dpe.c
rtorok-zr: Yes, it made it easier to call the right one for the correct top without having a bunch of `ifdef`s in the calling files. This can probably be reverted once Egret is migrated to DPE.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: This code is now removed entirely after reverting the method used to fetch the memory address constants.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: As discussed offline yesterday, there is no way to pull directly from `dt` without an explicit `call` instruction, because C (unlike C++ and Rust) does not allow evaluation of constant functions at compile-time.

To avoid invoking `call` instructions prior to the stack being set up, which is extremely fragile (remove one compiler optimization and you get two nested calls, which faults without a stack), I adjusted this back to the old method of using `li` directly, but with lightweight preprocessor macros that reduce the number of places that require referencing the top name directly.

sw/device/silicon_creator/rom/rom.c
rtorok-zr: `0x180` is the entry point of the second ROM relative to its base address. The fact that this is hard-coded is a known problem and is planned to be addressed in a future PR, possibly with help from topgen.

sw/device/silicon_creator/rom/rom.c
rtorok-zr: Fixed this. `boot_data` is now unconditionally included; `boot_data.c` now conjures up a default `boot_data` struct when there is no flash to load from.

sw/device/silicon_creator/rom/rom.c
rtorok-zr: This is addressed by the `boot_second_rom.c` / `boot_rom_ext.c` split mentioned below.

sw/device/silicon_creator/lib/drivers/keymgr_dpe.h
rtorok-zr: These track how many times certain keymgr APIs are probed. The middle value changed because set fewer distinct `sw_binding` values in DPE, because there are no separate sealing/attestation chains in HW.

sw/device/silicon_creator/second_rom/second_rom_start.S
rtorok-zr: I moved this to `second_rom.c` where slots 12-13 are set because there's little benefit to having it in assembly.

sw/device/silicon_creator/rom/rom_epmp_test.c
rtorok-zr: For now, replaced this with a nice `dt` trick: `kDtRomCtrlFirst`, but your comment is still valid.

sw/device/silicon_creator/rom/rom_epmp_init.S
rtorok-zr: This is back because the workaround I had to do before to pre-compute the ePMP values to avoid nested `calls` is no longer needed.

sw/device/silicon_creator/rom/rom_epmp.c
rtorok-zr: Correct. RV_DM no longer needs an explicit ePMP slot because it automatically has access everywhere when debugging is enabled.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: This will be removed in a follow-up PR that will introduce the `ROM_EXEC_EN` flag in Dragonfly.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: You're right that this was fragile. Nested `call`s do indeed fault prior to setting up the stack, and we happened to get away with it here because of a compiler optimization within the functions called from within assembly. After offline discussion, this was reverted to use safer `li` instructions with constant values, but with the toplevel name distilled to the top of the file using a preprocessor macro.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: Yep, when going back over this I was actually able to `#ifdef` out on the integrated design a larger block of assembly code that was checking for a strap to spin for an "enter RMA" signal.

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: Obsoleted

sw/device/silicon_creator/rom/rom_start.S
rtorok-zr: Good catch. I'll update this to point somewhere else when I update the docs in a future series of PRs.

rtorok-zr:
Based on the review comments, I've made three major updates to this PR:

1. Reverted the `dt`-based address constant fetching from the ROM assembly code (which was brittle to compiler optimizations when executing without a stack), and replaced it with a variant of the original `li` method that filters the top-level name to a preprocessor macro at the top of the file.
2. Distilled the `#ifdef`s related to `boot_data`/`bootstrap` only working when `flash_ctrl` is present to the relevant code in their respective libraries that actually use `flash_ctrl` APIs, removing the function prototype differences and creating a cleaner flow.
3. Factored more of the `rom_verify`/`rom_boot` flow for loading ROM_EXT into `boot_rom_ext.c`, which allowed the consolidation of several of the other separate library files I had previously introduced. There is now a clean split with `rom.c` calling into either `boot_second_rom.c` or `boot_rom_ext.c` depending on the design, and the ugly CFI increment callbacks have been removed thanks to an adjustment to the CFI macros.

ngildenhuys (approved):

PR #306, opened 2026-07-08
[topgen] Use JSON schema for data validation in topgen

vlpannel:
Currently, the required and allowed fields and data types for a top level configuration file are detailed in `util/topgen/validate.py` as a series of Python lists. This format is not straightforward to find nor very flexible. Instead, [JSON schema](https://json-schema.org/) allows users to more easily see and manipulate the requirements for metadata files such as top configuration Hjsons. It also allows schemas to be more easily documented with `jsonschema2md`.

This PR is limited to the schemas for topgen; future PRs will address the schemas for reggen/the IP block description Hjsons, dvsim/testplan Hjsons, and others.

Schemas also only address the required and optional keys and value types needed in the metadata, *not* whether or not the information can be actually feasible in hardware. Future work may consider grouping together more meaningful checks (for example, ensuring no invalid combination of options is selected or making sure that every module's assigned clocking exists at the top level as well).


vlpannel:
_Rationale for a few design decisions made in this PR (happy to change with feedback):_
* Schemas were placed under `util/basegen/schemas`, assuming that there exists a builtin (hence, placed in basegen) schemas library with the possibility to extend it/add more schema libraries.
  * This could easily be moved elsewhere in the repo, though, if there is a better place to put them.
* Schemas [can be nested](https://json-schema.org/understanding-json-schema/structuring), but this would limit the flexibility of them.
  * For now, each schema has been made standalone (using `"type": "object"` rather than `"$ref": ...` for fields) and individual tools can structure hierarchies as they like.
  * Schemas have no real semantic understanding (limited checks to see if a given configuration makes sense), so nesting should probably be done in the code (Python) instead.
* Basegen now has a `cast_hjson_values` function that automatically casts dictionary values to what they should be.
  * The repo's Hjson style guide allows for putting a value of any type in quotation marks, but this causes the `hjson` library to automatically cast these values as strings.
  * The goal is to eventually have this function obsolete the currently-used `check_bool`/`check_int`/etc. functions throughout the tooling.
  * This does, however, mean that you cannot have the string `"2"` in your config, as it will be automatically cast to the int `2`.
  * ~~Speaking of Hjson style, will fix up the schema Hjsons' formatting with a new commit later; for now, the content is correct despite bad/non-conformant style.~~ Hjson formatting now fixed.

util/basegen/schemas/eflash.hjson
```diff
@@ -0,0 +1,62 @@
...
+      "type": "number",
+      "description": "maximum number of flash words allowed to program at a time"
+    },
+    "words_per_page": {
+      "type": "number",
+      "description": "number of words per page"
+    },
+  }
+
+  "required": [
```
matutem: Please sort this (and perhaps other lists or maps) for ease of use.

matutem:
It may be useful to reference schemas of nested fields.

util/BUILD
```diff
@@ -111,6 +111,7 @@ py_binary(
         "//util/topgen",
         requirement("mako"),
         requirement("hjson"),
+        requirement("jsonschema"),
```
adurbin-zr, line 127: This seems unrelated to this particular commit. Make the change on its own?

util/basegen/schemas/pinmux_signal.hjson
```diff
@@ -26,5 +26,5 @@
     }
   }
 
-  "required": ["instance", "connection"]
```
adurbin-zr: could you please rebase the change onto the corresponding commit?

vlpannel:
> It may be useful to reference schemas of nested fields.

I think so; it makes more sense to check that _all_ needed metadata is present once in the script rather than multiple checks in the Python.

I also have decided to move the schemas files to `util/topgen/schemas/` instead of basegen because the top configuration is really only useful for topgen for now. If later the schemas are generalizable to other tools, then things can be adapted.

util/basegen/schemas/eflash.hjson
vlpannel: Now, all schema fields are sorted more by how they are in the example Hjsons.

vlpannel:
This PR is quite big, so I will close it and split it up into three smaller PRs.

It has been broken up into #387, #388, #389.

PR #307, opened 2026-07-08
[dvsim] Include wall clock time per test

LouisTheLuis:
This will improve the user's access to telemetry and for resource allocation. It also includes a "Simulation Wall Clock Time" table for the purpose of understanding how long a regression is.


qmn (approved):
Looks good to me! Not for this PR, but we should collect finer-grain timing information at the per-run level too.

PR #222, opened 2026-06-01
[df,rom] Backport second-stage ROM for integrated tops

rtorok-zr:
This PR backports the second-stage ROM for integrated top-level designs, which was originally developed primary by @sameo on the `integrated-dev` branch.

Also included are a few new features that were not present in the original implementation:
- A sigverify toolchain for verifying ROM_EXT by @pqcfox 
- An update to the ePMP setup step that enables the ROM to work on `exec_env`s where Ibex virtual addressing is disabled.

PR #221, opened 2026-06-01
[rom] Backport ROM for integrated top-level designs

rtorok-zr:
Duplicate of #182, which was closed due to fork provenance issues after `pavona/pavona` was made public.

Review discussion is located on #182.

PR #308, opened 2026-07-09
[dj,sw] Run hello_world binary on Dragonfly + Verilator

rtorok-zr:
This PR builds on the tooling introduced in previous PRs (#179, #221, #222, #245) to run an end-to-end software test for Dragonfly on Verilator.

This PR also contains a few miscellaneous improvements that did not fit well in the previous PRs:
- Enable `sigverify` support on Dragonfly as part of the `rom1` in #222 using the OTP signing keys in #245.
- Fix an issue with the clock settings for Dragonfly/Verilator.

Duplicate of #189 as a result of the repository becoming public.

### DIFF
--- a/ci/scripts/run-verilator-tests.sh
+++ b/ci/scripts/run-verilator-tests.sh
@@ -5,9 +5,41 @@
 
 set -e
 
+if [ $# != 1 ]; then
+    echo >&2 "Usage: run-verilator-tests.sh top_name"
+    echo >&2 "E.g. ./run-verilator-tests.sh egret"
+    exit 1
+fi
+
+TOP_NAME=$1
+
+TESTS=(
+    "//sw/device/tests:aes_smoketest_sim_verilator"
+    "//sw/device/tests:uart_smoketest_sim_verilator"
+    "//sw/device/tests:crt_test_sim_verilator"
+    "//sw/device/tests:acc_randomness_test_sim_verilator"
+    "//sw/device/tests:acc_irq_test_sim_verilator"
+    "//sw/device/tests:kmac_mode_cshake_test_sim_verilator"
+    "//sw/device/tests:kmac_mode_kmac_test_sim_verilator"
+    "//sw/device/silicon_creator/lib/drivers:hmac_functest_sim_verilator"
+    "//sw/device/silicon_creator/lib/drivers:uart_functest_sim_verilator"
+    "//sw/device/silicon_creator/lib/drivers:retention_sram_functest_sim_verilator"
+)
+
+if [ "${TOP_NAME}" == "egret" ]; then
+    TESTS+=("//sw/device/silicon_creator/lib/drivers:alert_functest_sim_verilator")
+    TESTS+=("//sw/device/tests:flash_ctrl_test_sim_verilator")
+    TESTS+=("//sw/device/silicon_creator/rom:rom_epmp_test_sim_verilator")
+    TESTS+=("//sw/device/tests:usbdev_test_sim_verilator")
+    # TODO: these two tests hang on Dragonfly.
+    TESTS+=("//sw/device/silicon_creator/lib/drivers:watchdog_functest_sim_verilator")
+    TESTS+=("//sw/device/silicon_creator/lib:irq_asm_functest_sim_verilator")
+fi
+
 # Increase the test_timeout due to slow performance on CI
 
 ./bazelisk.sh test \
+    --//hw/top="${TOP_NAME}" \
     --build_tests_only=true \
     --test_timeout=2400,2400,4000,-1 \
     --local_test_jobs=8 \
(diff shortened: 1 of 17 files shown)


ci/scripts/run-verilator-tests.sh
```diff
@@ -5,29 +5,45 @@
...
     --//hw:make_options=-j,8 \
-    //sw/device/tests:aes_smoketest_sim_verilator \
-    //sw/device/tests:uart_smoketest_sim_verilator \
-    //sw/device/tests:crt_test_sim_verilator \
-    //sw/device/tests:acc_randomness_test_sim_verilator \
-    //sw/device/tests:acc_irq_test_sim_verilator \
-    //sw/device/tests:kmac_mode_cshake_test_sim_verilator \
-    //sw/device/tests:kmac_mode_kmac_test_sim_verilator \
-    //sw/device/tests:flash_ctrl_test_sim_verilator \
-    //sw/device/tests:usbdev_test_sim_verilator \
```
qmn, line 26: Looks like you dropped usbdev, was this intentional?

hw/top_dragonfly/sw/dt/sim_verilator.c
```diff
@@ -9,7 +9,7 @@
 // FIXME Are those really fixed in DV? Does it matter?
 static const uint32_t clock_freqs[kDtClockCount] = {
     [kDtClockMain] = 500 * 1000,
-    [kDtClockIo] = 250 * 1000,
+    [kDtClockIo] = 500 * 1000,
```
qmn, line 15: It would be worth leaving a note to keep these values in sync with sw/device/lib/arch/device_sim_verilator.c, as well as to drop a pointer to the Verilator SystemVerilog testbench. We could also elaborate that FIXME above: yes, it's fixed in DV, and yes, it matters, e.g. because uartdpi depends on it:
https://github.com/pavona/pavona/blob/f25ee3449a0612a3ce6db5a77a5128a956342a4e/hw/top_dragonfly/dv/verilator/chip_sim_tb.sv#L55-L62

ci/scripts/run-verilator-tests.sh
```diff
@@ -5,29 +5,45 @@
...
+    "//sw/device/silicon_creator/lib/drivers:hmac_functest_sim_verilator"
+    "//sw/device/silicon_creator/lib/drivers:uart_functest_sim_verilator"
+    "//sw/device/silicon_creator/lib/drivers:retention_sram_functest_sim_verilator"
+)
+
+if [ "${TOP_NAME}" == "egret" ]; then
+    TESTS+=("//sw/device/silicon_creator/lib/drivers:alert_functest_sim_verilator")
+    TESTS+=("//sw/device/tests:flash_ctrl_test_sim_verilator")
+    TESTS+=("//sw/device/silicon_creator/rom:rom_epmp_test_sim_verilator")
+    # TODO: these two tests hang on Dragonfly.
```
pqcfox, line 34: (may be useful to add issue for this)

sw/device/silicon_creator/rom/sigverify_key_types.h
```diff
@@ -268,6 +312,12 @@ typedef struct sigverify_rom_spx_key_entry {
    * Type of the key.
    */
   sigverify_key_type_t key_type;
+#ifdef OTP_KEY_ROLE
+  /**
+   * Role of the key.
+   */
+  sigverify_key_role_t key_role;
+#endif
```
qmn, line 320: Appears that `OTP_KEY_ROLE` is only defined for the integrated / Dragonfly top. Would you be able to clarify what key roles are for, and why it only applies to integrated?

qmn (commented):
Looks mostly good! Thanks for making the Verilator tests more portable in general. A question about the usbdev test and also looking for clarification on OTP_KEY_ROLE.

sw/device/silicon_creator/rom/sigverify_key_types.h
pqcfox: (as a follow-up to this, wanted to confirm that `key_role` is a placeholder for later--though it does of course affect the later field offsets, so of course makes sense to include here--and that this struct member isn't referenced yet anywhere yet?)

pqcfox (approved):
Looks excellent; interested in the answers to the above questions, but these aren't blocking for me to approve. (Edit: saw some of Quan's additional comments; do think those make sense to address first, but will keep the approval in place.)

ci/scripts/run-verilator-tests.sh
rtorok-zr: No, thanks for catching. I temporarily removed it to check what would build on Dragonfly, but I meant to keep it in for discrete.

sw/device/silicon_creator/rom/sigverify_key_types.h
rtorok-zr: These were included in the Dragonfly OTP mmap in https://github.com/lowRISC/opentitan/pull/28840. These are intended to be removed in the a later PR for the OTP zipper merge, because there is only one reasonable (valid) value for each key.

PR #310, opened 2026-07-10
[reggen,rtl] Derive alert fatality from reg metadata

Razer6:
Emit AlertIsFatal from reggen using the existing alert metadata and wire prim_alert_sender.IsFatal from that generated vector. This removes local hardcoding of alert fatality while keeping the classification sourced from the register description.

### DIFF
--- a/hw/ip/ascon/rtl/ascon_reg_pkg.sv
+++ b/hw/ip/ascon/rtl/ascon_reg_pkg.sv
@@ -16,6 +16,21 @@ package ascon_reg_pkg;
   // Address widths within the block
   parameter int BlockAw = 8;
 
+  // Number of registers for every interface
+  parameter int NumRegs = 47;
+
+  // Alert indices
+  typedef enum int {
+    AlertRecovCtrlUpdateErrIdx = 0,
+    AlertFatalFaultIdx = 1
+  } ascon_alert_idx_t;
+
+  // Fatal alert classification
+  localparam bit [NumAlerts-1:0] AlertIsFatal = {
+    1'b0, // recov_ctrl_update_err
+    1'b1 // fatal_fault
+  };
+
   ////////////////////////////
   // Typedefs for registers //
   ////////////////////////////
@@ -171,30 +186,18 @@ package ascon_reg_pkg;
     struct packed {
       logic        d;
       logic        de;
-    } start;
+    } wipe;
     struct packed {
       logic        d;
       logic        de;
-    } wipe;
+    } start;
   } ascon_hw2reg_trigger_reg_t;
 
   typedef struct packed {
     struct packed {
       logic        d;
       logic        de;
-    } idle;
-    struct packed {
-      logic        d;
-      logic        de;
-    } stall;
-    struct packed {
-      logic        d;
-      logic        de;
-    } wait_edn;
-    struct packed {
-      logic        d;
-      logic        de;
-    } ascon_error;
+    } alert_fatal_fault;
     struct packed {
       logic        d;
       logic        de;
@@ -202,22 +205,34 @@ package ascon_reg_pkg;
     struct packed {
       logic        d;
       logic        de;
-    } alert_fatal_fault;
+    } ascon_error;
+    struct packed {
+      logic        d;
+      logic        de;
+    } wait_edn;
+    struct packed {
+      logic        d;
+      logic        de;
+    } stall;
+    struct packed {
+      logic        d;
+      logic        de;
+    } idle;
   } ascon_hw2reg_status_reg_t;
 
   typedef struct packed {
     struct packed {
-      logic        d;
+      logic [1:0]  d;
       logic        de;
-    } msg_valid;
+    } tag_comparison_valid;
     struct packed {
       logic        d;
       logic        de;
     } tag_valid;
     struct packed {
-      logic [1:0]  d;
+      logic        d;
       logic        de;
-    } tag_comparison_valid;
+    } msg_valid;
   } ascon_hw2reg_output_valid_reg_t;
 
   typedef struct packed {
@@ -228,11 +243,7 @@ package ascon_reg_pkg;
     struct packed {
       logic        d;
       logic        de;
-    } no_key;
-    struct packed {
-      logic        d;
-      logic        de;
-    } no_nonce;
+    } flag_input_missmatch;
     struct packed {
       logic        d;
       logic        de;
@@ -240,7 +251,11 @@ package ascon_reg_pkg;
     struct packed {
       logic        d;
       logic        de;
-    } flag_input_missmatch;
+    } no_nonce;
+    struct packed {
+      logic        d;
+      logic        de;
+    } no_key;
   } ascon_hw2reg_error_reg_t;
 
   // Register -> HW type
--- a/hw/ip/keymgr/rtl/keymgr.sv
+++ b/hw/ip/keymgr/rtl/keymgr.sv
@@ -742,7 +742,7 @@ module keymgr
   prim_alert_sender #(
     .AsyncOn(AlertAsyncOn[1]),
     .SkewCycles(AlertSkewCycles),
-    .IsFatal(1)
+    .IsFatal(AlertIsFatal[1])
   ) u_fault_alert (
     .clk_i,
     .rst_ni,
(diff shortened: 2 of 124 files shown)


hw/ip_templates/otp_ctrl/rtl/otp_ctrl.sv.tpl
```diff
@@ -719,7 +719,7 @@ end
     prim_alert_sender #(
       .AsyncOn(AlertAsyncOn[k]),
       .SkewCycles(AlertSkewCycles),
-      .IsFatal(AlertIsFatal[k])
+      .IsFatal(AlertIsFatal[i])
```
qmn: The genvar is on `k` so I think either this one should go back to `k` or the rest should be changed to `i`.

hw/ip/keymgr/rtl/keymgr.sv
```diff
@@ -742,7 +742,7 @@ module keymgr
   prim_alert_sender #(
     .AsyncOn(AlertAsyncOn[1]),
     .SkewCycles(AlertSkewCycles),
-    .IsFatal(1)
+    .IsFatal(AlertIsFatal[i])
```
qmn: It looks like the prim_alert_senders here are instantiated by hand, and not in a for loop, so I don't think `i` is defined here. For better symmetry we could do `AlertIsFatal[0]` here: 
```suggestion
    .IsFatal(AlertIsFatal[0])
```
(and `AlertIsFatal[1]` below)

hw/ip/lc_ctrl/rtl/lc_ctrl.sv
```diff
@@ -651,7 +651,7 @@ module lc_ctrl
     prim_alert_sender #(
       .AsyncOn(AlertAsyncOn[k]),
       .SkewCycles(AlertSkewCycles),
-      .IsFatal(1)
+      .IsFatal(AlertIsFatal[i])
```
qmn: The `genvar` here is on `k`:
```suggestion
      .IsFatal(AlertIsFatal[k])
```
(or change them all to `i`)

hw/ip/keymgr_dpe/rtl/keymgr_dpe.sv
```diff
@@ -762,7 +762,7 @@ module keymgr_dpe
   prim_alert_sender #(
     .AsyncOn(AlertAsyncOn[1]),
     .SkewCycles(AlertSkewCycles),
-    .IsFatal(1)
+    .IsFatal(AlertIsFatal[i])
```
qmn: Same as the comment in keymgr:
```suggestion
    .IsFatal(AlertIsFatal[0])
```

hw/ip/keymgr_dpe/rtl/keymgr_dpe.sv
qmn:
```suggestion
    .IsFatal(AlertIsFatal[1])
```

hw/ip/sram_ctrl/rtl/sram_ctrl.sv
```diff
@@ -207,7 +207,7 @@ module sram_ctrl
   prim_alert_sender #(
     .AsyncOn(AlertAsyncOn[0]),
     .SkewCycles(AlertSkewCycles),
-    .IsFatal(1)
+    .IsFatal(AlertIsFatal[i])
```
qmn:
```suggestion
    .IsFatal(AlertIsFatal[0])
```

hw/ip_templates/rv_core_ibex/rtl/rv_core_ibex_peri.sv.tpl
```diff
@@ -99,8 +99,6 @@ module ${module_instance_name}_peri
   assign alert_test[3] = reg2hw.alert_test.recov_hw_err.q &
                          reg2hw.alert_test.recov_hw_err.qe;
 
-  localparam bit [NumAlerts-1:0] AlertFatal = '{1, 0, 1, 0};
```
qmn, line 102: It seems that this `AlertFatal` array in rv_core_ibex_peri.sv.tpl (1, 0, 1, 0) is inverted from the one in rv_core_ibex.sv.tpl (0, 1, 0, 1). But in the version templatized for dragonfly/egret, only the `AlertIsFatal` array from the regular rv_core_ibex appears. Should this `AlertFatal` remain?

hw/ip_templates/flash_ctrl/rtl/flash_ctrl.sv.tpl
```diff
@@ -955,12 +955,11 @@ module flash_ctrl
   // reset for these bits.
   //
   // For more details, refer to lowRISC/OpenTitan#21353.
-  localparam logic [NumAlerts-1:0] IsFatal = {1'b0, 1'b1, 1'b0, 1'b1, 1'b0};
   for (genvar i = 0; i < NumAlerts; i++) begin : gen_alert_senders
     prim_alert_sender #(
       .AsyncOn(AlertAsyncOn[i]),
       .SkewCycles(AlertSkewCycles),
-      .IsFatal(IsFatal[i])
+      .IsFatal(i == AlertFatalErrIdx? 1'b0 : AlertIsFatal[i])
```
qmn, line 962: This is a bit unusual but given the comment above, we'll have to go with this for now. An alternate way of writing this is also:

```suggestion
      .IsFatal(AlertIsFatal[i] && i != AlertFatalErrIdx)
```

hw/ip/ascon/rtl/ascon_reg_pkg.sv
qmn, line 1: A bit curious to see that the register fields are reversed. Has it been that long since ascon's been re-generated?

qmn (commented):
Hi, thanks so much for the work! A few minor comments, mostly due to some genvars being `k` instead of the much-more-common `i`; others because there isn't a for loop over the `prim_alert_senders` at all:

hw/ip_templates/flash_ctrl/rtl/flash_ctrl.sv.tpl
Razer6: I wanted to be explicit and see the intent to set it to zero.

hw/ip/ascon/rtl/ascon_reg_pkg.sv
Razer6: Ascon is not yet production-ready and is therefore not included in the Makefile. The register generation flow was reversed at some point last year, but Ascon was never regenerated afterward. I manually regenerated the `reg_top` for Ascon in this change.

hw/ip_templates/rv_core_ibex/rtl/rv_core_ibex_peri.sv.tpl
Razer6: Yes, they are inverted, but I believe the current ordering is incorrect because it does not match the alert definitions. The ordering was corrected for `rv_core_ibex` in this commit, but the corresponding issue in the peripheral module was not fixed:

https://github.com/lowRISC/opentitan/commit/3cf0831f67d7adc16171f182f32961361ce5d981

Razer6:
Thanks for the review @qmn ! Those were some silly mistakes on my part. I appreciate you taking the time to catch them.

hw/ip_templates/rv_core_ibex/rtl/rv_core_ibex_peri.sv.tpl
qmn: Looks good, it looks like rv_core_ibex_peri isn't used at all anyway. It looks like f9a8ca24 folded a lot of `rv_core_ibex_peri` into simply `rv_core_ibex`. Thanks!

hw/ip_templates/flash_ctrl/rtl/flash_ctrl.sv.tpl
qmn: Sounds good!

hw/ip/ascon/rtl/ascon_reg_pkg.sv
qmn: Thanks so much for the clarification!

qmn (approved):
Thanks, always glad to help! LGTM now. That CW340 bitstream build failure is new to me, but shouldn't be connected to this.

PR #311, opened 2026-07-10
[hw,mbx,rtl] Gate DOE interrupt with enable

Razer6:
Honor the DOE interrupt enable when driving doe_intr_o, so the status bit only raises the output interrupt when DOE IRQ support is present and the interrupt is enabled.


qmn (approved):
LGTM, thanks!

PR #312, opened 2026-07-10
[hw,prim_prince,rtl] Correctly index from data_state_lo

Razer6:
The computation of data_state_middle_d used data_state_middle[NumRoundsHalf] as the sbox input, which is wrong. Correctly change that to data_state_lo[NumRoundsHalf] like it is used in the 64-bit instantiation.

Note, Pavona does not instantiate prim_prince in a 32-bit config, that's why this issue was never observed.


apinise (approved):
LGTM! Thanks for finding this, definitely a bug.

PR #313, opened 2026-07-10
[hw,prim_alert_sender,dv] Fix Ack seq copy/paste error

Razer6:
Fix a copy/paste error for the AckSigInt sequence. Correctly use the alert_*_i.ack signal instead of the ping signal.


ysekhar (approved):
LGTM

PR #314, opened 2026-07-10
[hw,pwrmgr,rtl] Correctly report escalation reset status

Razer6:
Previously, the escalation reset status CSR was sourced by the main-power-glitch reset bit, instead of rstreqs[ResetEscIdx]. The reset behavior was correct but the reporting was wrong.

This PR changes that to correctly index via ResetEscIdx.

### DIFF
--- a/hw/ip_templates/pwrmgr/rtl/pwrmgr.sv.tpl
+++ b/hw/ip_templates/pwrmgr/rtl/pwrmgr.sv.tpl
@@ -574,7 +574,7 @@ module pwrmgr
   end
 
   assign hw2reg.escalate_reset_status.de = 1'b1;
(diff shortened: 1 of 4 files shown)


matutem (approved):
Funny it took so long to fix this. You mean the escalate reset status was always wrong? There must have been a test that kept failing, or else some escalation test was not checking the reset_status CSR? Please respond to this so can close any DV hole.

LGTM

Razer6:
Yes, the reset status CSRs were always wrong. The reason this was not seen in DV is because the status registers are not read in the testbench. In https://github.com/pavona/pavona/blob/main/hw/top_dragonfly/ip_autogen/pwrmgr/dv/env/pwrmgr_scoreboard.sv#L339, the read check for those registers was disabled.

PR #315, opened 2026-07-10
[hw,soc_dbg_ctrl,rtl] Valid is software-controlled in Prod-state

Razer6:
In the production state, the valid bit of the debug category is determined by the software congtrolled CSR. It is not True all the time as currently present in the RTL.


LouisTheLuis (approved):
LGTM. It would be a good idea to update the `README.md` as well.

PR #319, opened 2026-07-13
[hw,soc_dbg_ctrl_decode,rtl] Safer decoding of relock signal

Razer6:
Decode unlocked only when reloacted is strictly false. Any invalid MuBi value shouled behave locked.


LouisTheLuis (approved):
LGTM.

PR #321, opened 2026-07-14
[verilator] Depend on full Verilator install for FuseSoC builds

### LINKED ISSUE #320: [verilator] Can't exec "verilator_bin": No such file or directory
I'm occasionally getting the following error when running a verilator test
```
ERROR: Can't exec "verilator_bin": No such file or directory at /home/matthias/.cache/bazel/_bazel_matthias/32f9016b04354262e73122e4804f800d/sandbox/processwrapper-sandbox/28/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/third_party/verilator/build_verilator/bin/verilator line 214.
Use of uninitialized value in string eq at /home/matthias/.cache/bazel/_bazel_matthias/32f9016b04354262e73122e4804f800d/sandbox/processwrapper-sandbox/28/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/third_party/verilator/build_verilator/bin/verilator line 214.
sh: 1: exec: verilator_bin: not found
%Error: Command Failed ulimit -s unlimited 2>/dev/null; exec verilator_bin -f lowrisc_dv_top_egret_chip_verilator_sim_0.1.vc -DDISABLE_PRIM_CDC_RAND_DELAY --no-timing --trace --trace-fst --trace-structs --trace-params --trace-max-array 1024 --unroll-count 512 -CFLAGS \ -std=c++17\ -Wall\ -DVM_TRACE_FMT_FST\ -DVL_USER_STOP\ -DTOPLEVEL_NAME=chip_sim_tb -LDFLAGS \ -pthread\ -lutil\ -lelf -Wall --threads 4 -Wno-fatal --threads 4
make: *** [Makefile:15: Vchip_sim_tb.mk] Error 127
```

Today, I figured out how to reproduce it: The error occurs only when verilator itself is cached AND the verilated model is not cached. In case verilator is not cached, everything works fine. In case the model is cached, everything works fine.

mkannwischer:
The FuseSoC action declares only the `verilator` wrapper as its verilator tool, but the wrapper exec()s `verilator_bin` from the same configure_make output. Bazel only materializes the output group a consumer references on a cache restore, so when the toolchain is restored from a cache only the wrapper lands on disk and the build fails with `Can't exec "verilator_bin"`. A fresh build materializes the whole install, so this only surfaces on a cache restore combined with a model cache miss resulting in a FuseSoC invocation.

Depend on the full installation instead of just the wrapper.

 - Resolves https://github.com/pavona/pavona/issues/320

### DIFF
--- a/rules/fusesoc.bzl
+++ b/rules/fusesoc.bzl
@@ -95,7 +95,7 @@ def _fusesoc_build_impl(ctx):
             ctx.executable._ar,
             ctx.executable._cc,
             ctx.executable._cxx,
-        ]) if ctx.attr.target != "synth" else [],
+        ] + ctx.files._verilator_install) if ctx.attr.target != "synth" else [],
         arguments = [args],
         executable = ctx.executable._fusesoc,
         use_default_shell_env = False,
@@ -165,6 +165,12 @@ fusesoc_build = rule(
             executable = True,
             default = "//third_party/verilator",
             doc = "Verilator binary target",
+            cfg = "exec",
+        ),
+        "_verilator_install": attr.label(
+            allow_files = True,
+            default = "//third_party/verilator:verilator_install",
+            doc = "Full Verilator installation staged alongside the wrapper.",
             cfg = "exec",
         ),
         "_ar": attr.label(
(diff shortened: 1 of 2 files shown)


rtorok-zr (approved):
Thanks for tracking this down. I tried combining the two `filegroup` targets into a single one with `output_groups` removed, but that prevents Bazel from finding the verilator binary.

PR #323, opened 2026-07-14
[rstmgr_cnsty_chk,dv] Update sim_cfg

LouisTheLuis:
Include an override for its scratch path so that it includes the variant; otherwise, when `egret` and `dragonfly` runs happen concurrently, they will overwrite each other's scratch directory.


ysekhar (approved):
LGTM

matutem (approved):
LGTM

PR #324, opened 2026-07-14
[dvsim] Publish report.json privately

LouisTheLuis:
This JSON file gets generated at the end of a run. It can be used for grading later and to store information about the run. With this commit, now it can be publishable with `--publish-mode private`.


ysekhar (approved):
LGTM

matutem (approved):
LGTM

PR #328, opened 2026-07-15
[ci, verilator] Add a `PQCEn=1` verilator execution environment and exercise in CI

mkannwischer:
- Depends on #334 

This PR does 3 things (each in a separate commit):
1) It wires up the PQCEn parameter to the verilator model exposed as `//hw:verilator_pqc`  and registers it as a new egret execution environment.
2) It adds a CI run that duplicates the current verilator job to run the same fast tests on the PQCEn=1 model
3) It extends the nightly Verilator jobs to additionally run ML-KEM and ML-DSA functests (those take around an hour each) against both models.

Still work in progress if any timeouts need to be extended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,24 @@ jobs:
         with:
           artifact-name: verilator_egret-test-results
 
+  verilator_egret_pqc:
+    name: Egret Verilator PQC Tests
+    runs-on:
+      group: pavona-large
+    needs: quick_lint
+    timeout-minutes: 240
+    steps:
+      - uses: actions/checkout@v7
+      - name: Prepare environment
+        uses: ./.github/actions/prepare-env
+      - name: Run fast Verilator tests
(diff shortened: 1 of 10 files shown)


pqcfox:
(Added #334 to help as well with a Verilator issue I observed w/ `PQCEn=1`.)

mkannwischer:
> (Added #334 to help as well with a Verilator issue I observed w/ `PQCEn=1`.)

Thanks! I have cherry-picked that commit here. Let's see if that works.

mkannwischer:
@pqcfox, that does not seem to fix it for the PQCEn=1 configuration. 
If you have some time to take a look, I'd greatly appreciate it.

pqcfox:
Let me take a look now, thanks for heads up!

pqcfox:
Hmm, I can't reproduce either timeout locally--both tests pass quickly locally without flakiness, so I think this is load-related.

PR #329, opened 2026-07-15
[ci] Update rule on check-sim-cfgs

LouisTheLuis:
The script was ignoring the case for which a `sim_cfg` existed under `ip_autogen` with a `variant`, which currently is most autogen'd blocks. It also supports extensions of `all_sim_cfgs.hjson` for specific projects.

#330 must be merged first.


ci/scripts/check-sim-cfgs.py
```diff
@@ -136,13 +161,24 @@ def check_sim_cfg_files(grouped: dict) -> bool:
...
-                                      f" should be named {expected}")
-                elif 'name' not in fields:
-                    mismatches.append(f"Filename {directory}/{file}"
-                                      f" should contain field 'name'")
+                if file == "all_sim_cfgs.hjson":
+                    if 'name' not in fields:
+                        mismatches.append(f"Filename {directory}/{file}"
+                                          f" should contain field 'name'")
+                else:
+                    # Additional batch cfgs under hw/dv/ are variant selections
```
apinise: What would a variant of the `all_sim_cfgs.hjson` look like?

LouisTheLuis: A file as follows:
```
{
  // Name of the sim cfg variant
  variant: {variant}

  import_cfgs: [// Project wide common cfg file
                "{proj_root}/hw/dv/all_sim_cfgs.hjson"]

  // Need to override the default output directory
  overrides: [
    {
        name: scratch_path
        value: "{scratch_base_path}/all_tops_batch-sim"
    }
  ]

  use_cfgs: [
    /* list of configs */
  ]
 
  /*config specific*/
}
```
The point being that in specific projects, we might want to isolate a subset of the `sim_cfgs` for our preference or to add additional constraints (e.g. run_modes or build_modes). `all_sim_cfgs.hjson` is auto-generated, but for project-specific usage, it might be useful to allow more specific configurations to exist.

LouisTheLuis: Actually, based on #332, removing this check.

apinise (approved):
LGTM

PR #330, opened 2026-07-15
[flash_ctrl,dv] Simplify sim_cfg

LouisTheLuis:
We do not need a `base_sim_cfg` now that `dvsim` can downstream the `tool` key.


apinise (approved):
Cleanup LGTM!

PR #332, opened 2026-07-15
[dvsim] Propagate configuration from batch-level to tests

LouisTheLuis:
This PR is composed of two commits:
- The first commit aims to allow `run_timeout_mins` to be set within a run mode for a regression and to be downstreamed to the tests, overriding their run timeout. This is done so that separate regressions might have different timeouts. It's also useful at the batch level: we might want a single block to have a certain timeout whereas the rest run with the default.
- The second commit introduces the concept of `cfg_groups`: similar to how `regressions` allows the user to set up particular configuration parameters for multiple tests, this concept allows parameters to be set for a particular subset of blocks, which can then be called with the `--cfg-groups` argument.

For an illustration, this allows batch `sim_cfgs` to be shaped as follows:
```
{ 
    name: test_batch
    import_cfgs: ["{proj_root}/hw/data/common_project_cfg.hjson"]
    flow: sim 
    rel_path: "hw/dv/summary"
    overrides: [
      { name: scratch_path, value: "{scratch_base_path}/{name}-{flow}" }
    ]

    // Pool of cfgs the groups draw from — batchgroups untouched, exactly the
    // all_sim_cfgs shape (copy or subset).
    use_cfgs: {
      ip: [
        "{proj_root}/hw/ip/acc/dv/uvm/acc_standard_sim_cfg.hjson",
        "{proj_root}/hw/ip/uart/dv/uart_sim_cfg.hjson",
        "{proj_root}/hw/ip/aes/dv/aes_masked_sim_cfg.hjson",
        "{proj_root}/hw/ip/hmac/dv/hmac_sim_cfg.hjson",
      ]
    }
    
    // Named groups: subset + per-group scalar overrides.
    cfg_groups: {
      group1: {
        cfgs: ["acc_standard", "uart"]
        run_timeout_mins: 360
        reseed: 3
      }
      crypto: {
        cfgs: ["aes_masked", "hmac"]
        run_timeout_multiplier: 2
      }                                                                                                                                                                                                             
    }                                                                                                                                                                                                               
  }     

```

### DIFF
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -47,6 +47,11 @@ class FlowCfg():
         self.items = list(dict.fromkeys(args.items))
         self.list_items = args.list
         self.select_cfgs = args.select_cfgs
+        self.cfg_group_sel = args.cfg_groups
+        self.cfg_groups = None
+        # Scalar overrides captured from the selected cfg_groups entry, stamped
+        # onto each selected child cfg (see prune_selected_cfgs).
+        self.group_overrides = {}
         self.flow_cfg_file = flow_cfg_file
         self.args = args
         self.scratch_root = args.scratch_root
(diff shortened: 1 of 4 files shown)


qmn (approved):
Thanks for the illuminating example. LGTM

PR #333, opened 2026-07-15
[util] Filter per-top sim_cfgs by hjson

LouisTheLuis:
`gen_ip_collection.py` shared every `hw/ip` block across all tops, so both top-specific `sim_cfg` listed blocks that their tops don't instantiate.

This commit reads the top's hjson `module` list, scanning for `type` and only keeping an IP block if and only if the IP is contained there. UVCs and prims are nested infrastructure so they are always kept.


qmn (approved):
Great idea, LGTM, thanks!

PR #335, opened 2026-07-20
[dvsim] Allow for reseed_multiplier to be inherited from regression

LouisTheLuis:
This way, the user can run multiple blocks at a time each with a separate reseed multiplier; leading to more flexibility.


matutem (approved):
LGTM

PR #337, opened 2026-07-21
[spi,work-around] Fix workaround for wait statement

### LINKED ISSUE #28: [spi_agent,monitor] Work-around apparent simulator bug
### Description

A simulator used for DV mistakenly causes the statement
  `wait (&cfg.vif.csb === 0);`
to succeed for csb set to all ones. We confirmed this with
uvm_info invocations right after the wait. This statement is in
spi_monitor.sv, task collect_trans, which is called in a forever loop.

The work-around that succeeded was the logically equivalent
  `@cfg.vif.csb;
   if (&cfg.vif.csb !== 1'b0) return;`

The return causes the task to be invoked again, which is equivalent
to ignoring transitions to all ones and trying again.

matutem:
Replace the work-around for VCS faulty handling of `wait (&cfg.vif.csb===1'b0)` by `wait (cfg.vif.csb == '0)`, as suggested by VCS support. The work-around was put in place in #71.

Fixes #28

### DIFF
--- a/hw/dv/sv/spi_agent/spi_monitor.sv
+++ b/hw/dv/sv/spi_agent/spi_monitor.sv
@@ -47,12 +47,7 @@ class spi_monitor extends dv_base_monitor#(
     bit flash_opcode_received;
 
     wait (cfg.en_monitor);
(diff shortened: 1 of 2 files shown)


ysekhar (approved):
LGTM

PR #339, opened 2026-07-24
[fusesoc] Prioritize Bazel-provided hermetic Verilator in build

pqcfox:
This one-commit PR tweaks fusesoc_build.py to always prioritize the `VERILATOR_BINARY`-provided Verilator binary over any pre-existing Verilator binaries in the path, ensuring presence of a separate Verilator version in the path cannot break builds.

As an alternative to modifying `PATH`, I observed that edalize accepts a `VERILATOR_ROOT` environment variable for setting the Verilator root path. Unfortunately, edalize also searches for the Verilator `include` directory in the same place, which is incompatible with how Bazel presently builds the hermetic Verilator; changing how `PATH` was set in fusesoc_build.py was the most reasonable approach I could identify.

### DIFF
--- a/util/fusesoc_build.py
+++ b/util/fusesoc_build.py
@@ -80,11 +80,13 @@ if __name__ == "__main__":
     make_options += " LDFLAGS='{}'".format(" ".join(ldflags))
 
     # `VERILATOR_BINARY` is passed in as a relative path to the Bazel
-    # `execroot`. Convert this to an absolute path in PATH, if present.
+    # `execroot`. Convert this to an absolute path in PATH, if present, at the
+    # start of PATH to ensure the hermetic Verilator binary is selected over any
+    # locally installed version.
     if "VERILATOR_BINARY" in os.environ:
         verilator_binary = os.environ["VERILATOR_BINARY"]
         verilator_dirname = os.path.dirname(os.path.join(cwd, verilator_binary))
-        path_env += os.pathsep + verilator_dirname
+        path_env = verilator_dirname + os.pathsep + path_env
         # If running verilator, include make options.
         sys.argv = others + ["--make_options=" + make_options]
     os.environ["PATH"] = path_env


rtorok-zr (approved):
LGTM, this looks like a more robust version of #177

util/fusesoc_build.py
```diff
@@ -80,11 +80,13 @@
...
     # `VERILATOR_BINARY` is passed in as a relative path to the Bazel
-    # `execroot`. Convert this to an absolute path in PATH, if present.
+    # `execroot`. Convert this to an absolute path in PATH, if present, at the
+    # start of PATH to ensure the hermetic Verilator binary is selected over any
+    # locally installed version.
     if "VERILATOR_BINARY" in os.environ:
         verilator_binary = os.environ["VERILATOR_BINARY"]
         verilator_dirname = os.path.dirname(os.path.join(cwd, verilator_binary))
-        path_env += os.pathsep + verilator_dirname
+        path_env = verilator_dirname + os.pathsep + path_env
```
fragglet, line 89: Nit: might be slightly clearer to use [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join):
```suggestion
        path_env = os.path.join(verilator_dirname, path_env)
```

pqcfox:
@rtorok-zr Apologies, I missed #177 in tossing this together! It seems like maybe rebasing this atop #177 could be useful, e.g. to ensure that `VERILATOR_*` env variables are set properly; should I rebase atop that and wait for it to merge first?

@fragglet agree on that change, I'll push in a moment to include that revision w/ signoff. Thanks!

util/fusesoc_build.py
ngildenhuys: this is not the same, `path.join` appends a directory onto an existing path.
the string addition here using `os.pathsep` is adding a new full path to the path env string.
https://docs.python.org/3/library/os.html#os.pathsep
https://docs.python.org/3/library/os.path.html#os.path.join

pqcfox: Thank you for catching this, I'll revert now and push.

pqcfox:
@fragglet @rtorok-zr CI should be working now for this PR (just rerunning, hit an issue I think resolved by a now-merged PR that just needed to be incorporated). If it passes, feel free to merge as you need, otherwise I'll be sure to merge tomorrow morning. Thanks!

mkannwischer:
> @fragglet @rtorok-zr CI should be working now for this PR (just rerunning, hit an issue I think resolved by a now-merged PR that just needed to be incorporated). If it passes, feel free to merge as you need, otherwise I'll be sure to merge tomorrow morning. Thanks!

Hmm, there was a linting failure related to https://github.com/pavona/pavona/commit/5fea7bd6d3b06c369f0ae5fd312ff11dad81c47f - that should not be there. I have rebased your commits on top of main to check if that resolves it.

PR #340, opened 2026-07-25
[acc,mlkem] Add masked ML-KEM decapsulation and key generation

mkannwischer:
- Depends on #289 
- Hoisted out from #230

Wires the ML-KEM ACC implementation up to the masking gadgets from #289, adding
a first-order side-channel-hardened build of ML-KEM decapsulation and key generation.
The same sources compile to `run_mlkem` (unprotected) and
`run_mlkem_hardened` (`-DHARDENED`, Boolean-masked keygen and decap).

This is hoisted out of the original masked ML-KEM draft in #230, split into
reviewable pieces (gadgets in #289, this implementation here, cryptolib to
follow).

Most of the code was written by @phamhnh; with some polishing by @pqcfox and myself.

PR #338, opened 2026-07-21
[acc,mldsa] Add ML-DSA masking gadgets.

mkannwischer:
This commit adds masking gadgets required for an ML-DSA implementation hardened against first-order side-channel attacks. The ML-DSA implementation itself is not yet adjusted - that will be done in a follow up.
The masking gadgets are currently undergoing side-channel evaluation and minor changes are still expected.

- Hoisted out from https://github.com/pavona/pavona/pull/261

Compared to the code in https://github.com/pavona/pavona/pull/261, the following changes have been applied

- Gadgets are now combined into a single gadgets.s
 - Code size has been reduced
 - Documentation has be unified
 - `endloop` pseudo-instructions have been added
 - Whitening has been added in some places
 - The bibliography has been cleaned up

PR #261, opened 2026-06-16
[acc,masking] Add first-order masked ML-DSA key generation and signing

mkannwischer:
This PR adds a masked ML-DSA implementation for Pavona's ACC.

It adds first-order-DPA protection to ML-DSA-44, -65, and -87 signing and key generation. Developed in parallel with last week's masked ML-KEM work, it follows the same design: it uses the first-order masked KMAC hardware already present in the ACC, with everything else in software to set a first baseline to build on. It fits within the 32 KB of DMEM and IMEM in the default ACC post-quantum configuration. Overhead versus the unprotected implementation is 2.4–2.8× for key generation and 3.6–4.1× for signing. This implementation is currently undergoing SCA evaluation.

I have also written a [blog post](https://www.zerorisc.com/blog/hardened-pqc-on-pavona-part-ii-masking-ml-dsa) about this implementation.

This draft PR is ready for evaluation, but there's still work we'd like to do:

 - Restructuring the implementation to align with the other ACC code 
 - Consolidating the documentation with the ML-KEM implementation
 - Extend the testing

We'd welcome community feedback while we incorporate these changes.

PR #341, opened 2026-07-27
[acc,mldsa] Add masked ML-DSA signing and key generation

mkannwischer:
- Depends on #338
- Hoisted out from #261
- Mirrors #340 for ML-DSA

Wires the ML-DSA ACC implementation up to the masking gadgets from #338, adding a first-order side-channel-hardened build of ML-DSA key generation and signing. The same sources compile to `run_mldsa` (unprotected) and `run_mldsa_hardened` (`-DHARDENED`, Boolean-masked keygen and signing); verification only ever touches public data and stays unmasked in both builds.

This is hoisted out of the original masked ML-DSA draft in #261, split into reviewable pieces (gadgets in #338, this implementation here, cryptolib to follow), and follows the masked ML-KEM implementation in #340.

Beyond the gadget call sites, DMEM pressure forced some changes:

- The DMEM layout moves into its own `mldsa_dmem.s`, worst-case (ML-DSA-87) sized, with buffers of disjoint operations overlaid.
- Hardened builds drop the static twiddle tables (and a few other constant tables) from the loaded image and rebuild them into `scratch` at runtime (`gen_twiddles.s`); this is the only difference in verify between the two builds. The main reason that made this change necessary was that ACVP tests so far only provide expanded secret keys making it impossible to store secret keys as seeds instead. This has changed in the recent v1.1.0.43 ACVP pre-release, so we can consider switching back to a seed key format which would enable precomputing twiddle factors and storing them in DMEM.
- The previous implementation was wrongly unmasking+packing z and rtilde directly after their bounds check had passed even though it must remain secret until all bounds checks have passed. The unmasking+packing is now moved to the end.
- A companion change keeps the NOLOAD scratchpad from being assigned a load address, which would otherwise overflow the loaded ELF image.

Tests now use the memory layout from `run_mldsa` and `run_mldsa_hardened`, rather than allocating their own DMEM buffers.
For ML-DSA signing we add the 10-vector mean benchmark set from https://github.com/zerorisc/mldsa-sign-bench as described in https://www.zerorisc.com/blog/efficiently-measuring-ml-dsa-signing-performance


pqcfox:
Started in on a review today, made it a good bit through; will continue chipping away tomorrow :) thanks!

mkannwischer:
@dop-amin, could you kindly also take a look at this PR?

sw/acc/crypto/mldsa/mldsa_dmem.s
```diff
@@ -0,0 +1,368 @@
...
+.set keygen_tmp, sig + 2816
+
+/* c_poly (NTT(c) for masked sign + verify) overlays the BIG_BUF tail; the .set
+ * is emitted after w1_polyvec below (it must be defined first). */
+
+/* rhoprime/ctilde (sign/verify) alias zeta_shares; w1_repvec (verify) aliases
+ * sk; the masked-sign scratch (sign_y/sign_tmp/sign_w1_repvec/sign_gamma1_buf)
+ * aliases the sign-dead pk; mask_stack aliases tmp_poly (all .set above). */
+
+/* Big polyvec buffer shared by disjoint ops: keygen t (2*K*1024), masked sign
```
pqcfox: [nit] possible tweak, for clarity of references to `BIG_BUF` 
```suggestion
/* Big polyvec buffer (BIG_BUF) shared by disjoint ops: keygen t (2*K*1024), masked sign
```

sw/acc/crypto/mldsa/mldsa_dmem.s
```diff
@@ -0,0 +1,368 @@
...
+w1_polyvec:
+    .zero 16384
+/* Verify packs the z shares into the BIG_BUF tail: w1_polyvec uses [0:8192],
+ * z_polyvec uses [8192:15360] (L*1024 = 7168 worst case). */
+.globl z_polyvec
+.set z_polyvec, w1_polyvec + 8192
+/* NTT(c) challenge buffer (masked sign + verify) reuses the BIG_BUF tail: verify
+ * leaves [15360:16384] free after w1[0:8192]+z[8192:15360], and masked sign's
+ * dead region past the packed-b'/c_poly_shares rows covers the same top slot.
+ * Matches the per-op sign overlay (mldsa_sign_bss.s), just shifted to clear z. */
```
pqcfox: [nit] (does `mldsa_sign_bss.s` still exist?)

sw/acc/crypto/mldsa/mldsa_dmem.s
```diff
@@ -0,0 +1,368 @@
...
+.globl y_poly
+.globl s1_poly
+.set c_poly,  w0_polyvec_shares + 15360
+.set y_poly,  w0_polyvec_shares + 15360
+.set s1_poly, w0_polyvec_shares + 15360
+
+/* Masked-sign overlays into w0_polyvec_shares (W0_POLYS=8, W0_SHARE_STRIDE=
+ * 8192, NSHARES=2 fixed). sign_w0_l2_* hardcode the K=4 offset since the L2
+ * decompose path only runs for ML-DSA-44 (c_poly is a standalone buffer above
+ * to avoid colliding with the unprotected verify's c_poly). */
```
pqcfox: [nit] 
```suggestion
 * to avoid the unprotected verify's c_poly colliding with sign_c_poly_shares. */
```

sw/acc/crypto/tests/mldsa/gen_twiddles_inv_test.s
```diff
@@ -0,0 +1,29 @@
+/* Copyright zeroRISC Inc. */
+/* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+/**
+ * Test for the inverse twiddle generation (gen_twiddles_fwd + _inv_transform):
+ * the result written to `scratch` must match intt's runtime a506 derivation.
```
pqcfox: [clarification q] what's a506 here?

sw/acc/crypto/mldsa/mldsa_verify.s
```diff
@@ -210,12 +210,23 @@ _ctilde_unpack_done:
...
+     * ntt(t1). */
+    jal  x1, gen_twiddles_fwd
+#endif
     /* NTT(z) */
     la   a0, z_polyvec
     addi a2, a0, 0 /* inplace */
 
     lw t0, MLDSA_PARAM_L_OFFSET(s11)
+#ifdef HARDENED
+    LOOP t0, 4
```
pqcfox: [super nit] for consistency could change this plus pre-existing `LOOP` -> `loop` (similar below)

mkannwischer: I fixed this in a separate commit for the entire ML-DSA code.

sw/acc/crypto/tests/mldsa/gen_twiddles_inv_test.s
mkannwischer: That was outdated, sorry. Fixed.

sw/acc/crypto/mldsa/mldsa_dmem.s
mkannwischer: Fixed.

sw/acc/crypto/mldsa/mldsa_dmem.s
mkannwischer: Good catch. That got consolidated in the meantime.

sw/acc/crypto/mldsa/mldsa_dmem.s
mkannwischer: Added.

sw/acc/crypto/mldsa/gen_twiddles.s
```diff
@@ -0,0 +1,471 @@
...
+ * inv[255] = ninv).
+ *
+ * Entry points:
+ *   gen_twiddles_fwd - build the forward table in `scratch`.
+ *   _inv_transform   - convert the forward table in `scratch` to the inverse.
+ */
+
+.text
+
+/* gen_twiddles_fwd: regenerate the forward NTT twiddle table (256 words).
```
pqcfox, line 42: [super nit] could replace this with the standard function header format

sw/acc/crypto/mldsa/gen_twiddles.s
```diff
@@ -0,0 +1,471 @@
...
+    bn.mulv.l.8S.even.acc.z.lo w30, w0, sw1.7
+    bn.mulv.l.8S.even.lo       w30, w30, sw0.1
+    bn.mulv.l.8S.even.acc.hi   w30, w30, sw0.0
+    bn.mulv.l.8S.odd.acc.z.lo  w30, w30, sw1.7
+    bn.mulv.l.8S.odd.lo        w30, w30, sw0.1
+    bn.mulv.l.8S.odd.acc.hi    w30, w30, sw0.0
+    bn.mov w0, w30
+    bn.sid x0, 0(x7++)
+    la   x6, scratch
+    sw   x0, 60(x6)
```
pqcfox, line 342: [clarifying q] what does this store do at the end of `gen_twiddles_fwd`?

sw/acc/crypto/mldsa/mldsa_keypair.s
```diff
@@ -86,6 +86,495 @@
...
+    li t0, 31
+    addi t1, s2, 0
+    lw   t2, MLDSA_PARAM_K_OFFSET(s11)
+    loop t2, 3
+        loopi 32, 1
+          bn.sid t0, 0(t1++)
+        endloop
+        nop
+    endloop
+    lw   t2, MLDSA_PARAM_K_OFFSET(s11)
```
pqcfox: [super nit] second K load can be dropped I think (?)

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+    bn.xor  w0, w0, w0            /* Whitening */
+    bn.lid  x0, 32(t0)
+    bn.xor  w0, w0, w2
+    bn.wsrw kmac_msg1, w0
+
+    /* Send rnd as (rnd, 0): public input trivially shared on share 0. */
+    la     t0, rnd
+    bn.lid x0, 0(t0)
+    bn.wsrw kmac_msg, w0
+    bn.xor w0, w0, w0
```
pqcfox: [super nit] line here (and two below) could be marked as `/* Whitening */` for consistency

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+    addi a5, a5, 1024
+    addi a5, a5, 1536
+    jal  x1, sign_w0_w1_ctilde
+
+    la   a0, c_poly                    /* ntt_c output */
+    la   a1, sign_tmp                /* aligned c~ stash */
+    jal  x1, sign_c
+
+    /* sign_z_check params: ntt_c, rho_prime(gamma1),
+     * z_staging, tmp_poly, cs1_share1, seca2b_scratch.  s1 is now derived
```
pqcfox: [possible nit] `seca2b_scratch` appears to be at `sig + 2560`, but `sign_hint_b2a` is provided as the last param below; any chance this comment might be out of sync?

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+          bn.sid t0, 0(t1++)
+        endloop
+        nop
+    endloop
+    .endr
+
+    /* Per-column constants. */
+    li s6, W0_SHARE_STRIDE           /* stride between w shares */
+    bn.xor w23, w23, w23             /* matrix nonce = byte(i) || byte(j) */
+    li s5, 31                        /* w31 ref (unused but historical) */
```
pqcfox: [nit] (worth making a small issue for this?)

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+
+    la   t0, mldsa_params
+    lw   t0, MLDSA_PARAM_L_OFFSET(t0)
+    loop t0, 94
+        addi a0, s3, 0
+        addi a1, s9, 0
+        addi a2, s11, 0
+        addi a3, s2, 0
+        addi a4, s7, 0
+        /* gamma_1 dispatches on a5 (2 => POLYZ_BITS 18 for ML-DSA-44, else 20);
```
pqcfox: [nit] maybe intended to be `(2 => POLYZ_BITS = 18 for MLDSA-44, else 20)`?

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+    la   t3, mldsa_params
+    lw   t3, MLDSA_PARAM_K_OFFSET(t3)
+    li   t4, 4
+    beq  t3, t4, _sign_pack_ctilde_44
+    li   t4, 6
+    beq  t3, t4, _sign_pack_ctilde_65
+    /* ML-DSA-87 (K=8, CTILDEBYTES=64). */
+    bn.sid  t1, 0(t0)
+    bn.sid  t1, 0(s3)
+    bn.wsrr w8, 0xA
```
pqcfox: [nit] could replace `0xA` with named constant for consistency (ditto once above and once below)

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+ */
+.global sign_w0_w1_ctilde
+.type sign_w0_w1_ctilde, @function
+sign_w0_w1_ctilde:
+    /* Park pointer args.  s7 holds mu just long enough for the
+     * keccak_send_message; the K-loop later overwrites it with NSHARES. */
+    addi s0, a0, 0            /* w polyvec walker */
+    addi s7, a1, 0            /* mu (consumed below) */
+    addi s1, a2, 0            /* sign_w1_repvec walker */
+    addi s3, a3, 0            /* sig (for c~ writes) */
```
pqcfox, line 523: [super nit] could be useful to document that a copy of `c~` appears (it seems like?) in `sign_tmp` too as an intended side effect

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1312 @@
...
+    /* Refresh both shares in place with fresh urnd (preserves the XOR). */
+    addi t0, a1, 0
+    add  t1, a1, t4
+    li   t2, 0
+    loop t3, 7
+        bn.wsrr w2, urnd
+        bn.lid  t2, 0(t0)
+        bn.xor  w0, w0, w2
+        bn.sid  t2, 0(t0++)
+        bn.lid  t2, 0(t1)
```
pqcfox, line 731: [possible nit] would whitening between share accesses here make sense, in line with remasking of K shares at line 137?

sw/acc/crypto/mldsa/gen_twiddles.s
mkannwischer: It's stores 4 bytes of zeros to match the padding that was in the original table.
Technically we could drop it because it's unused in the NTT anyway. But dropping it breaks the test checking the gen_twiddles_fwd produces the same table as we had before.
I prefer to keep it for now.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Probably not strictly needed due to the urnd, but let's add it anyway.

sw/acc/crypto/mldsa/mldsa_keypair.s
mkannwischer: correct. deleted.

sw/acc/crypto/mldsa/gen_twiddles.s
mkannwischer: Indeed. Added. Apologies for the laziness.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: This isn't whitening.
We are computing `SHAKE256(K || rnd || mu)` where K is masked, but rnd and mu are unmasked. so we need to set the second share of rnd and mu to zero. 
I've added a comment saying /* share1 = 0 */ to each of them.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Sorry, the memory layout has changed too many times.
Let's comment it at the callee - that's easier to keep in sync. 
Removed the comment.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: I just dropped it.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Fixed.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Good catch. replaced.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Yep, because the next function needs it 32-byte aligned.
I have added an explanation to the function description.

sw/acc/crypto/mldsa/mldsa_sign.s
```diff
@@ -107,6 +107,1339 @@
...
+    la   s10, sign_tmp
+
+    /* Initialize the coefficient sum for the hint for post-check. */
+    li  s4, 0
+
+    /* Hint loop counter. */
+    la  s8, mldsa_params
+    lw  s8, MLDSA_PARAM_K_OFFSET(s8)
+
+    /* Normalize w0 to the [0, q) range (in-place).  The masked path
```
pqcfox: [nit] code this comment refers to appears to be missing now

sw/acc/crypto/mldsa/mldsa_sign.s
pqcfox: Thank you! Apologies for misunderstanding there.

pqcfox (approved):
Looks fantastic! Thank you for your patience with the long review process.

sw/acc/crypto/mldsa/mldsa_sign.s
mkannwischer: Good catch. This is outdated. Removed.

PR #347, opened 2026-08-05
[acc,util] Fix parsing of a comment before a mnemonic

### LINKED ISSUE #345: [acc,util] acc_as misparses an inline comment before a mnemonic
A `/* ... */` comment before the mnemonic makes `acc_as` resume parsing at the wrong column, so the rest of the line is misread.
Trailing comments are fine.

```asm
/* spill */ sw s8, 24(sp)
```

Fails with `Cannot match syntax for 'sw' ('<grs2>, <offset>(<grs1>)')`, though the syntax is valid.

```asm
/* spill s8 */ sw s8, 24(sp)
```

Crashes with `IndexError: string index out of range` in `_eat_ws`.

```asm
/* bit 0 */ .fill 32, 1, 0xff
```

No diagnostic at all; emits `.fillf` and exits 0.

Which of these you get depends on the length of the comment relative to the statement it precedes; a short one lands inside the operands, a longer one runs off the end of the line.
Indentation makes no difference.

## Reproduction

```
$ printf '\n/* spill */ sw s8, 24(sp)\n' >> sw/acc/code-snippets/loop.s
$ ./bazelisk.sh build //sw/acc/code-snippets:loop
...
/tmp/tmphdvzcftb.acc-gcc/0.s:43: Cannot match syntax for 'sw' ('<grs2>, <offset>(<grs1>)').
```

## Workaround

Put the comment on its own line, or after the mnemonic.

mkannwischer:
acc_as deleted block comments from a line before parsing it, but then carried on parsing at an offset computed against the original line. A comment to the left of the mnemonic shifted the operands out from under the parser, giving either a bogus syntax error, an IndexError, or silently dropped operands on a data directive.

Blank the comments out in place instead, so that the offset stays valid.

- Fixes #345.

### DIFF
--- a/hw/ip/acc/util/acc_as.py
+++ b/hw/ip/acc/util/acc_as.py
@@ -974,7 +974,7 @@ class Transformer:
             pos += end
 
         # Return if at EOL
-        if pos == len(line):
+        if pos >= len(line):
             return pos
 
         # Spot a line comment ('#'). In that case, eat to EOL and return.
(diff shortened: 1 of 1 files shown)


pqcfox (approved):
Looks excellent, thank you so much for getting this!

PR #348, opened 2026-08-05
[acc,iflow] Speed-up instruction counting to support masked PQC

mkannwischer:
- Depends on #288

### DIFF
--- a/hw/ip/acc/util/shared/information_flow.py
+++ b/hw/ip/acc/util/shared/information_flow.py
@@ -797,12 +797,23 @@ class InsnIndirectWDRNode(InsnInformationFlowNode):
         return self.op.op_type.op_val_to_str(op_vals[self.op.name], None)
 
     def required_constants(self, op_vals: Dict[str, int], coarse_dmem: bool) -> Set[str]:
+        # In coarse mode (instruction-count analysis) the specific WDR touched
+        # by an indirect reference does not affect control flow, so do not
+        # require the index register to be a compile-time constant.
+        if coarse_dmem:
+            return set()
         return {self._get_gpr_name(op_vals)}
 
     def evaluate(self, op_vals: Dict[str, int],
                  constant_regs: Dict[str, int],
                  coarse_dmem: bool) -> List[InformationFlowNode]:
         gpr = self._get_gpr_name(op_vals)
+        if coarse_dmem and gpr not in constant_regs:
+            # The index is not statically known (e.g. a loop-carried WDR index
+            # in a variable-count loop). Fall back to a single conservative node
+            # standing in for the referenced WDR; the exact WDR is irrelevant to
+            # instruction counting.
+            return [InformationFlowNode('wref-{}'.format(gpr))]
         return [_eval_indirect_wdr(gpr, constant_regs)]
 
 
--- a/hw/ip/acc/util/shared/information_flow.py
+++ b/hw/ip/acc/util/shared/information_flow.py
@@ -168,6 +168,17 @@ class DmemInformationFlowNode(InformationFlowNode):
         return result
 
 
+def _copy_flow(
+    flow: Dict[InformationFlowNode, Set[InformationFlowNode]]
+) -> Dict[InformationFlowNode, Set[InformationFlowNode]]:
+    '''Copies a flow dictionary, sharing the nodes it refers to.
+
+    Nodes are immutable value objects, so only the dictionary and the source
+    sets need to be fresh.
+    '''
+    return {sink: set(sources) for sink, sources in flow.items()}
+
+
 class InformationFlowGraph:
     '''Represents an information flow graph.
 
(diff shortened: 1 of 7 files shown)


hw/ip/acc/util/shared/instruction_count_range.py
```diff
@@ -25,7 +25,8 @@ def _get_insn_count_range(
         program: ACCProgram, graph: ControlGraph, start_pc: int,
         stop_at: StopPoint, loop_iters: Optional[Dict[int, int]],
         thru_label: Optional[str],
-        exclude_labels: list[str]) -> Optional[Tuple[int, int]]:
+        exclude_labels: list[str],
+        memo: Optional[Dict] = None) -> Optional[Tuple[int, int]]:
```
pqcfox: [super nit] could give full type as 
```suggestion
        memo: Optional[Dict[Tuple[int, int, str], Tuple[int, int]]] = None) -> Optional[Tuple[int, int]]:
```
for consistency

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -90,6 +90,10 @@ def pretty(self, dmem_symbols: Dict[str, int]) -> List[str]:
     def __hash__(self) -> int:
         return hash(self.name)
 
+    def __deepcopy__(self, memo: Optional[Dict[int, Any]]) -> 'InformationFlowNode':
+        # Immutable value object; a deep copy can share the same instance.
+        return self
+
```
pqcfox: [super nit] may be helpful (possibly in a follow up PR) for us to explicitly comment in docstring below that mutating e.g. `DmemInformationFlowNode.{start,end}` (e.g. when simplifying information flows) could lead to spooky action at a distance (also the option of preventing direct setting via `__setattr__`, but this would be pretty unidiomatic for Python)

hw/ip/acc/util/shared/information_flow_analysis.py
```diff
@@ -70,7 +70,31 @@ class IFlowCache(Cache[int, ConstantContext, IFlowResult]):
...
     the cached result is returned.
     '''
-    pass
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._index: Dict[int, Dict[Tuple[str, ...], Dict[Tuple[int, ...],
+                                                          IFlowResult]]] = {}
+
+    def add(self, index: int, entry: CacheEntry) -> None:
```
pqcfox: [super nit] (similarly
```suggestion
    def add(self, index: int, entry: CacheEntry[ConstantContext, IFlowResult]) -> None:
```
could provide full type, though also verbose)

pqcfox (approved):
Looks fantastic! Left some very optional nits, feel free to go ahead and merge if these seem not useful.

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -790,12 +790,23 @@ def _get_gpr_name(self, op_vals: Dict[str, int]) -> str:
         return self.op.op_type.op_val_to_str(op_vals[self.op.name], None)
 
     def required_constants(self, op_vals: Dict[str, int], coarse_dmem: bool) -> Set[str]:
+        # In coarse mode (instruction-count analysis) the specific WDR touched
```
pqcfox, line 800: Actually, one possible edge case to consider: `bn.movr` followed by a comparison downstream of the destination WDR could cause either GPR to influence control flow. 

Right now `bn.movr` is only (I believe) used in the Montgomery arithmetic implementation for RSA, and there I don't believe (?) there's any such instance of this edge case, but it could be worth considering this possibility in any case.

hw/ip/acc/util/shared/information_flow_analysis.py
mkannwischer: Added

hw/ip/acc/util/shared/instruction_count_range.py
mkannwischer: Added

hw/ip/acc/util/shared/information_flow.py
mkannwischer: Thinking about this again, overwriting the deepcopy is a very hacky solution. Let's not do that. 
I'll look into alternatives.

mkannwischer: I've instead introduced a `_copy_flow` helper does the copying.

hw/ip/acc/util/shared/information_flow.py
mkannwischer: I agree that this is a soundness risk here. If you don't mind I would like to fix this in a follow up as I think this needs another round of review.

pqcfox: Makes sense to me; I agree this makes the most sense and the PR is a substantial improvement that's worth having in place right away.

PR #349, opened 2026-08-05
[setup] Add Python 3.13 support

### LINKED ISSUE #346: [setup] Python dependencies fail to install with Debian 13 (trixie, Python 3.13)
I'm trying to build Pavona (following the getting-started guide) on Debian 13, which, I fully understand, is not an officially supported platform.
I'm getting fairly far, but then the step ```pip3 install -r python-requirements.txt --require-hashes``` fails with error message
```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    pyyaml-ft>=8.0.0 from https://files.pythonhosted.org/packages/35/be/58a4dcae8854f2fdca9b28d9495298fd5571a50d8430b1c3033ec95d2d0e/pyyaml_ft-8.0.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (from libcst==1.8.6->-r python-requirements.txt (line 226))
```
If this is not much effort to fix, it would be great to have support for the latest Debian stable distribution.

mkannwischer:
`pyproject.toml` claims support for Python 3.10 through 3.13, but CI exclusively tests Python 3.10. In fact, the `python-requirements.txt` fail to install with Python 3.13 as observed by @cryptojedi in #346. Python 3.11 and Python 3.12 seem to work fine. 

For everyone only using Python through bazel this isn't an issue because MODULE.bazel is pinning Python 3.10. However, there is still a number of tools (e.g., dvsim) that are not executed through bazel and for which we have to rely on the system Python interpreter. Hence, we should make sure these tools work with the Python version shipped by the default on the supported platforms, i.e., Ubuntu 22.04 (Python 3.10), Ubuntu 24.04 (Python 3.12), and Ubuntu 26.04 (Python 3.14).

This PR works towards that and adds support for Python 3.13 building upon #366. It also adds setup tests that install `python-requirements.txt` using Python 3.{10,11,12,13} and run a few selected tools. 

Python 3.14 support will be added in a follow-up PR.

- Resolves #346


fragglet (approved):
Nice work!

PR #351, opened 2026-08-05
[verilator] Link only the C++ runtime statically to fix build on Debian 13

### LINKED ISSUE #350: [verilator] Simulation fails to link on Debian 13
On Debian 13 (trixie, elfutils 0.192-4) the Verilator simulation does not link:

    $ ./bazelisk.sh test --test_output=errors \
        //sw/device/tests:uart_smoketest_sim_verilator
    ERROR: hw/BUILD:93:14: FuseSoC .../Vchip_sim_tb [for tool] failed: (Exit 1)

The error is in `bazel-out/_tmp/actions/stderr-*`:

    ld.lld: error: undefined symbol: eu_search_tree_init
    >>> referenced by elf_begin.o:(file_read_elf) in archive /lib/x86_64-linux-gnu/libelf.a

Potentially related: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1102039

Debian is not a supported platform, but other platforms may run into the same
thing as they pick up newer elfutils, and it looks easy enough to fix by not
linking libelf statically.

mkannwischer:
Debian isn't a supported OS, but having at least partial support seems worthwhile regardless and the [actual fix](https://github.com/pavona/pavona/pull/351/commits/b064d8ea5435322f8dedfc2d5b7484da5c314182) is very small. 

- Fixes #350 

Also adds a CI smoketest on Debian 12 and 13.

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -82,7 +82,15 @@ runs:
     - name: Install UTF-8 language pack
       run: |
         if command -v apt > /dev/null; then
-          sudo apt install -y language-pack-gnome-en-base
+          # Only Ubuntu ships the language-pack-* packages. On Debian the
+          # locale has to be generated from the `locales` package instead.
+          if apt-cache show language-pack-gnome-en-base > /dev/null 2>&1; then
+            sudo apt install -y language-pack-gnome-en-base
+          else
+            sudo apt install -y locales
+            echo "en_US.UTF-8 UTF-8" | sudo tee -a /etc/locale.gen > /dev/null
+            sudo locale-gen
+          fi
           echo "LANG=en_US.UTF-8" >> "$GITHUB_ENV"
           echo "LANGUAGE=en.UTF-8" >> "$GITHUB_ENV"
         fi
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -129,8 +129,8 @@ runs:
         VERIBLE_URL="https://github.com/chipsalliance/verible/releases/download/${{ inputs.verible-version }}/${VERIBLE_TAR}"
         source ci/scripts/utils.sh
         sudo mkdir -p "${{ inputs.verible-path }}"
-        retry 5 2 2 curl -sSfL "$VERIBLE_URL" -o "${{ runner.temp }}/${VERIBLE_TAR}"
-        sudo tar -C "${{ inputs.verible-path }}" -xvzf "${{ runner.temp }}/${VERIBLE_TAR}" --strip-components=1
+        retry 5 2 2 curl -sSfL "$VERIBLE_URL" -o "$RUNNER_TEMP/${VERIBLE_TAR}"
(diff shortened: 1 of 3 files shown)


mkannwischer:
I got slightly confused why the installation of the `python-requirements.txt` passes even though Debian 13 defaults to Python 3.13 which is incompatible with our current pinned dependencies. The reason is that CI actually uses Python 3.10 installed via `uv`.

mkannwischer:
I've briefly tried to also add an AlmaLinux setup test, but it fails with:
```
ERROR: /__w/pavona/pavona/sw/device/lib/testing/test_rom/BUILD:35:14: Action sw/device/lib/testing/test_rom/test_rom_sim_dv.32.vmem failed: (Exit 1): srec_cat failed: error executing Action command (from target //sw/device/lib/testing/test_rom:test_rom) srec_cat bazel-out/k8-fastbuild-ST-0276d9ae5970/bin/sw/device/lib/testing/test_rom/test_rom_sim_dv.bin --binary --offset 0x0 --byte-swap 4 --fill 0xff -within ... (remaining 8 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
src/main/tools/process-wrapper-legacy.cc:80: "execvp(srec_cat, ...)": No such file or directory
Target //sw/device/examples/hello_world:hello_world_sim_dv failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 494.020s, Critical Path: 57.93s
INFO: 1618 processes: 615 internal, 1003 processwrapper-sandbox.
ERROR: Build did NOT complete successfully
```
see [this log](https://github.com/pavona/pavona/actions/runs/31139337137/job/92745700157?pr=351). Might be worth fixing in a follow-up PR.

fragglet (approved):

.github/actions/prepare-env/action.yml
```diff
@@ -121,8 +129,8 @@ runs:
         VERIBLE_URL="https://github.com/chipsalliance/verible/releases/download/${{ inputs.verible-version }}/${VERIBLE_TAR}"
         source ci/scripts/utils.sh
         sudo mkdir -p "${{ inputs.verible-path }}"
-        retry 5 2 2 curl -sSfL "$VERIBLE_URL" -o "${{ runner.temp }}/${VERIBLE_TAR}"
```
rtorok-zr, line 124: Did using `${{ runner.temp }}` break something? From my read, `${{ runner.temp }}` and `$RUNNER_TEMP` should behave identically.

rtorok-zr (approved):
This seems fine. There are some portability concerns about changing verilator to use some dynamic libraries, but it should be okay for the platforms we currently support.

.github/actions/prepare-env/action.yml
mkannwischer: Yeah this broke for the setup for jobs using 
```
    runs-on: XXX
    container:
      image: YYY
```

I tried to explain it in the commit msg:
```
The next commit adds a job that runs inside a container. There the
runner.temp context expands to the temporary directory's path on the
host, which is not mounted in the container, so downloading into it
fails. The variable holds the path the directory is mounted at and is
correct in both cases.
```
When not using a container they behave identically, I think.

PR #352, opened 2026-08-05
[acc,mlkem] Add masked ACC ML-KEM cryptolib integration

mkannwischer:
Wire the cryptolib ML-KEM backend to the hardened run_mlkem_hardened ACC
binary. Key generation and decapsulation operate on Boolean-masked key
material, with the masked decapsulation key stored in the ACC key
layout. The derandomized keygen API takes a randomness buffer twice as
long, holding two shares of the seed coins.

Also enable the ACVP and Wycheproof tests (run in CI) for the masked implementation.

- Depends on #348 
- Depends on #340
- Depends on #288


sw/device/lib/crypto/impl/mlkem.c
```diff
@@ -71,35 +79,48 @@ otcrypto_status_t otcrypto_mlkem512_keygen_derand(
...
+#else
+  if (launder32(secret_key->config.security_level) !=
+      kOtcryptoKeySecurityLevelPassiveRemote) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  HARDENED_CHECK_EQ(secret_key->config.security_level,
+                    kOtcryptoKeySecurityLevelPassiveRemote);
+#endif
+
+#if defined(ACC_HAS_PQC) && defined(ACC_MLKEM_HARDENED)
```
pqcfox, line 104: [super nit] for later, could be worth standardizing to one of `defined(ACC_HAS_PQC) && defined(ACC_MLKEM_HARDENED)` or just `defined(ACC_MLKEM_HARDENED)` (e.g. above), ditto `ifdef` and `if defined(...)` for cases where boolean operators aren't needed

sw/device/lib/crypto/impl/mlkem.c
```diff
@@ -71,35 +79,48 @@ otcrypto_status_t otcrypto_mlkem512_keygen_derand(
...
   (void)work;
-  HARDENED_TRY(
-      mlkem_acc_512_keygen(randomness.data, public_key->key, sk_share0));
+  const uint32_t *rand_share0 = randomness.data;
```
pqcfox, line 106: [super nit] for later (something I should do), may be nice to push the `randomness.data` splitting into shares down into the internal functions as with `secret_key`

sw/device/lib/crypto/impl/mlkem.c
```diff
@@ -244,37 +264,57 @@ otcrypto_status_t otcrypto_mlkem512_decapsulate(
...
-  // Secret keys are stored unmasked in share0 (share1 is zero).
-  uint32_t *sk_share0;
-  uint32_t *sk_share1;
-  HARDENED_TRY(keyblob_to_shares(secret_key, &sk_share0, &sk_share1));
+  HARDENED_TRY(check_keyblob_length(secret_key));
 
   // Write the unmasked shared secret into the first share of the keyblob.
   uint32_t *ss_share0;
   uint32_t *ss_share1;
   HARDENED_TRY(keyblob_to_shares(shared_secret, &ss_share0, &ss_share1));
```
pqcfox, line 284: [super nit] (possibly ditto for shared secret here)

pqcfox (approved):

sw/device/lib/crypto/impl/mlkem.c
mkannwischer: Agreed, we should standardize this. I'll handle this in a follow-up PR.

sw/device/lib/crypto/impl/mlkem.c
mkannwischer: Sounds good.

PR #357, opened 2026-08-05
[env] Clean up hard coded bash and sh paths to be more portable

ngildenhuys:
This PR changes all scripts and spots where scripts are generated to not use `/bin/bash` system paths and instead default to the environment's bash interpreter by use of `/usr/bin/env bash`. This is also done for sh invocation, except for Makefiles. In those cases of setting the `SHELL` we default to `/bin/sh` instead of `/bin/bash` since it is generally more portable. Invocations of direct paths are changed to just `bash` invocations. This makes the project much less reliant on the system provided utilities existing in a set place.


qmn:
recheck

signing/logs/2023-06-06.md
fragglet, line 1: Unrelated change?

PR #359, opened 2026-08-06
[acc,rtl] Fixed FIFO deadlock constraints to resolve ML-DSA failures

apinise:
The scenarios for FIFO deadlock were too strict based on a partial write and pending message share being written into the FIFO. The first deadlock conditions still cover this scenario where a pending write is attempted, but one FIFO share is empty/ready, and the other has >8 Bytes in it. This should unblock any issues with the ACC triggering a fatal alert and secure wipe while executing ML-DSA / ML-KEM code.


mkannwischer (approved):
Thanks @apinise. LGTM.

PR #361, opened 2026-08-07
[doc] Move the Python setup out of the getting started guide

mkannwischer:
The guide sets up a virtual environment and installs the requirements before anything else, but Bazel provides its own interpreter for the tools it runs and needs neither. Only the tools run outside it do, and several pages install their packages into the system interpreter with `pip3 install --user`, which is not the environment the guide creates.

Give the setup its own page next to the other setup guides, and link it from the tools that need it.

Context: Our current `python-requirements.txt` is incompatible with anything newer than Python 3.13 (Ubuntu 26 ships 3.14; Debian 13 ships 3.13). Requiring all users to install it even though only a subset of tools will need it feels unnecessary. In the medium term, it would be good to make it compatible (most changes for that seem to be needed for bumping the Python version used within bazel anyway). In the long term, it would be good if we do not rely on the system Python interpreter.

### DIFF
--- a/util/reggen/doc/setup_and_use.md
+++ b/util/reggen/doc/setup_and_use.md
@@ -6,14 +6,7 @@ The example commands assume `$REPO_TOP` is set to the toplevel directory of the 
 
 ### Setup
 
-If packages have not previously been installed you will need to set a few things up.
-First use `pip3` to install some required packages:
-
-```console
-$ pip3 install --user hjson
-$ pip3 install --user mistletoe
-$ pip3 install --user mako
-```
+`regtool.py` runs outside Bazel, so it needs the environment described in [Python Environment Setup](../../../doc/getting_started/setup_python.md).
 
 ### Register JSON Format
 
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -9,6 +9,7 @@
 
 # Using Pavona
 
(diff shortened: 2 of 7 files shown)


adurbin-zr (approved):

fragglet (approved):

util/reggen/doc/setup_and_use.md
```diff
@@ -6,14 +6,7 @@ The example commands assume `$REPO_TOP` is set to the toplevel directory of the
...
 
-If packages have not previously been installed you will need to set a few things up.
-First use `pip3` to install some required packages:
-
-```console
-$ pip3 install --user hjson
-$ pip3 install --user mistletoe
-$ pip3 install --user mako
-```
+`regtool.py` runs outside Bazel, so it needs the environment described in [Python Environment Setup](../../../doc/getting_started/setup_python.md).
```
matutem, line 9: This is okay, but fwi, we also run regtool via bazel: see rules/autogen.bzl for example.

mkannwischer: Thanks! I've never used regtool, but the pip instructions here definitely looked wrong.

Do you mind opening a follow-up PR updating all the regtool docs to bazel? Or nominate someone else who could do it? I see 166 mentions of regtool.py and it looks to me like most/all of them need to change.

cc @adurbin-zr @fragglet

PR #362, opened 2026-08-07
[env] fixing undeclared dependencies in pyproject toml

ngildenhuys:
These changed packages don't declare setuptools as a dependency. So in a heavily sandboxed environment when building these packages, the build fails to find the `setuptools` module.


mkannwischer (changes requested):
CI seems to be using an old uv (0.7.14): https://github.com/pavona/pavona/blob/df8acd07e7056232bb5e1f59cd7c651c795400fe/.github/actions/prepare-env/action.yml#L93

`extra-build-dependencies` was introduced in 0.8.4 https://github.com/astral-sh/uv/releases/tag/0.8.4. 

Could you bump uv to a recent version, please?

PR #366, opened 2026-08-11
[bazel] Bump `rules_python` to 2.2.0

mkannwischer:
Context (not all done in this PR): Our `python-requirements.txt` fails to install with newer Python versions (e.g., Python 3.13). Ideally, we would be running everything though bazel allowing us to pin the used Python version, but today there are still a number of tools that rely on the system Python interpreter (e.g, dvsim). While there are essential tools run outside of Bazel, I think we need to at least support the default Python versions shipped by the supported operating systems (Ubuntu 22/24/26), i.e., Python 3.10 - Python 3.14. To do so, we'll need to switch to universal lock files that pin dependencies for each Python version (and fix a few other things). However, we currently pin Bazel's `rules_python` 1.2.0 in MODULE.bazel which doesn't support universal lock files yet.

This PR updates `rules_python` to the latest version and makes the necessary changes. The biggest change is that `rules_python` now runs targets in a virtual environment, so every missing dependency declaration bites. I think declaring all dependencies is a good idea even independent of the update.


fragglet (approved):
Only thing that seems a bit ugly is the `imports = [".."]` pattern. Would it be better to give absolute paths, eg. `imports = ["//util"]` and equivalent?

Just a question; the PR looks fine as it is and happy to merge either way.

mkannwischer:
> Only thing that seems a bit ugly is the `imports = [".."]` pattern. Would it be better to give absolute paths, eg. `imports = ["//util"]` and equivalent?
> 
> Just a question; the PR looks fine as it is and happy to merge either way.

I tried this just now, but it does not work. Imports are always relative in `py_library`.
`imports = ["../XXX"]` is also a pre-existing pattern in many places.

I agree that it is rather ugly and it makes it very hard to track which files are actually imported.
I think a good way to fix this is to drop all imports in `py_library` and instead use fully qualified imports in the Python scripts.

For example,
```py
py_binary(
    name = "mldsa_sign_testgen",
    srcs = ["mldsa_sign_testgen.py"],
    imports = [
        "../../../../../hw/ip/acc/util",
    ],
    deps = [
        "//hw/ip/acc/util/shared:testgen",
        requirement("dilithium-py"),
    ],
)
```

would become
```py
py_binary(
    name = "mldsa_sign_testgen",
    srcs = ["mldsa_sign_testgen.py"],
    deps = [
        "//hw/ip/acc/util/shared:testgen",
        requirement("dilithium-py"),
    ],
)
```

and the import would change from 
```py
from shared.testgen import write_test_data, write_test_exp, write_test_dexp
```
to 
```py
from hw.ip.acc.util.shared.testgen import write_test_data, write_test_exp, write_test_dexp
```

But of course this breaks when invoking scripts directly (the example above isn't invoked directly, but others are). We could fix that by always putting the repo root on the path.

If you agree with this, then I can prepare a follow-up PR.

PR #368, opened 2026-08-11
[dvsim] Count license timeout separately

LouisTheLuis:
This PR contains two commits:
- The first one aims to make `dvsim` count the beginning of a run from polling the `run.log` and waiting for certain patterns that might indicate that the run is underway (for e.g. a `UVM_INFO`) , if applicable. This is because originally, `dvsim` would count the `run_timeout` from the moment the `make` file launches `dvsim`, which meant that if the process launched but it's waiting on a license, it might run into a timeout before it even started simulating. It also includes a `dvsim` argument for adjusting the timeout window for license queuing. 
- The second one simply differentiates the failures; a DV timeout should present a different error message from a license timeout.

### DIFF
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -120,6 +120,7 @@ class Deploy():
             "dry_run": False,
             "exports": False,
             "flow_makefile": False,
+            "license_error_patterns": False,
             "name": False,
         }
 
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -40,6 +40,15 @@ class Deploy():
     # TODO: Allow these to be set in the HJson.
     weight = 1
 
+    # Patterns whose first appearance in the job's log marks where it stopped
+    # waiting for a license and started work. The timeout is measured from there.
+    # Empty means nothing to wait for, so the job is timed from its launch.
+    #
+    # The marker depends on the flow and the tool, so the targets that have one
+    # take it from the HJson, alongside their pass and fail patterns. This is the
+    # fallback for the targets that have none.
+    started_patterns = []
+
     def __str__(self):
         return (pprint.pformat(self.__dict__)
                 if log.getLogger().isEnabledFor(VERBOSE) else self.full_name)
@@ -372,6 +381,7 @@ class CompileSim(Deploy):
         self.mandatory_misc_attrs.update({
             "build_fail_patterns": False,
             "build_pass_patterns": False,
+            "build_started_patterns": False,
             "build_timeout_mins": False,
             "cov_db_dir": False,
         })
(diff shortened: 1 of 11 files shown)


ysekhar (approved):
Lets see how this pans out! 
LGTM

util/dvsim/Deploy.py
```diff
@@ -40,6 +44,26 @@ class Deploy():
...
+    # charged against it. An empty list means there is nothing to wait for and
+    # the job is timed from the moment it was launched.
+    started_patterns = []
+
+    # Patterns saying the job died because of the license server rather than
+    # because of anything in the design or the testbench. A failing job whose log
+    # matches one of these is reported as killed instead, so that a license
+    # outage does not read as a DV failure.
+    #
+    # These are the messages VCS prints when the server rereads its license file
```
matutem: Please add some comment indicating additional patterns may be needed for different flows and tools. In fact, perhaps we should make this a map indexed by tool and flow?

util/dvsim/Deploy.py
```diff
@@ -482,6 +506,10 @@ class RunTest(Deploy):
     target = "run"
     seeds = []
     fixed_seed = None
+
+    # Any UVM report means the simulator has its license, has loaded the design
+    # and is away.
+    started_patterns = [r"^UVM_(INFO|WARNING|ERROR|FATAL)"]
```
matutem: This may also depend on the flow, like lint and formal.

util/dvsim/Launcher.py
```diff
@@ -237,11 +249,61 @@ def kill(self) -> None:
...
+        if not patterns:
+            return None
+
+        try:
+            with open(
+                self.deploy.get_log_path(),
+                encoding="UTF-8",
+                errors="surrogateescape",
+            ) as f:
+                lines = f.readlines()
```
matutem: This will read the full file into an array of strings which could be huge, just to scan the first few lines. It is much better to iterate over the file object to get the lines on demand. Try to move the logic around line 291 into a function and call it on an iterator here.

matutem (commented):
Could you deal with the comments?

util/dvsim/Launcher.py
LouisTheLuis: Moved into a helper function. Fixed.

util/dvsim/Deploy.py
LouisTheLuis: You are right... I moved the patterns to the corresponding tools' hjsons. That way it is not hardcoded in the Python.

util/dvsim/Deploy.py
LouisTheLuis: With the move to the tools' hjson, this should be clear (also explained in the comments). Unsupported flows just pass an empty list of patterns.

matutem:
You are still missing the lint hjsons for the new patterns and settings. These can be found under hw/lint/tools/dvsim/<tool>.hjson. I think veriblelint and verilator are license free. ascentlint is licensed and we used it in older times, so we can postpone supporting the license flow for it. Since the additional flows we actually use are veriblelint and perhaps verilator lint, and both are license-free we may not need changes, unless you add a comment saying they don't need these patterns.

For future reference, the "formal", "syn", "cdc", and "rdc" flows have some dedicated hjson under hw/<flow>/tools/dvsim, but we are not currently using these tools.

matutem (approved):
Adding some comment in some lint files is not needed at this point. My last comment was informational.

LGTM

PR #369, opened 2026-08-12
[acc,dv] Run the ACC smoke tests as hermetic bazel tests

mkannwischer:
Thus far the ACC smoke tests spawn the ISS with the system
Python interpreter requiring the user to install Python
dependencies locally.
This feels unnecessary as we might as well run the ISS through
bazel.

Add a `py_binary` ISS (`stepped.py`) to use Bazel's pinned Python
environment with the pinned dependencies, and let `ACC_ISS` name an
ISS executable to run instead of the model in the source tree.
Both smoke tests become `sh_test`s that pass the launcher,
the Verilated model and the ELF from their runfiles, so neither
the interpreter nor the repository path is taken from the environment.

Keep the fallback so other tools (e.g., dvsim) do not break:
If ACC_ISS is unset, the iss_wrapper stays unchanged.

The scripts also avoid `readlink -e` and `timeout`, neither of which
macOS provides. The wrappers only need to reach the repository root,
and Bazel already bounds the runtime of the `sh_test`.

Both tests can be run with:
```
  ./bazelisk.sh test --test_output=errors //hw/ip/acc/dv/smoke:run_smoke_test //hw/ip/acc/dv/smoke_pqc:run_smoke_test
```

### DIFF
--- a/hw/ip/acc/util/run_smoke_test.sh
+++ b/hw/ip/acc/util/run_smoke_test.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+# Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Runs an ACC smoke test (runs the simulation and checks expected output). The
+# wrapping sh_test passes the paths of everything this needs, relative to the
+# runfiles directory.
+
+fail() {
+    echo >&2 "ACC SMOKE FAILURE: $*"
+    exit 1
+}
+
+set -euo pipefail
+
+SIM_BINARY=$1
+SMOKE_ELF=$2
+SMOKE_EXPECTED=$3
+# Keep the ISS path inside the runfiles tree, so that the launcher finds the
+# Python libraries it was built with.
+export ACC_ISS=$PWD/$4
+
+RUN_LOG=$TEST_TMPDIR/smoke.log
+
+if ! "$SIM_BINARY" --load-elf="$SMOKE_ELF" -t | tee "$RUN_LOG"; then
+  fail "Simulator run failed"
+fi
+
+# The expected output is the "Call Stack:" line and everything after it.
+expected_lines=$(wc -l < "$SMOKE_EXPECTED")
+if grep -A "$((expected_lines - 1))" "Call Stack:" "$RUN_LOG" \
+  | diff -U3 "$SMOKE_EXPECTED" -; then
+  echo "ACC SMOKE PASS"
+else
+  fail "Simulator output does not match expected output"
+fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,9 +251,9 @@ jobs:
       - name: ACC ISS test
         run: make -C hw/ip/acc/dv/accsim test
       - name: ACC smoke test
-        run: ./hw/ip/acc/dv/smoke/run_smoke.sh
+        run: ./bazelisk.sh test --test_output=errors //hw/ip/acc/dv/smoke:run_smoke_test
       - name: ACC PQC smoke test
-        run: ./hw/ip/acc/dv/smoke_pqc/run_smoke.sh
+        run: ./bazelisk.sh test --test_output=errors //hw/ip/acc/dv/smoke_pqc:run_smoke_test
       - name: Assemble & link code snippets
         run: make -C hw/ip/acc/util asm-check
 
--- a/hw/ip/acc/doc/developing_acc.md
+++ b/hw/ip/acc/doc/developing_acc.md
@@ -87,13 +87,19 @@ The generated binaries, a Verilated model and the output from running them can a
 ### Run the smoke test
 
 A smoke test which exercises some functionality of ACC can be found, together with its expected outputs (in the form of final register values), in `./hw/ip/acc/dv/smoke`.
-The test can be run using a script.
+The test can be run with Bazel.
 
 ```sh
-hw/ip/acc/dv/smoke/run_smoke.sh
+./bazelisk.sh test --test_output=errors //hw/ip/acc/dv/smoke:run_smoke_test
 ```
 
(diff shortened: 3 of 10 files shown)


mkannwischer:
I was unsure if we need to keep the `run_smoke.sh` scripts. They are now thin wrappers around the Bazel test and are not strictly needed. I went for keeping them to make this change transparent to developers.

hw/ip/acc/util/run_smoke_test.sh
```diff
@@ -0,0 +1,42 @@
...
+
+status=0
+"$SIM_BINARY" --load-elf="$SMOKE_ELF" -t | tee "$RUN_LOG" || status=$?
+
+if [ $status -ne 0 ]; then
+  fail "Simulator run failed"
+fi
+
+had_diff=0
+grep -A 74 "Call Stack:" "$RUN_LOG" | diff -U3 "$SMOKE_EXPECTED" - || had_diff=1
```
vlpannel: Nit: while I don't expect this changes much, it might be useful to know the number of lines in the expected file instead of hard-coding 74. (Something like `cat "$SMOKE_EXPECTED" | wc -l` should yield 75 lines in this case.)

hw/ip/acc/util/run_smoke_test.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
vlpannel:
```suggestion
#!/usr/bin/env bash
```

vlpannel (approved):
LGTM with just a few nits, though I would leave @apinise to give the final approval. Ran the tests locally, and they passed.

hw/ip/acc/util/run_smoke_test.sh
mkannwischer: Fixed.

hw/ip/acc/util/run_smoke_test.sh
mkannwischer: Fixed.

apinise (approved):
LGTM! I think leaving the `run_smoke.sh` files in are fine for the reason you said. I also confirmed that ACC dvsim regressions were unaffected for both variants.

qmn:
Two quick drive-by remarks:
* May as well update .github/workflows/ci.yml to invoke the ACC smoke tests through Bazel rather than the thinly wrapped scripts, too.
* There used to be a timeout of 5 seconds in the run_smoke.sh and run_pqc_smoke.sh scripts that, in my experience, triggered the ACC smoke timeouts when run on CI. The `timeout 5s` call is absent from the new run_smoke_test.sh, and I think this is fine going forward.

hw/ip/acc/util/run_smoke_test.sh
```diff
@@ -0,0 +1,45 @@
...
+  fail "Simulator run failed"
+fi
+
+had_diff=0
+# The expected output is the "Call Stack:" line and everything after it.
+expected_lines=$(wc -l < "$SMOKE_EXPECTED")
+grep -A "$((expected_lines - 1))" "Call Stack:" "$RUN_LOG" \
+  | diff -U3 "$SMOKE_EXPECTED" - || had_diff=1
+
+if [ $had_diff == 0 ]; then
```
fragglet: Can you eliminate the `$had_diff` variable entirely?
```suggestion
if grep -A "$((expected_lines - 1))" "Call Stack:" "$RUN_LOG" \
  | diff -U3 "$SMOKE_EXPECTED" -; then
```

hw/ip/acc/util/run_smoke_test.sh
```diff
@@ -0,0 +1,45 @@
...
+# Keep the ISS path inside the runfiles tree, so that the launcher finds the
+# Python libraries it was built with.
+export ACC_ISS=$PWD/$4
+
+RUN_LOG=$TEST_TMPDIR/smoke.log
+
+status=0
+"$SIM_BINARY" --load-elf="$SMOKE_ELF" -t | tee "$RUN_LOG" || status=$?
+
+if [ $status -ne 0 ]; then
```
fragglet: I think you can eliminate `$status`? ie.
```suggestion
if ! "$SIM_BINARY" --load-elf="$SMOKE_ELF" -t | tee "$RUN_LOG"; then
```

hw/ip/acc/util/run_smoke_test.sh
```diff
@@ -0,0 +1,45 @@
...
+# wrapping sh_test passes the paths of everything this needs, relative to the
+# runfiles directory.
+
+fail() {
+    echo >&2 "ACC SMOKE FAILURE: $*"
+    exit 1
+}
+
+set -o pipefail
+set -e
```
fragglet: Combine and also set `-u`:
```suggestion
set -euo pipefail
```

fragglet (dismissed):

mkannwischer:
> Two quick drive-by remarks:
> 
> * May as well update .github/workflows/ci.yml to invoke the ACC smoke tests through Bazel rather than the thinly wrapped scripts, too.

Yes, good idea. Will do. 

> * There used to be a timeout of 5 seconds in the run_smoke.sh and run_pqc_smoke.sh scripts that, in my experience, triggered the ACC smoke timeouts when run on CI. The `timeout 5s` call is absent from the new run_smoke_test.sh, and I think this is fine going forward.

This just happened again in the nightly: https://github.com/pavona/pavona/actions/runs/33037009784/job/98402561434
Would be nice if that is accidentally fixed. In my local experiments, 5s was rather generous though.
I'll closely watch the nightly.

hw/ip/acc/util/run_smoke_test.sh
mkannwischer: Yes, fixed.

hw/ip/acc/util/run_smoke_test.sh
mkannwischer: Fixed.

hw/ip/acc/util/run_smoke_test.sh
mkannwischer: Fixed.

Issue #372, opened 2026-08-12
DMA Abort with an Accepted Bus Request


ysekhar:
## TL;DR

The DMA abort behavior is not defined precisely enough when a Host, CTN, or
System-bus request has already been accepted but its response has not returned.
The current RTL immediately returns the functional state machine to Idle,
clears `STATUS.busy`, and sets `STATUS.aborted`. It does not retain explicit
ownership of the accepted request before allowing a new DMA operation.

A standalone legacy-DV testcase now reproduces this on the Host TL path. It
holds the D response after an accepted A request, issues abort, and reads
`STATUS` while the response is still outstanding. The observed value has
`busy=0`, contrary to the required drain boundary. The testcase is based
directly on `main` and does not depend on reset-safe infrastructure.

That is unsafe. Abort can stop the requested copy or hash operation, but it
cannot cancel a bus request that another component has already accepted. The
old response must still be consumed and kept separate from any later operation.
Otherwise the DMA can exceed its one-outstanding-request design, associate an
old response with a new transfer, or leave interconnect state that can prevent
forward progress.

The proposed contract is:

1. Abort stops new DMA requests immediately.
2. A request that has not been accepted may be withdrawn.
3. A request that has been accepted must be drained.
4. Its response is consumed but has no functional effect on the aborted copy or
   hash operation.
5. A new operation cannot reuse that interface until the old response drains.
6. `STATUS.aborted` is set only when this drain is complete.

This needs an RTL change for the Host and CTN TL-UL interfaces. A System-interface
change is also expected, but its exact implementation depends on resolving the
meaning of the System write grant. Reset must not be used as the normal recovery
mechanism.

## Issue summary

The DMA can report an abort as complete while a Host TL request accepted before
the abort still has no D response. The functional state machine forgets that
transport work and permits the interface to be reused, even though the DMA is
configured for only one outstanding request. An old response can therefore
overlap a later operation or be associated with the wrong operation.

The standalone `dma_abort_outstanding` testcase demonstrates the first
software-visible failure: `STATUS.busy` clears while the accepted Host response
is deliberately held pending. The register description says
`STATUS.aborted` is set once the aborted operation drains.

## Purpose of this document

This note records the observed behavior, the system risk, and the decisions
needed from architecture and RTL. Dragonfly is one integration of the DMA, but
block-level verification must cover every interface supported by the DMA IP.
Dragonfly's current System-interface tie-off therefore does not remove System
abort behavior from the DMA specification or verification scope.

[...]

PR #373, opened 2026-08-12
[dma,dv] Add directed abort drain reproducer

ysekhar:
This PR contains the test case and RTL fixes that showcases the bug highlighted in PR #372

### DIFF
--- a/hw/ip/dma/rtl/dma.sv
+++ b/hw/ip/dma/rtl/dma.sv
@@ -85,6 +85,9 @@ module dma
   logic dma_state_error;
   // SEC_CM: FSM.SPARSE
   dma_ctrl_state_e ctrl_state_q, ctrl_state_d;
+  logic abort_complete;
+  logic abort_rsp_pending, abort_rsp_complete;
+  dma_ctrl_state_e abort_wait_state;
   logic set_error_code, clear_go, clear_status, clear_sha_status, chunk_done;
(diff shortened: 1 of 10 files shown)


hw/ip/dma/rtl/dma.sv
```diff
@@ -636,13 +640,51 @@ module dma
     // Default assignments for the muxed config signals for the idle state
     cfg_handshake_en = control_q.cfg_handshake_en;
 
-    // Abort has the highest priority in the state machine. In all cases, if the abort is raised,
-    // the DMA is reset to the idle state. This includes the error state and the default state,
-    // which should never be reached during normal operation. The abort condition has precedence
-    // over any outstanding TL-UL transaction.
+    // Derive the transport state that must survive an abort. A response-wait state means the
```
matutem, line 643: It is unusual and confusing to split a separate case statement that drives the same state as another case statement. It is way clearer to have one chunk of code that performs the actions for each state, together with any transitions. Could you fold this code into the main case for the corresponding states?

This code was already merged in walnut/sow, and got past the reviews with this funny construct. I wish I had flagged this before.

matutem (changes requested):
Flagged some code in the dma FSM. Could you comment?

matutem:
Could you remove the RTL changes since they are already done?

PR #374, opened 2026-08-12
[dvsim] Keep the information when a run or a job is killed

LouisTheLuis:
Built on top of #368.

Killing a `dvsim` run used to throw away nearly everything. A second Ctrl+C escaped before the report was written, the simulators kept running afterwards, and a killed job reported no runtime, no memory and no disk footprint, which are exactly the numbers wanted when a test has run away.
- An interrupted run still writes its report, seed manifest and summary. A second Ctrl+C force-kills what is left instead of crashing out.
- Killing a job signals its whole process group. The command `dvsim` runs is a `make` wrapper, so signalling only that left `simv` alive holding its license. Interactive Ctrl+C hid this, because the terminal was doing the killing.
- A killed job's log is now read, so its runtime and simulated time survive.
- Every job reports `peak_rss_mb` and `odir_size_mb` in `report.json`, beside a new `fail_msg` saying why it died. The peak comes from the `wait4()` rusage for a job that finished, otherwise from `VmHWM` across the job's process tree, measured before the job is signalled.
- New `--max-job-mem-gb` kills a single runaway job and lets the rest of the regression finish, with the reason recorded in the report.

### DIFF
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -326,8 +326,12 @@ class Deploy():
             time, unit = get_job_runtime(log_text, self.sim_cfg.tool)
             self.job_runtime.set(time, unit)
         except RuntimeError as e:
-            log.warning(f"{self.full_name}: {e} Using dvsim-maintained "
-                        "job_runtime instead.")
+            # A job that was killed never got to print its runtime, so this is
+            # expected rather than noteworthy. Interrupting a large regression
+            # would otherwise bury the report behind one warning per job.
+            level = VERBOSE if self.launcher.status == "K" else log.WARNING
+            log.log(level, f"{self.full_name}: {e} Using dvsim-maintained "
+                           "job_runtime instead.")
             self.job_runtime.set(self.launcher.job_runtime_secs, "s")
 
         # Extract the tool-reported memory footprint if available.
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -97,6 +97,15 @@ class Deploy():
         # does not report it). For VCS this is the design's
         # "data structure size", not the process's peak resident memory.
         self.job_mem = None
+        # Job's peak resident set size in MB, as measured by dvsim rather than
+        # reported by the tool (None if it could not be measured). This is the
+        # real memory the job occupied, and unlike job_mem it is known even for
+        # a job that was killed before the tool printed its summary.
+        self.job_peak_rss = None
+        # Size of the job's output directory in MB when the job finished (None
+        # if it could not be measured). Measured before any post-run cleanup,
+        # so it reflects what the job actually wrote.
+        self.job_odir_size = None
 
     def _define_attrs(self):
         """Defines the attributes this instance needs to have.
--- a/util/dvsim/Launcher.py
+++ b/util/dvsim/Launcher.py
@@ -268,6 +268,19 @@ class Launcher:
         """Terminate the job."""
         raise NotImplementedError
 
+    def force_kill(self) -> None:
+        """Terminate the job at once, without waiting for it to wind down.
+
+        Invoked when dvsim is force-quit while this job is still running.
+
+        This default implementation simply defers to kill(). That still gives
+        the job a grace period, which is not what a force-quit asks for, but
+        terminating it slowly beats leaving it behind: an abandoned job carries
+        on consuming licenses, memory and disk long after dvsim has exited.
+        Launchers that can terminate a job more abruptly should override this.
+        """
+        self.kill()
+
     def _check_status(self):
         """Determine the outcome of the job (P/F/K if it ran to completion).
 
--- a/util/dvsim/Launcher.py
+++ b/util/dvsim/Launcher.py
@@ -20,6 +20,11 @@ from utils import VERBOSE, clean_odirs, mk_symlink, rm_path
 # fell over before it got going from one that reported a testbench failure
 _EXIT_CODE_FAIL_MSG = "Job returned non-zero exit code"
 
+# Lines kept from the end of a killed job's log when mining it for runtime and
+# simulated time. Every sim_utils parser scans backwards for the last match, so
+# only the end can carry them, and a runaway job's log has no bound.
+_LOG_TAIL_LINES = 10000
+
 
 def find_first_match(lines, patterns, context_lines=5):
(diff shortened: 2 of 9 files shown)


util/dvsim/Deploy.py
```diff
@@ -17,6 +17,10 @@
 from utils import (VERBOSE, clean_odirs, find_and_substitute_wildcards,
                    rm_path, subst_wildcards)
 
+# Tools that check out no license, so their jobs never queue for one and have
+# nothing to wait for before they start.
+_TOOLS_WITHOUT_LICENSE = frozenset({"verilator"})
```
matutem: Maybe add the verible tools, since they can be run with dvsim using a lint_sim_cfg hjson file.

util/dvsim/Deploy.py
```diff
@@ -482,6 +519,10 @@ class RunTest(Deploy):
     target = "run"
     seeds = []
     fixed_seed = None
+
+    # Any UVM report means the simulator has its license, has loaded the design
+    # and is away.
```
matutem: What does "and is away" mean?

util/dvsim/dvsim.py
```diff
@@ -388,6 +388,26 @@ def parse_args():
...
+                      help=('Kill any single job whose peak resident set size '
+                            'grows beyond GB gigabytes. Only that job is '
+                            'killed; the rest of the regression carries on, '
+                            'and the report says why it died. Off by default. '
+                            'Only applicable when launching jobs locally.'))
+
+    disg.add_argument("--run-wait-timeout-mins",
+                      type=int,
+                      metavar="MINUTES",
+                      help=('Give up on a run that has not started simulating '
```
matutem, line 404: Set default value of 0 so you avoid the confusing code around line 826.

util/dvsim/dvsim.py
```diff
@@ -802,6 +822,10 @@ def main():
     LsfLauncher.LsfLauncher.max_parallel = args.max_parallel
     NcLauncher.NcLauncher.max_parallel = args.max_parallel
     Launcher.Launcher.max_odirs = args.max_odirs
+    Launcher.Launcher.max_job_mem_gb = args.max_job_mem_gb
+    if args.run_wait_timeout_mins is not None:
```
matutem: This code is confusing: the if prevents it from running if run_wait_timeout is not None, yet in line 828 you consider None. It seems better for the arg to default to zero (meaning wait forever), and assign unconditionally.

util/dvsim/Launcher.py
```diff
@@ -236,12 +258,75 @@ def kill(self) -> None:
...
+        """
+        status, err_msg = self._check_log_status()
+        if status != "F":
+            return status, err_msg
+
+        if err_msg.message != _EXIT_CODE_FAIL_MSG:
+            return status, err_msg
+
+        license_err_msg = self._find_license_error()
+        if license_err_msg is not None:
```
matutem: What about just "if license_err_msg:"? The query also fails for empty strings or numerical zero, which is good since an empty string has no information.

util/dvsim/Launcher.py
```diff
@@ -236,12 +258,75 @@ def kill(self) -> None:
...
+        if not patterns:
+            return None
+
+        try:
+            with open(
+                self.deploy.get_log_path(),
+                encoding="UTF-8",
+                errors="surrogateescape",
+            ) as f:
+                lines = f.readlines()
```
matutem: It may be better to iterate over f (which gives lines), so that you don't need to read the full log into memory.

util/dvsim/Launcher.py
```diff
@@ -321,6 +407,36 @@ def _find_patterns(patterns, line):
...
+        much it simulated, and it reporting nothing at all.
+        """
+        if self._log_parsed:
+            return
+        self._log_parsed = True
+
+        log_path = self.deploy.get_log_path()
+        try:
+            with open(log_path, encoding="UTF-8", errors="surrogateescape") as f:
+                lines = f.readlines()
```
matutem: Same as above regarding the iterator over an open file.

util/dvsim/LocalLauncher.py
```diff
@@ -5,17 +5,109 @@
...
+    object, which halves the cost of a call made for every running job on every
+    poll. VmHWM sits within the first kilobyte, well inside a single read, so
+    the long CPU and NUMA bitmasks at the end of the file do not matter.
+
+    Returns None if the process has already exited or procfs cannot be read. A
+    process with no memory map of its own, such as a kernel thread, reports no
+    VmHWM at all and also comes back as None.
+    """
+    try:
+        fd = os.open(f"/proc/{pid}/status", os.O_RDONLY)
```
matutem: Use python `with` to open files (i.e. context managers).

matutem (commented):
Some suggestions

util/dvsim/Deploy.py
LouisTheLuis: Took this out after moving everything into the tools' hjson in #368.

util/dvsim/Deploy.py
LouisTheLuis: It means the run is under way and that the design has finished elaboration. In other words, the license is finally being used.

Moved this comment to `vcs.hjson` in #368.

util/dvsim/dvsim.py
LouisTheLuis: I set the default to 180 minutes instead. Probably better.

util/dvsim/dvsim.py
LouisTheLuis: With the default set as before we can avoid the None branch.

util/dvsim/Launcher.py
LouisTheLuis: Fixed.

util/dvsim/Launcher.py
LouisTheLuis: Addressed in #368.

util/dvsim/LocalLauncher.py
LouisTheLuis: Done.

util/dvsim/Launcher.py
LouisTheLuis: Now keeping track of a max of just 10000 lines.

matutem (approved):
LGTM

PR #376, opened 2026-08-13
[kmac] correctly derive AppIdxW

rpls:
Using `$clog2` yields the wrong result if KMAC only has one application interface, `vbits` returns the correct result.


ysekhar (approved):
LGTM

PR #381, opened 2026-08-14
[topgen,ipgen] Minor fixes for ipconfig generation

vlpannel:
@rtorok-zr recently found a bug while adding a new partition to an OTP memory map where OTP-related collateral was not getting regenerated properly.

Bugs fixed:
* Output path for generated IP configuration files was not quite the same path as the input to ipgen --> fixed filepath
* Errors when getting OTP parameters during topgen (where the OTP information has already been cast as an object) and ipgen (where OTP information is taken straight from the complete top configuration) --> allowed for either type to be used
* Minor typo fix in `util/design/gen-otp-mmap.py` of wrongly-named input argument --> arg renamed

The `util/design/gen-otp-mmap.py` script is old enough that it may be possible to deprecate it, as ipgen generation of otp_ctrl should properly generate all the same output files; happy to remove the file in this PR if feedback says it's warranted.

### DIFF
--- a/hw/top_egret/ip_autogen/flash_ctrl/data/top_egret_flash_ctrl.ipconfig.hjson
+++ b/hw/top_egret/ip_autogen/flash_ctrl/data/top_egret_flash_ctrl.ipconfig.hjson
@@ -16,11 +16,11 @@
     ]
     pages_per_bank: 256
     program_resolution: 8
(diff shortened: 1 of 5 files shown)


rtorok-zr (approved):
LGTM, thanks for tracking this down!

Issue #343, opened 2026-07-30
[topgen] More dependency-driven Makefile


vlpannel:
### Description

The current topgen Makefile (`hw/Makefile`) is driven by few targets which are not related to any specific output files. Especially if multiple tools are to be run (not just topgen), this results in a confusing and rigid Makefile. A better idea would to have the hardware build file better link targets to their dependencies.

adurbin-zr:
@vlpannel -- could you please specify the various targets you are referring to as well as the specifics of the proposal you are thinking of?

vlpannel:
@adurbin-zr This issue is mainly a consequence of breaking topgen up into multiple files in PR pavona/pavona#279.

The current rules in the Makefile are:
* `top`: top hardware generation for every top from a hard-coded list
* `cmdgen`: documentation updates for every markdown file, depends on `top`
* `regs`: register RTL and security countermeasures testplan generation (both via regtool) for every IP block from a hard-coded list
* `top_and_cmdgen`: `cmdgen` (`cmdgen` also requires `top`)
* `dvgen`: DV IP collection for both tops
* `all`: `cmdgen`, `regs`, `dvgen`

The Makefile has no way of generating one specific top or one specific IP block. It does not regard how recently any given file was updated, nor does it separate any of the various top-level outputs

I'd currently propose having as targets instead:
* each top level, separately (depends on tool outputs below)
  * for example, `top_dragonfly`
* the output for each of the broken-out tools (see pavona/pavona#279; includes RTL, documentation, software headers, crossbars, ipgen IPs) per tool and per top, all depending on the complete configuration Hjson
  * for example, `top_dragonfly_sv`, `top_dragonfly_docs`, `top_dragonfly_xbar`, etc.
* each individual complete top configuration Hjson, separately
* the reggen stuff (RTL and security countermeasures testplan) per each IP block that needs them
  * for example, `adc_ctrl_regs`
* `cmdgen`, `dvgen`, and `all` can stay mostly as-is but with minor tweaks

I imagine there are other ways it could be organized, but this presents a more directly useful organization of it; it's unlikely that most people will want to always generate all tops and all registers every time. @matutem has also [suggested using Bazel](https://github.com/pavona/pavona/pull/279#discussion_r3677660007) instead of Make here if that allows for greater sophistication.

PR #383, opened 2026-08-17
[topgen,bazel] Run topgen with Bazel

vlpannel:
Related issue: https://github.com/pavona/pavona/issues/343.

Currently, topgen is most commonly run with a Makefile (`hw/Makefile`) which batch generates all hardware-side collateral for every top with one command. There are two structural changes that might improve building tops without invoking topgen directly:
1) Splitting the build targets up by top and output file group
2) Using Bazel rather than Make, as this conforms to the rest of the repo's build system and allows use of Bazel's more consistent Python environment

There is another draft PR, https://github.com/pavona/pavona/pull/369, that similarly replaces Makefiles with Bazel targets.

This PR:
* exposes more files to Bazel because topgen requires access to all IP block descriptions and IP templates
* creates a custom Bazel rule for building complete top configurations using topgen
* creates a rule for copying autogenerated files back into the repo (as topgen's outputs are all checked-in to the repo)

Because this PR only covers generating the complete config itself, it does not yet obsolete `hw/Makefile` (which generates additional files).

### DIFF
--- a/hw/top_egret/data/BUILD
+++ b/hw/top_egret/data/BUILD
@@ -1,4 +1,5 @@
 # Copyright lowRISC contributors (OpenTitan project).
+# Copyright zeroRISC Inc.
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
@@ -12,4 +13,9 @@ filegroup(
 filegroup(
     name = "doc_files",
     srcs = glob(["**/*_testplan.hjson"]),
+)
+
+filegroup(
+    name = "top_xbar_hjsons",
+    srcs = glob(["xbar*.hjson"]),
 )
--- a/hw/top_egret/data/BUILD
+++ b/hw/top_egret/data/BUILD
@@ -6,6 +6,7 @@
 package(default_visibility = ["//visibility:public"])
 
 load("//rules:autogen.bzl", "pavona_completecfg")
+load("//rules:files.bzl", "copy_files")
 
 filegroup(
     name = "all_files",
@@ -31,4 +32,11 @@ pavona_completecfg(
     ],
     seed = ":top_egret_seed.testing.hjson",
     top = ":top_egret.hjson",
+)
+
+copy_files(
+    name = "gen_completecfg",
+    srcs = [":completecfg"],
+    dest = "hw/top_egret/data/autogen",
+    tags = ["manual"],
 )
--- a/hw/top_egret/data/BUILD
+++ b/hw/top_egret/data/BUILD
@@ -5,6 +5,8 @@
 
 package(default_visibility = ["//visibility:public"])
 
+load("//rules:autogen.bzl", "pavona_completecfg")
+
 filegroup(
     name = "all_files",
     srcs = glob(["**"]),
@@ -18,4 +20,15 @@ filegroup(
 filegroup(
     name = "top_xbar_hjsons",
     srcs = glob(["xbar*.hjson"]),
+)
+
+pavona_completecfg(
+    "completecfg",
+    data = [
+        ":top_xbar_hjsons",
+        "//hw/top_egret/data/otp:otp_ctrl_mmap.hjson",
+        "//hw/top_egret/ip:block_descriptions",
+    ],
+    seed = ":top_egret_seed.testing.hjson",
+    top = ":top_egret.hjson",
 )
--- a/rules/autogen.bzl
+++ b/rules/autogen.bzl
@@ -13,12 +13,79 @@ load(
     "pavona_select_ip_attr",
 )
 
-"""Autogeneration rules for OpenTitan.
+"""Autogeneration rules for Pavona.
 
 The rules in this file are for autogenerating various file resources
-used by the OpenTitan build, such as register definition files generated
+used by the Pavona build, such as register definition files generated
 from hjson register descriptions.
 """
+
+def _pavona_completecfg_impl(ctx):
+    # get paths to output files (complete cfg and secrets file)
+    topname = ctx.file.topcfg.basename.removesuffix(".hjson")
+    seed_mode = ctx.file.seedcfg.basename.split(".")[1]  # top_<topname>.<seed_mode>.hjson
+    completecfg = ctx.actions.declare_file("autogen/" + topname + ".gen.hjson")
+    secrets = ctx.actions.declare_file("autogen/" + topname + ".secrets." + seed_mode + ".gen.hjson")
+
+    # run topgen with given inputs
+    arguments = [
+        "-t",
+        ctx.file.topcfg.path,
+        "-s",
+        ctx.file.seedcfg.path,
+        "-o",
+        "/".join(completecfg.path.split("/")[:-3]),  # top directory in sandbox
+    ]
+    ctx.actions.run(
+        outputs = [completecfg, secrets],
+        inputs = [ctx.file.topcfg, ctx.file.seedcfg] + ctx.files.data,
+        arguments = arguments,
+        executable = ctx.executable._topgen,
+    )
+
+    return [
+        DefaultInfo(files = depset([completecfg, secrets])),
+    ]
+
+pavona_completecfg_rule = rule(
+    implementation = _pavona_completecfg_impl,
+    doc = "Generate complete top config and secrets file with topgen",
+    attrs = {
+        # TODO: add more arguments to pass into topgen; for now, assuming default paths
+        "topcfg": attr.label(allow_single_file = True, doc = "Top configuration Hjson"),
+        "seedcfg": attr.label(allow_single_file = True, doc = "Seed configuration Hjson"),
+        "data": attr.label_list(allow_files = True, doc = "Other top-specific sources needed"),
+        "_topgen": attr.label(
+            default = "//util:topgen.py",
+            executable = True,
+            allow_single_file = True,
+            cfg = "exec",
+        ),
+    },
+)
+
+def pavona_completecfg(name, top, seed, data = [], **kwargs):
+    """
+    Macro around `pavona_completecfg_rule` which uses top config and seed config
+    to generate complete top information with topgen.
+
+    Data should include top-reggen IPs, the RACL config, the OTP memory map, and
+    all xbar configuration hjsons.
+    """
+    needed_for_topgen = [
+        # always required whenever running topgen
+        "//util/topgen:topgen",
+        "//hw/ip:block_descriptions",
+        "//hw/ip_templates:all_files",
+    ]
+
+    pavona_completecfg_rule(
+        name = name,
+        topcfg = top,
+        seedcfg = seed,
+        data = data + needed_for_topgen,
+        **kwargs
+    )
 
 def _pavona_ip_c_header_impl(ctx):
     header = ctx.actions.declare_file("{}_regs.h".format(ctx.attr.ip))
--- a/rules/autogen.bzl
+++ b/rules/autogen.bzl
@@ -56,9 +56,8 @@ pavona_completecfg_rule = rule(
         "seedcfg": attr.label(allow_single_file = True, doc = "Seed configuration Hjson"),
         "data": attr.label_list(allow_files = True, doc = "Other top-specific sources needed"),
         "_topgen": attr.label(
-            default = "//util:topgen.py",
+            default = "//util:toptool",
             executable = True,
-            allow_single_file = True,
             cfg = "exec",
         ),
     },
@@ -72,18 +71,11 @@ def pavona_completecfg(name, top, seed, data = [], **kwargs):
     Data should include top-reggen IPs, the RACL config, the OTP memory map, and
     all xbar configuration hjsons.
     """
-    needed_for_topgen = [
-        # always required whenever running topgen
-        "//util/topgen:topgen",
-        "//hw/ip:block_descriptions",
-        "//hw/ip_templates:all_files",
-    ]
-
     pavona_completecfg_rule(
         name = name,
         topcfg = top,
         seedcfg = seed,
-        data = data + needed_for_topgen,
+        data = data,
         **kwargs
     )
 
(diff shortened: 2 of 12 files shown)


rules/autogen.bzl
```diff
@@ -13,13 +13,140 @@ load(
...
+        relpath = ctx.attr.target_to_copy.label.package
+        if relpath in outp.dirname:
+            relpath += outp.dirname.split(relpath)[-1]
+        elif outp.dirname.startswith("bazel-out/k8-fastbuild/bin/"):  # known possible prefix
+            relpath = outp.dirname.removeprefix("bazel-out/k8-fastbuild/bin/")
+        else:
+            print("WARNING: relative path unknown, copying files to current working directory")
+            relpath = ""
+
+        content += "\ncp {} ./{}".format(bazel_outpath, "/".join([relpath, outp.basename]))
```
fragglet: Nit: it would be more efficient do something like:
```python
copy_commands = [] # before the loop
.....
copy_commands.append("cp {} ./{}".format(bazel_outpath, "/".join([relpath, outp.basename]))
```
and then after the loop:
```python
content += "\n".join(copy_commands)
```
Doing it this way avoids excessive string allocations. It's arguably a tiny bit more readable too.

fragglet (approved):
Great to see more bazelification taking place, thank you!

rules/autogen.bzl
vlpannel: Thanks—change made and will wait to see CI green (or more review feedback) before merging.

hw/top_egret/data/BUILD
```diff
@@ -13,3 +16,24 @@ filegroup(
...
     srcs = glob(["**/*_testplan.hjson"]),
 )
+
+filegroup(
+    name = "top_xbar_hjsons",
+    srcs = glob(["xbar*.hjson"]),
+)
+
+pavona_completecfg(
+    "completecfg",
```
mkannwischer, line 27: I'm trying to run `./bazelisk.sh build //hw/top_egret/data:completecfg`, but I am getting 

```
ERROR: /home/matthias/git/pav/pr-383/hw/top_egret/data/BUILD:25:19: Action hw/top_egret/data/autogen/top_egret.gen.hjson failed: (Exit 1): topgen.py failed: error executing Action command (from target //hw/top_egret/data:completecfg) util/topgen.py -t hw/top_egret/data/top_egret.hjson -s hw/top_egret/data/top_egret_seed.testing.hjson -o bazel-out/k8-fastbuild/bin/hw/top_egret

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Traceback (most recent call last):
  File "/home/matthias/.cache/bazel/_bazel_matthias/a1829d20218b0928b4104e0b87f99928/sandbox/processwrapper-sandbox/2/execroot/_main/util/topgen.py", line 22, in <module>
    from ipgen.clkmgr_gen import get_clkmgr_params
  File "/home/matthias/git/pav/pr-383/util/ipgen/clkmgr_gen.py", line 11, in <module>
    from topgen.clocks import Clocks, ClockSignal
  File "/home/matthias/git/pav/pr-383/util/topgen/__init__.py", line 7, in <module>
    from .merge import merge_top  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matthias/git/pav/pr-383/util/topgen/merge.py", line 14, in <module>
    from raclgen.lib import parse_racl_config, parse_racl_mapping
  File "/home/matthias/git/pav/pr-383/util/raclgen/lib.py", line 13, in <module>
    from reggen.md_helpers import mono, list_item, table, title
  File "/home/matthias/git/pav/pr-383/util/reggen/md_helpers.py", line 9, in <module>
    import tabulate
ModuleNotFoundError: No module named 'tabulate'
Target //hw/top_egret/data:completecfg failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.601s, Critical Path: 0.45s
INFO: 2 processes: 2 internal.

```

rules/autogen.bzl
```diff
@@ -13,13 +13,144 @@ load(
...
+pavona_completecfg_rule = rule(
+    implementation = _pavona_completecfg_impl,
+    doc = "Generate complete top config and secrets file with topgen",
+    attrs = {
+        # TODO: add more arguments to pass into topgen; for now, assuming default paths
+        "topcfg": attr.label(allow_single_file = True, doc = "Top configuration Hjson"),
+        "seedcfg": attr.label(allow_single_file = True, doc = "Seed configuration Hjson"),
+        "data": attr.label_list(allow_files = True, doc = "Other top-specific sources needed"),
+        "_topgen": attr.label(
+            default = "//util:topgen.py",
```
mkannwischer: I think you will need to wrap this into a `py_binary`

rules/autogen.bzl
```diff
@@ -13,13 +13,144 @@ load(
...
 The rules in this file are for autogenerating various file resources
-used by the OpenTitan build, such as register definition files generated
+used by the Pavona build, such as register definition files generated
 from hjson register descriptions.
 """
 
+def _update_generated_impl(ctx):
+    out_file = ctx.actions.declare_file("copyscript.sh")
+
+    content = """#!/usr/bin/bash
```
mkannwischer: We recently switched to ` #!/usr/bin/env bash`

util/topgen/README.md
```diff
@@ -128,6 +128,13 @@ $ cd ${REPO_TOP}
 $ make -C hw top
 ```
 
+Another means of generating the complete config with topgen is by invoking it through Bazel.
```
mkannwischer, line 131: More general point for discussion: Is it a good idea to document multiple ways of achieving something? Shouldn't we rather document one intended way?

hw/top_egret/data/BUILD
vlpannel: Thanks for the check; now this is in the dependencies of the topgen lib. Should work now?

util/topgen/README.md
vlpannel: You might be right here; however, the original "right" way is to run the Makefile.

The Makefile creates the complete top config and then creates a host of other files based on the complete config. Eventually, a set of Bazel targets should obsolete the Makefile, but for now more Bazel targets are needed for generating the other files.

Should we just keep the mention of the Bazel option out of the docs until the Makefile is obsolete?

rules/autogen.bzl
vlpannel: The one tricky thing about this is this will require topgen.py to be renamed, similarly to reggen/regtool and dtgen/dttool. Seems to be because there exists `//util:topgen.py` file and `//util/topgen` package. For now, `util/topgen.py` has been renamed to `util/toptool.py` in a similar fashion.

A follow on PR will more thoroughly replace instances of "topgen.py" in text.

rules/autogen.bzl
vlpannel: Made a fixup commit; thanks.

util/topgen/README.md
mkannwischer: Thanks for the clarification!
I think it's fine to keep it for now as long as we agree that we want to move to only Bazel in the long term and eventually document only that.

cc @fragglet @adurbin-zr @qmn

rules/autogen.bzl
mkannwischer: py_binary has a `main` attribute that defaults to the name. From reading the documentation, it sounds like one could maybe keep main as `topgen.py`, while name becomes `toptool`. Then the Python script itself would not have to be renamed. Have you tried that?

hw/top_egret/data/BUILD
mkannwischer: Yes, works now. Nice. Thanks!

rules/autogen.bzl
vlpannel: Ah, I didn't realize this! The `main` attribute works in this case. Thank you--it's much nicer to not have to rename topgen.

util/topgen/README.md
adurbin-zr: We know bazel is a pain point for many people to use. There's a much broader discussion required within the project to determine a particular direction on this topic. I don't have a strong opinion on this commentary in the README.md at the moment, though.

rules/autogen.bzl
```diff
@@ -13,13 +13,136 @@ load(
...
+        relpath = ctx.attr.target_to_copy.label.package
+        if relpath in outp.dirname:
+            relpath += outp.dirname.split(relpath)[-1]
+        elif outp.dirname.startswith("bazel-out/k8-fastbuild/bin/"):  # known possible prefix
+            relpath = outp.dirname.removeprefix("bazel-out/k8-fastbuild/bin/")
+        else:
+            print("WARNING: relative path unknown, copying files to current working directory")
+            relpath = ""
+
+        copy_commands.append("cp {} ./{}".format(bazel_outpath, "/".join([relpath, outp.basename])))
```
mkannwischer: I think this should be a no-op:
```
rm hw/top_egret/data/autogen/top_egret.gen.hjson
./bazelisk.sh run --run_in_cwd //hw/top_egret/data:gen_completecfg
```
but it actually re-generates the file with different file permissions:

```
$ git diff
diff --git a/hw/top_egret/data/autogen/top_egret.gen.hjson b/hw/top_egret/data/autogen/top_egret.gen.hjson
old mode 100644
new mode 100755
```

If I then run 
```
./bazelisk.sh run --run_in_cwd //hw/top_egret/data:gen_completecfg
```
again, I'm getting
```
cp: cannot create regular file './hw/top_egret/data/autogen/top_egret.gen.hjson': Permission denied
```

rules/autogen.bzl
```diff
@@ -13,13 +13,136 @@ load(
...
-used by the OpenTitan build, such as register definition files generated
+used by the Pavona build, such as register definition files generated
 from hjson register descriptions.
 """
 
+def _update_generated_impl(ctx):
+    out_file = ctx.actions.declare_file("copyscript.sh")
+
+    content = """#!/usr/bin/env bash
+
```
mkannwischer: Probably want to add a `set -e` so this doesn't silently fail without Bazel noticing.

rules/autogen.bzl
```diff
@@ -13,13 +13,136 @@ load(
...
+    # collect all files to copy over and find proper paths
+    copy_commands = []
+    for outp in ctx.attr.target_to_copy.files.to_list():
+        bazel_outpath = "/".join([outp.dirname, outp.basename])
+
+        # find path relative to repo root
+        relpath = ctx.attr.target_to_copy.label.package
+        if relpath in outp.dirname:
+            relpath += outp.dirname.split(relpath)[-1]
+        elif outp.dirname.startswith("bazel-out/k8-fastbuild/bin/"):  # known possible prefix
```
mkannwischer: Unfortunately, neither the `k8`, nor the `fastbuild` are constant here. 
`k8` is only the case when building on x86_64; `fastbuild` is only the default mode. If you set a different `--compilation_mode`, it breaks.

I'm afraid this script is too fragile and is going to bite us later.

util/topgen/README.md
```diff
@@ -128,6 +128,13 @@ $ cd ${REPO_TOP}
...
+Each top can make use of the `//hw/top_*/data:gen_completecfg` target in order to regenerate it (invoked from the repo top):
+```console
+$ ./bazelisk.sh run --run_in_cwd //hw/top_*/data:gen_completecfg
```
mkannwischer: It would be better if we can support this without `--run_in_cwd`.

mkannwischer (changes requested):
Sorry for the incremental and slow review!

I'm a little worried that the copying step is not very robust right now. Have you considered to instead modify `copy_files` and use that? That would allow removing the `update_generated` code, the `--run_in_cwd` and also make it more robust.
Here is a patch that makes that works for me. 

[pr-383-copy-files.patch](https://github.com/user-attachments/files/31288150/pr-383-copy-files.patch)

rules/autogen.bzl
fragglet: You can potentially get the path(s) by using `bazel info`, eg.
```
$ ./bazelisk.sh info
Starting local Bazel server (8.7.0) and connecting to it...
bazel-bin: /home/fra/.cache/bazel/_bazel_fra/ddd316f370480a87e4444160829f25c2/execroot/_main/bazel-out/k8-fastbuild/bin
bazel-genfiles: /home/fra/.cache/bazel/_bazel_fra/ddd316f370480a87e4444160829f25c2/execroot/_main/bazel-out/k8-fastbuild/bin
bazel-testlogs: /home/fra/.cache/bazel/_bazel_fra/ddd316f370480a87e4444160829f25c2/execroot/_main/bazel-out/k8-fastbuild/testlogs
```
Matthias is correct that the path name changes though depending on architecture, eg. on ARM64:
```
$ ./bazelisk.sh info
INFO: Invocation ID: 34bcc85f-d0d5-4e41-b657-543ee31eb608
bazel-bin: /home/fraggle/.cache/bazel/_bazel_fraggle/0c6667ece54a84cb3ef8aeb511b1142d/execroot/_main/bazel-out/aarch64-fastbuild/bin
bazel-genfiles: /home/fraggle/.cache/bazel/_bazel_fraggle/0c6667ece54a84cb3ef8aeb511b1142d/execroot/_main/bazel-out/aarch64-fastbuild/bin
bazel-testlogs: /home/fraggle/.cache/bazel/_bazel_fraggle/0c6667ece54a84cb3ef8aeb511b1142d/execroot/_main/bazel-out/aarch64-fastbuild/testlogs
```
(Apocrypha: 'k8' is a reference to the [AMD K8 architecture](https://en.wikipedia.org/wiki/AMD_K8), the first CPUs that introduced the x86-64 instruction set)

vlpannel:
I did not realize the `copy_files` rule existed; it's much better than the current brittle copying over. Thank you for the thorough review! All the problems mentioned (change in file permissions, variable paths, run_in_cwd) should now be solved.

rules/autogen.bzl
vlpannel: As far as I can tell, `copy_files` does not seem to have this issue.

mkannwischer (approved):
Thanks @vlpannel! LGTM now.
Please still squash that last commit.

PR #387, opened 2026-08-18
[topgen] Set up JSON schema validation

vlpannel:
This PR is the first of three deriving from https://github.com/pavona/pavona/pull/306 (follow on PRs are #388 and #389, in which more of topgen's schemas are exposed).

It has four commits which set up necessary code for using JSON schema to validate the Hjson information fed into topgen for hardware generation.
* add Hjson handling functions to basegen (assuming other tools will need these functions as well), including reading/writing to Hjsons, correctly casting values (related issue: #336), unit tests
* create additional basegen functions for schema validation (creating validators and documenting schemas)
* create a basic schema for a top configuration file and accompanying functions necessary to validate using the schema; also write unit tests for validation functions using the top config schema
* update Bazel dependencies for topgen to require the `jsonschema` Python package

PR #388, opened 2026-08-18
[topgen] First set of schemas

vlpannel:
This PR is the second of three deriving from #306 (following on from #387, preceding #389).

While #387 adds a basic top configuration schema, it does not delve deep into the requirements for each field, instead simply requiring `array` or `object` types without specifying the data types within those structures. Once the schemas are codified in JSON, it will be easier to update them as the tooling needs.

This PR converts various field requirements (which were originally in Python lists in `util/topgen/validate.py`) into JSON schemas as well:
* seed configuration (which is not part of top config, but still should be ported)
* pinmux
  * pinmux signal
    * pinmux signal I/O
  * pinmux I/O count
* pinout
* pad
* target
  * target pinmux
  * target pinout
* straps
* special signals

These are not all the schemas listed in `util/topgen/validate.py`, but another PR will complete the conversion. They should be near-direct translations from the Python lists to JSON schema.


qmn (approved):
Thanks for the great work! I appreciate that you've structured the commits so they build on one another, so it's easy to see that you're recursively enhancing the schema to progressively capture (and thus validate) a larger part of the overall configuration. LGTM

You noted that straps are unused in topgen; we may want to reconsider the future of straps.

PR #392, opened 2026-08-19
[dif,otp] Migrate `otp_ctrl` DIFs to DT

rtorok-zr:
The OTP DIFs had relied on a significant amount of now-outdated metadata that was hard-coded. This included:

- Partition IDs that only matched Egret were hard-coded, preventing the DAI APIs
  from working on Dragonfly.
- The digest offset was assumed to be always the last field, which is no longer always true after zeroization was introduced, as the zeroization field comes after the digest in zeroizable partitions. This caused digest-locking partitions with a software digest to fail, because the DIF wrote the digest to the zeroization field instead.

This commit makes the DIFs utilize DT for partition information, which was updated in the OTP DtGen script to include the necessary layout and operation-compatibility information for the DIFs, as well as metadata for non-sofware-readable partitions, which can still be written using the DAI (e.g. during provisioning in TEST_UNLOCK). This allowed removing the hard coded layout data structures in the OTP DIFs and making the DIFs compatible with a multi-top architecture.

The unit tests were slightly updated to addresss some Egret-specific behavior. Some checks the tests perform on Egret-specific partitions were temporarily commented out. These can be reintroduced later pending the planned OTP MMAP zipper merge, which will unify the partition names for secure boot keys.

### DIFF
--- a/sw/device/lib/dif/dif_otp_ctrl.h
+++ b/sw/device/lib/dif/dif_otp_ctrl.h
@@ -24,238 +24,6 @@ extern "C" {
 #endif  // __cplusplus
 
 /**
- * A partition within OTP memory.
- */
-typedef enum dif_otp_ctrl_partition {
-  /**
-   * Vendor test partition.
-   *
-   * This is reserved for manufacturing smoke checks. The OTP wrapper
-   * control logic inside prim_otp is allowed to read/write to this
-   * region. ECC uncorrectable errors seen on the functional prim_otp
-   * interface will not lead to an alert for this partition.
-   * Instead, such errors will be reported as correctable ECC errors.
-   */
-  kDifOtpCtrlPartitionVendorTest,
-  /**
-   * Software configuration partition.
-   *
-   * This is for device-specific calibration data, e.g, clock, LDO, RNG,
-   * and configuration settings set by the ROM.
-   */
-  kDifOtpCtrlPartitionCreatorSwCfg,
-  /**
-   * Software configuration partition.
-   *
-   * This contains data that changes software behavior in the ROM, for
-   * example enabling defensive features in ROM or selecting failure
-   * modes if verification fails.
-   */
-  kDifOtpCtrlPartitionOwnerSwCfg,
-#if defined(OPENTITAN_IS_EGRET)
-  /**
-   * This OTP partition is used to store four P-256 keys and four Sphincs+ keys.
-   *
-   * The partition requires 464
-   * bytes of software visible storage. The partition is
-   * locked at manufacturing time to protect against
-   * malicious write attempts.
-   */
-  kDifOtpCtrlPartitionRotCreatorAuthCodesign,
-  /**
-   * This OTP partition is used to capture the state of each key slot.
-   *
-   * Each key can be in one of the
-   * following states: BLANK, ENABLED, DISABLED. The
-   * encoded values are such that transitions between
-   * BLANK -> ENABLED ->  DISABLED are possible without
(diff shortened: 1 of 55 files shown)


matutem:
Do you intend to go through with this now that you found there is DT support that would probably make this unnecessary?

rtorok-zr:
> Do you intend to go through with this now that you found there is DT support that would probably make this unnecessary?

As described in #393, this PR has now been completely rearchitected to migrate the OTP DIFs to DT. Some changes to the DtGen for OTP were necesssary to pull this off, but I was able to leave the OTP memory maps alone (e.g. the same partition can be marked zeroizable on one top-level design but not on another).

matutem:
So you added a bunch more partition info in dt. I am curious if it ends up increasing the size of various images, or does LTO remove all unused code?

sw/device/lib/testing/otp_ctrl_testutils.c
```diff
@@ -38,15 +38,16 @@ status_t otp_ctrl_testutils_dai_access_error_check(
...
+    // causes[kOtpPartitionCount] is resrved for the DAI error.
+    if (status.causes[kOtpPartitionCount] != kDifOtpCtrlErrorLockedAccess) {
       LOG_ERROR("Expected access locked error for access to 0x%x", address);
     }
   } else {
     if (bitfield_bit32_read(status.codes, kDifOtpCtrlStatusCodeDaiError)) {
       LOG_ERROR("No DAI error expected for access to 0x%x", address);
     }
-    if (status.causes[kDifOtpCtrlPartitionDaiError] != kDifOtpCtrlErrorOk) {
+    // causes[kOtpPartitionCount] is resrved for the DAI error.
```
velrox77: Nit: resrved --> reserved

sw/device/lib/dif/dif_otp_ctrl.h
```diff
@@ -761,7 +529,7 @@ dif_result_t dif_otp_ctrl_is_digest_computed(const dif_otp_ctrl_t *otp,
  */
 OT_WARN_UNUSED_RESULT
 dif_result_t dif_otp_ctrl_get_digest(const dif_otp_ctrl_t *otp,
```
velrox77, line 532: Dumb question, but if this function (and other similar ones) are no longer using kDifs for the partitions, are the function names still relevant because of the param type dif_otp_ctrl_t for *otp and return type of dif_result_t?

rtorok-zr: I had the same question, actually. I think that the actual `dif_otp_ctrl_t` value is mostly irrelevant now, but because the type is included from autogenerated code (try `bazel build sw/device/lib/dif/autogen:otp_ctrl_src`), I decided to leave it in.

Edit: When running this command just now I realized the `dif_otp_ctrl_t` actually includes a reference to the DT type (`dt_otp_ctrl_t`), which might be more idiomatic to use instead of hard-coding `kDifOtpCtrl`, though it should be functionally the same.

rtorok-zr: Just fixed this now, and in the process the unit tests caught a null-checking bug in `dif_otp_ctrl_read_blocking`.

rtorok-zr:
> So you added a bunch more partition info in dt. I am curious if it ends up increasing the size of various images, or does LTO remove all unused code?

I built the ROM both before and after the change, and the `.text` section is the exact same size, which suggests LTO is helping out.

matutem (approved):
LGTM

PR #394, opened 2026-08-23
[acc,mlkem] Fix off-by-one in the poly_rej_samp accept bound

mkannwischer:
poly_rej_samp draws the uniform polynomial used by refreshmodq and secb2amodq to refresh masking, so the bug below does not change any ML-KEM output; it only makes the refresh mask slightly non-uniform.

The sign of 19*Q - x is only negative for x > 19*Q, so a draw of exactly 19*Q was accepted giving 0 mod q a slightly higher probability than all other values.
Instead use 19*Q-1 so the outputs are uniform mod q.

### DIFF
--- a/sw/acc/crypto/mlkem/gadgets.s
+++ b/sw/acc/crypto/mlkem/gadgets.s
@@ -487,9 +487,9 @@ refreshios:
 .globl poly_rej_samp
 .type poly_rej_samp, @function
 poly_rej_samp:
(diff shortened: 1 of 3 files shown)


mkannwischer:
@phamhnh, could take a look please?

phamhnh (commented):
LGTM! Thank you for catching that.

pqcfox (approved):
Nice catch!

PR #405, opened 2026-08-27
[acc,iflow] Fix simplify method to be dict-ordering agnostic

pqcfox:
Previously, the implementation of `InstructionFlowGraph.simplify()` would fully simplify an instruction flow graph, but only when the `DmemInformationFlowNode` sources for each sink in the graph were correctly ordered (e.g. with `.start` non-decreasing). 

For instance, if three `DmemInformationFlowNode` sources with memory address ranges `0x0 - 0x100`, `0x100 - 0x200`, and `0x200 - 0x300` for a sink were traversed in the order 

```
0x0 - 0x100   ->   0x200 - 0x300   ->   0x100 - 0x200
```

then the resulting simplified sources for the sink would have ranges `0x0 - 0x200` and `0x200 - 0x300`, as the first two (non-adjacent) sources wouldn't be joined.

This PR sorts each sink's sources by start address to remedy this, and introduces unit tests to verify this indeed resolves the issue in simple cases.

### DIFF
--- a/hw/ip/acc/util/shared/BUILD
+++ b/hw/ip/acc/util/shared/BUILD
@@ -3,7 +3,7 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-load("@rules_python//python:defs.bzl", "py_library")
+load("@rules_python//python:defs.bzl", "py_library", "py_test")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -86,7 +86,10 @@ py_library(
 
 py_library(
     name = "encoding",
-    srcs = ["encoding.py"],
+    srcs = [
+        "__init__.py",
+        "encoding.py",
+    ],
     imports = [".."],
     deps = [
         ":bool_literal",
@@ -97,7 +100,10 @@ py_library(
 
 py_library(
     name = "encoding_scheme",
-    srcs = ["encoding_scheme.py"],
+    srcs = [
+        "__init__.py",
+        "encoding_scheme.py",
+    ],
     imports = [".."],
     deps = [
         ":bit_ranges",
@@ -108,7 +114,10 @@ py_library(
 
 py_library(
     name = "information_flow",
-    srcs = ["information_flow.py"],
+    srcs = [
+        "__init__.py",
+        "information_flow.py",
+    ],
     imports = [".."],
     deps = [
         ":operand",
@@ -118,7 +127,10 @@ py_library(
 
 py_library(
     name = "information_flow_analysis",
-    srcs = ["information_flow_analysis.py"],
+    srcs = [
+        "__init__.py",
+        "information_flow_analysis.py",
+    ],
     imports = [".."],
     deps = [
         ":cache",
@@ -133,7 +145,10 @@ py_library(
 
 py_library(
     name = "isr",
-    srcs = ["isr.py"],
+    srcs = [
+        "__init__.py",
+        "isr.py",
+    ],
     imports = [".."],
     deps = [
         "//util/serialize:parse_helpers",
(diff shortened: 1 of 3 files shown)


hw/ip/acc/util/shared/information_flow.py
```diff
@@ -12,8 +12,8 @@
 
 from serialize.parse_helpers import check_keys, check_list, check_str
 
-from .isr import Isr
-from .operand import ImmOperandType, IsrOperandType, Operand, RegOperandType
+from shared.isr import Isr
+from shared.operand import ImmOperandType, IsrOperandType, Operand, RegOperandType
```
mkannwischer: For my understanding: Why was this change made?

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -85,6 +85,13 @@ def union(self, other: 'InformationFlowNode') -> Optional['InformationFlowNode']
             return self
         return None
 
+    def starts_at(self) -> bool:
+        '''Try to determine if this node can be ordered with another node.
+
+        Returns None if the two nodes cannot be ordered.
```
mkannwischer: This doc string does not seem to match what the function is doing? also this isn't returning a bool?

hw/ip/acc/util/shared/BUILD
```diff
@@ -247,3 +271,17 @@ py_library(
     srcs = ["toolchain.py"],
     imports = [".."],
 )
+
+py_test(
```
mkannwischer, line 275: I was surprised that you do not have to load `py_test` in this file. Turns out Bazel autoloads it, but that's going to break in Bazel 9. Probably a good idea to already enforce that now. I'll open a PR setting the corresponding Bazel flag. Can you add the import for `py_test` please?

hw/ip/acc/util/shared/BUILD
```diff
@@ -247,3 +271,17 @@ py_library(
...
+    name = "information_flow_test",
```
mkannwischer, line 276: This test works locally for me, but it's never run in CI.
There seem to be more py_tests tests that are never run. Maybe we open an issue for it? 

As far as I can see these 5 py_tests in  //hw/ are never run
 -  //hw/ip/acc/util/shared:information_flow_test 
 - //hw/ip/rom_ctrl/util:gen_vivado_mem_image_test (fails)
 - util/dvsim/utils_test.py
 - util/topgen/secure_prng_test.py (fails)
 - util/test_reggen/test_rtl.py (fails)

mkannwischer (approved):
Thanks @pqcfox. I've confirmed that this works. Linting is still failing, please fix :-)
Could you please put the description in the commit message in addition to the PR description? That will be much easier to find in the future.

hw/ip/acc/util/shared/BUILD
mkannwischer: See https://github.com/pavona/pavona/pull/408

hw/ip/acc/util/shared/information_flow.py
pqcfox: Thank you! Fixed.

hw/ip/acc/util/shared/BUILD
pqcfox: Fixed, thank you so much! Good catch on this. Could #408 use a preliminary review?

hw/ip/acc/util/shared/information_flow.py
pqcfox: This previously didn't work with the approach I was using for getting the module-level imports to work, but with the `rules_python` bump appears to now! Will revert, thank you for flagging this :)

hw/ip/acc/util/shared/BUILD
mkannwischer: Thanks!
Yes, please feel free to take a look. It's not super urgent - I just thought if I already took the time to understand how this works, I might as well make the improvement.

hw/ip/acc/util/shared/BUILD
pqcfox: Added: https://github.com/pavona/pavona/issues/409, thank you so much!

hw/ip/acc/util/shared/information_flow.py
pqcfox: Done.

mkannwischer:
Took the liberty to resolve the linting failure.

PR #408, opened 2026-08-28
[bazel] Require explicit loads for Python and shell rules

mkannwischer:
Bazel 9 defaults --incompatible_autoload_externally to the empty string (see https://blog.bazel.build/2026/01/20/bazel-9.html), so py_library/ sh_test/etc. must be loaded explicitly.
Set the flag already now to prepare Bazel 9 support and add the missing loads throughout the codebase.

Bump aspect_rules_lint to 1.13.0 as 1.2.2 relied on sh_test being autoloaded.

Bump rules_fuzzing to 0.8.0 as 0.5.2 relied on sh_binary being
autoloaded.

More changes will be needed to migrate to Bazel 9.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,6 +6,9 @@
 common --enable_bzlmod
 common --enable_workspace=no
 
+# Require explicit `load()` statements, as Bazel 9 does by default.
+common --incompatible_autoload_externally=
+
 # Share extracted external repositories (e.g. the LLVM toolchain) across pavona checkouts.
 common --repo_contents_cache=~/.cache/pavona-repo-contents-cache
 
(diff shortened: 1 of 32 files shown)


pqcfox (approved):
Looks excellent to me! Out of curiosity, are the version bumps needed for compatibility with this change, or just incidental as part of possibly moving to Bazel 9?

mkannwischer:
> Looks excellent to me! Out of curiosity, are the version bumps needed for compatibility with this change, or just incidental as part of possibly moving to Bazel 9?

Yes the version bumps were needed as the versions we were pinning relied on autoloading. See the commit message:
```
Bump aspect_rules_lint to 1.13.0 as 1.2.2 relied on sh_test being autoloaded.

Bump rules_fuzzing to 0.8.0 as 0.5.2 relied on sh_binary being
autoloaded.
```

pqcfox:
Makes sense, thank you so much!

PR #417, opened 2026-08-31
[test] Mark flaky FPGA e2e tests broken

mkannwischer:
These tests are the largest single source of spurious CI failures over the past week.
I suggest we open an issue to investigate those separately. 

Failure counts per target over that week:

   9  rescue_rom_ext_slot_a_update_slot_b
   5  e2e_chip_specific_startup_rma
   4  rescue_rom_ext_slot_b_update_slot_a
   4  secver_write_test
   4  rescue_rate_test
   3  rescue_rom_ext_slot_a_update_slot_a
   3  bad_activate_test
   2  rescue_rom_ext_slot_b_update_slot_b
   2  transfer_pq_to_pq_test
   
CI already runs the FPGA tests with --flaky_test_attempts=3, so each of
these failures is a test that failed three times in a row.

### DIFF
--- a/sw/device/silicon_creator/rom/e2e/chip_specific_startup/BUILD
+++ b/sw/device/silicon_creator/rom/e2e/chip_specific_startup/BUILD
@@ -28,7 +28,8 @@ package(default_visibility = ["//visibility:public"])
         },
         fpga = fpga_params(
             otp = "//sw/device/silicon_creator/rom/e2e/rom_e2e_bootstrap_entry:otp_img_e2e_bootstrap_entry_{}".format(lc_state),
-            tags = maybe_skip_in_ci(lc_state_val),
+            # Broken: fails intermittently on CW310.
+            tags = maybe_skip_in_ci(lc_state_val) + ["broken"],
             test_cmd = """
                 --bootstrap="{firmware}"
                 --otp-unprogrammed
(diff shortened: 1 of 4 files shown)


sw/device/silicon_creator/rom/e2e/chip_specific_startup/BUILD
```diff
@@ -28,7 +28,9 @@ package(default_visibility = ["//visibility:public"])
         },
         fpga = fpga_params(
             otp = "//sw/device/silicon_creator/rom/e2e/rom_e2e_bootstrap_entry:otp_img_e2e_bootstrap_entry_{}".format(lc_state),
-            tags = maybe_skip_in_ci(lc_state_val),
+            # Broken: the rma variant fails intermittently on CW310.
+            tags = maybe_skip_in_ci(lc_state_val) +
+                   (["broken"] if lc_state == "rma" else []),
```
rtorok-zr: I can't speak for this exact test, but for most of the silicon creator software tests, the specific lifecycle state won't have much impact on the reliability. For this one I would either mark as broken for all values of `lc_state` or not at all.

rtorok-zr (approved):

sw/device/silicon_creator/rom/e2e/chip_specific_startup/BUILD
mkannwischer: Thanks @rtorok-zr!
interesting. I've only seen this test fail for lc_state=rma in the past week.
Disabling for all states for now until we have time to investigate.

mkannwischer: Ahh, all other states `lc_state` aren't run in CI anyway due to the maybe_skip_in_ci. That explains.

PR #420, opened 2026-08-31
[acc,iflow] Track every WDR a non-constant index could reference 

mkannwischer:
An indirect WDR reference whose index register is not a known constant
was represented by a single synthetic "wref-<gpr>" node, which is not
the register the instruction touches. The real register was never
marked a sink and kept its stale dependencies, so a secret loaded by
bn.lid or stored by bn.sid never reached the graph and a const-time
check could pass while missing the flow.

Name all the WDRs the reference could mean instead: over-reporting as a
source is safe, and as a sink each one is a may-write, so nothing
already tracked is erased. The index register becomes a source too.
Coarse DMEM writes use the same may-write path, replacing a special
case that only covered the last node evaluated.

- Based on #405

### DIFF
--- a/hw/ip/acc/util/shared/information_flow.py
+++ b/hw/ip/acc/util/shared/information_flow.py
@@ -20,6 +20,9 @@ FLAG_NAMES = ['c', 'm', 'l', 'z']
 # Deliberately invalid CSR address used by unimp.
 UNIMP_CSR_ADDR = 3072
 
+# Number of wide data registers.
+NUM_WDRS = 32
+
 # Registers that can be involved in information flow without being an operand
 SPECIAL_REG_NAMES = ['mod', 'acc', 'acch', 'kmac_core', 'rnd', 'urnd']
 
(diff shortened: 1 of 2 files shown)


hw/ip/acc/util/shared/information_flow_test.py
```diff
@@ -76,5 +78,73 @@ def test_different_types_out_of_order(self) -> None:
...
+
+        # x5 is known, so only w3 is written, and fully.
+        self.assertEqual(flow.flow, {
+            InformationFlowNode('w3'): {InformationFlowNode('dmem')}
+        })
+
+    def test_unknown_index_as_sink(self) -> None:
+        flow = self._iflow('bn.lid').evaluate(self.LID_OPS, {}, True)
+
+        # Every WDR is a may-write sink, and x5 selects which one.
```
pqcfox: kat: why is it node -> same node in the analysis? couldn't a separate variable cause a may-write of one node over another?

hw/ip/acc/util/shared/information_flow.py
```diff
@@ -694,6 +697,23 @@ def required_constants(self, op_vals: Dict[str, int],
     def is_read_only(self, op_vals: Dict[str, int]) -> bool:
         return False
 
+    def is_may_write(self, op_vals: Dict[str, int],
+                     constant_regs: Dict[str, int], coarse_dmem: bool) -> bool:
+        '''True if this sink may be written rather than certainly written.
+
+        A may-write keeps whatever the node held before, since the write is
```
pqcfox: [super nit] if I understand the changes right, for clarity it might be useful to note that the may-write keeps (iflow-wise) whatever the node held before, as well as the usual set of nodes that might write to it?

pqcfox (approved):
Looks fantastic, great catch on this!

hw/ip/acc/util/shared/information_flow.py
mkannwischer: Thanks! I've clarified the comment.

PR #353, opened 2026-08-05
[acc,mldsa] Add masked ACC ML-DSA cryptolib integration

mkannwischer:
Wire the cryptolib ML-DSA backend to the hardened run_mldsa_hardened ACC
binary. Key generation and signing operate on Boolean-masked key
material, with the masked signing key stored in the ACC key layout. The
derandomized keygen API takes a seed buffer twice as long, holding two
shares of the seed. Verification is public and unchanged.

Also enable the ACVP and Wycheproof tests (run in CI) for the masked implementation.

- Depends on https://github.com/pavona/pavona/pull/348
- Depends on https://github.com/pavona/pavona/pull/340
- Depends on https://github.com/pavona/pavona/pull/288
- Depends on #338 
- Depends on #341
- Depends on #359

PR #423, opened 2026-09-01
[acvp,mldsa] Cover the HashML-DSA and external-mu modes in ACVP tests

mkannwischer:
Add tests for HashML-DSA w/ SHA2-256/384/512, SHA3-224/256/384/512,
SHAKE128/256 as pre hash.
SHA2-224, SHA2-512/224, SHA2-512/256 are not (yet) supported as the
cryptolib doesn't currently support those hash functions.

- Depends on #353


pqcfox (approved):
Looks excellent!

PR #424, opened 2026-09-01
[acc,mlkem] Clean up ML-KEM's gadgets.

phamhnh:
(updated on Sept 7)

This PR proposes mainly refactoring and consistency work for ML-KEM and its gadgets,
together with some small clean ups for ML-KEM.

### Gadgets
- `secand` and `secfulladder` no longer advance their input/output
  pointers; callers do it explicitly.
- Clean up the remaining gadgets.
- `w16` (`sw0`) is now a constant maintained by the caller across the
  whole ML-KEM flow, holding `q` in lane 0 and `-q^-1 mod 2^16` in lane 2.
  Gadgets no longer re-derive it, and the clobber lists reflect that.

### Naming and formatting

Conventions adopted across the ML-KEM sources and tests:

- Rename the ML-KEM constants to a `const_*` convention.
- `poly_hocompress` -> `poly_hocompress_dv`,
  `polyvec_hocompress` -> `poly_hocompress_du`,
  `poly_masked_compare_{du,dv}` -> `masked_poly_compare_{du,dv}`.
- Numeric `x*` register names throughout.
- 2-space indentation, matching the rest of `sw/acc/crypto`.
- Improve all code comments.
- Use Doxygen-style comment header blocks.

### Tests
- Fix the assembly test files for gadgets that require `w16`. Several set
  only `q` in `mod` and left `sw0.2` (`-q^-1 mod 2^16`) at 0; others never
  loaded `w16` at all, relying on `secb2amodq` deriving it internally,
  which this PR removes. Each such test now loads `qinv | q` into `w16`
  and writes `mod` from it.

  Note these tests passed either way: the affected lane only feeds
  `poly_rej_samp`, whose output is used as a refresh mask that cancels
  when the shares are recombined, so a wrong `sw0.2` is unobservable to
  them. This class of setup bug is invisible to the gadget tests as they
  stand.
  
- Clean up the ML-KEM and gadget assembly test files.
- Clean up the Python testgens.

### DIFF
--- a/sw/acc/crypto/mlkem/gadgets.s
+++ b/sw/acc/crypto/mlkem/gadgets.s
@@ -68,9 +68,9 @@
 .type secand, @function
 secand:
   /* Save addresses. */
-  addi x5, x10, 0
-  addi x6, x12, 0
-  addi x7, x15, 0
+  add x5, x10, x0
+  add x6, x12, x0
+  add x7, x15, x0
 
   /* Load x. */
   bn.xor w0, w0, w0   /* Whitening. */
@@ -288,16 +288,16 @@ secadd:
 
   /* Initialize c = 0. */
   bn.xor w0, w0, w0
-  addi   x5, x2, 0
+  add    x5, x2, x0
   loopi 2, 1
     bn.sid x0, 0(x5++)
   endloop
 
   /* Ripple-carry adder. */
   addi x8, x17, -1
-  addi x17, x2, 0
+  add  x17, x2, x0
   addi x29, x0, 32
-  addi x30, x2, 0
+  add  x30, x2, x0
   addi x31, x0, 32
   /* Handle bit i = 0..k - 2. */
   loop x8, 2
@@ -306,7 +306,7 @@ secadd:
   endloop
 
   /* Handle bit k - 1. */
-  addi x5, x2, 0
+  add  x5, x2, x0
   addi x4, x0, 1
   addi x6, x0, 2
   addi x7, x0, 3
@@ -598,8 +598,8 @@ poly_to_bitsliced:
   jal x1, _bitslice_transpose
 
   /* Store the 12 bitsliced words r[0..11] via x10 so x11 is left unchanged. */
-  addi   x4, x0, 0
-  addi   x10, x11, 0
+  add x4, x0, x0
+  add x10, x11, x0
   loopi 12, 2
     bn.sid x4, 0(x10++)
     addi   x4, x4, 1
@@ -629,7 +629,7 @@ poly_to_bitsliced:
 .type poly_from_bitsliced, @function
 poly_from_bitsliced:
   /* Load x[0..11] into w0..w11; zero the upper bit positions w12..w15. */
-  addi x4, x0, 0
+  add x4, x0, x0
   loopi 12, 2
     bn.lid x4, 0(x10++)
     addi   x4, x4, 1
@@ -896,7 +896,7 @@ seca2b:
   /* Save x3 to stack */
   addi x2, x2, -32
   sw   x3, 0(x2)
-  addi x3, x2, 0
+  add  x3, x2, x0
 
   /* Adjust stack for temp variables. */
   slli x5, x12, 1
@@ -935,17 +935,17 @@ seca2b:
   add x28, x14, x0
 
   /* Compute r = secadd(s, s', k). */
-  addi x10, x7, 0
-  addi x11, x6, 0
-  addi x12, x2, 0
-  addi x13, x6, 0
-  addi x15, x28, 0
-  addi x16, x6, 0
-  addi x17, x5, 0
-  jal  x1, secadd
+  add x10, x7, x0
+  add x11, x6, x0
+  add x12, x2, x0
+  add x13, x6, x0
+  add x15, x28, x0
+  add x16, x6, x0
+  add x17, x5, x0
+  jal x1, secadd
 
   /* Restore x2 and x3. */
-  addi x2, x3, 0
+  add  x2, x3, x0
   lw   x3, 0(x2)
   addi x2, x2, 32
   ret
@@ -983,7 +983,7 @@ seca2bmodq:
    *   x2 + 1728 : saved x18 */
   addi x2, x2, -1760
   sw   x18, 1728(x2)
-  addi x18, x12, 0
+  add  x18, x12, x0
 
   /* Compute s = p + x_0, k + 1 bits (one share).
    * p = 2^(k + 1) - q = 4863 = 0b1001011111111.
@@ -1068,7 +1068,7 @@ seca2bmodq:
   endloop
 
   /* Build s' = (0, x_1) for (k + 1) bits. */
-  addi x5, x2, 0 /* s' */
+  add x5, x2, x0 /* s' */
   loopi 13, 1
     bn.sid x0, 0(x5++)
   endloop
@@ -1090,7 +1090,7 @@ seca2bmodq:
 
   addi x10, x2, 896
   addi x11, x0, 416
-  addi x12, x2, 0
+  add  x12, x2, x0
   addi x13, x0, 416
   addi x15, x2, 896
   addi x16, x0, 416
@@ -1130,7 +1130,7 @@ seca2bmodq:
   /* Compute a = bitcopymask(u[k], (k + 1) * 32). */
   addi x10, x2, 1280
   addi x11, x0, 416
-  addi x13, x2, 0
+  add  x13, x2, x0
   jal  x1, bitcopymask
 
   /********** Start inline r = secadd(a, u, k). **********/
@@ -1141,11 +1141,11 @@ seca2bmodq:
     bn.sid x0, 0(x5++)
   endloop
 
-  addi x10, x2, 0
+  add  x10, x2, x0
   addi x11, x0, 384
   addi x12, x2, 896
   addi x13, x0, 416
-  addi x15, x18, 0
+  add  x15, x18, x0
   addi x16, x0, 384
   addi x17, x2, 832
   addi x29, x0, 32
@@ -1216,12 +1216,12 @@ seconebitb2amodq:
   addi x2, x2, -1056
   sw   x8, 1024(x2)
   sw   x18, 1028(x2)
-  addi x8, x10, 0
-  addi x18, x12, 0
+  add  x8, x10, x0
+  add  x18, x12, x0
 
   /* Build v = (x_0, 0). */
-  addi x5, x2, 0
-  add  x6, x10, x0
+  add x5, x2, x0
+  add x6, x10, x0
   loopi 16, 2
     bn.lid x0, 0(x6++)
     bn.sid x0, 0(x5++)
@@ -1237,9 +1237,9 @@ seconebitb2amodq:
   bn.shv.16h w30, w30 >> 15
 
   /* Compute v = refreshmodq(v). */
-  addi x10, x2, 0
-  addi x12, x10, 0
-  jal  x1, refreshmodq
+  add x10, x2, x0
+  add x12, x10, x0
+  jal x1, refreshmodq
 
   /* We need to compute r = (1 - 2 * x_1) * v_j for j = 0..1.
    * Since x_1 is either 0 or 1, this translates to:
@@ -1260,12 +1260,12 @@ seconebitb2amodq:
    *    Then (t1 - v_j) mod q = (-v_j) mod q.
    */
   addi x5, x8, 512 /* x_1 */
-  addi x6, x2, 0   /* v */
+  add  x6, x2, x0  /* v */
   addi x4, x0, 1
   loopi 16, 16
     bn.lid         x0, 0(x5++)
     bn.subv.16h    w2, w0, w30
-    addi           x7, x6, 0
+    add            x7, x6, x0
     /* Handle v_0. */
     bn.lid       x4, 0(x6)
     bn.and       w3, w1, w2
@@ -1284,9 +1284,9 @@ seconebitb2amodq:
   endloop
 
   /* Compute r = refreshmodq(v). */
-  addi x10, x2, 0
-  addi x12, x18, 0
-  jal  x1, refreshmodq
+  add x10, x2, x0
+  add x12, x18, x0
+  jal x1, refreshmodq
 
   /* Restore registers and frame. */
   lw   x8, 1024(x2)
@@ -1338,8 +1338,8 @@ secb2amodq:
   sw x21, 16(x2)
 
   /* Save input/output addresses and buffer bases. */
-  addi x8, x10, 0
-  addi x18, x12, 0
+  add  x8, x10, x0
+  add  x18, x12, x0
   addi x19, x2, 1888 /* a, b, c, zp (bitsliced) */
   addi x20, x2, 1056 /* s */
   /* zp = x2 + 32 */
@@ -1348,7 +1348,7 @@ secb2amodq:
   addi    x4, x0, 30
   la      x5, modulus_bn
   bn.lid  x4, 0(x5)
-  addi    x10, x12, 0
+  add     x10, x12, x0
   addi    x7, x2, 32
   bn.wsrr w16, mod
   jal x1, poly_rej_samp
@@ -1360,7 +1360,7 @@ secb2amodq:
 
   /* Bitslice zp_0 and clear zp_1 (bitsliced). */
   addi x10, x2, 32
-  addi x11, x19, 0
+  add  x11, x19, x0
   jal  x1, poly_to_bitsliced
   addi x11, x11, 384
   bn.xor w0, w0, w0
@@ -1369,8 +1369,8 @@ secb2amodq:
   endloop
 
   /* Compute a = seca2bmodq(zp). */
-  addi x10, x19, 0
-  addi x12, x19, 0
+  add  x10, x19, x0
+  add  x12, x19, x0
   jal  x1, seca2bmodq
 
   /* Compute b = secaddmodq(a, x). */
@@ -1384,11 +1384,11 @@ secb2amodq:
     addi   x5, x5, 416
   endloop
 
-  addi x10, x19, 0
+  add  x10, x19, x0
   addi x11, x0, 384
-  addi x12, x8, 0
+  add  x12, x8, x0
   addi x13, x0, 384
-  addi x15, x20, 0
+  add  x15, x20, x0
   addi x16, x0, 416
   addi x17, x20, 384
   addi x29, x0, 416
@@ -1409,19 +1409,19 @@ secb2amodq:
   loopi 2, 1
     bn.sid x0, 0(x5++)
   endloop
-  addi x21, x5, 0 /* p */
+  add x21, x5, x0 /* p */
 
-  addi    x5, x21, 0
+  add     x5, x21, x0
   bn.subi w1, w0, 1
   addi    x4, x0, 1
   bn.sid  x4, 0(x5++)
   bn.sid  x0, 0(x5++)
 
-  addi x10, x21, 0
+  add  x10, x21, x0
   addi x11, x0, 32
-  addi x12, x20, 0
+  add  x12, x20, x0
   addi x13, x0, 416
-  addi x15, x20, 0
+  add  x15, x20, x0
   addi x16, x0, 416
   addi x17, x2, 32
   addi x29, x0, 32
@@ -1436,20 +1436,20 @@ secb2amodq:
   /* Bit 8: p[i] = 0. */
   bn.xor w0, w0, w0
   bn.sid x0, 0(x10)
-  addi   x10, x21, 0
+  add    x10, x21, x0
   jal    x1, secfulladder
   /* Bit 9: p[i] = 1. */
   bn.xor  w0, w0, w0
   bn.subi w0, w0, 1
-  addi    x10, x21, 0
+  add     x10, x21, x0
   bn.sid  x0, 0(x10)
   jal     x1, secfulladder
   /* Bits 10..11: p[i] = 0. */
   bn.xor w0, w0, w0
-  addi   x10, x21, 0
+  add    x10, x21, x0
   bn.sid x0, 0(x10)
   jal    x1, secfulladder
-  addi   x10, x21, 0
+  add    x10, x21, x0
   jal    x1, secfulladder
 
   /* Bit 12: p[i] = 1. */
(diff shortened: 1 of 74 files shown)


mkannwischer (changes requested):
Thanks @phamhnh for this massive PR which looks like a big improvement!

As dicussed offline:
The current structure of the PR makes it somewhat hard to review. To speed up the review process, could you please separate out cosmetic changes that don't change the resulting object code into it's own PR? For the other changes it would help us a lot if you could split the changes by logical changes rather than by gadget, e.g., apply the `bn.xor wX, w31, w31`  whitening changes to all gadgets in one commit. Thanks again for all your work! If the above is not feasible, @pqcfox and I can help out.

phamhnh:
Thanks @mkannwischer for the advice. I have split the changes into smaller
commits to make the review easier:

- **`secand` and `secfulladder`:** I could not separate these cleanly, because
  I also rewrote them so that the shares are no longer accessed consecutively.
  Both gadgets are short, so I hope they are still manageable as a single
  commit each.
- **All other gadgets:** these commits are still titled "Clean up ...", but each
  one is small in scope, e.g. improved register usage.
- **Whitening:** this is split into four commits, in order:
  1. remove the whitening between different bits of one share,
  2. move the whitening step to after the registers are used,
  3. add the missing whitening steps for share isolation,
  4. use `w31` for whitening.

  The list above covers the ML-KEM gadgets; the same sequence is applied
  separately to the ML-KEM main functions.
 
As discussed offline, I have also removed the whitening of the LSU paths from
this PR, since this can be done with evaluation later.

Happy to split anything further if that would help.

rtorok-zr:
FYI: Please rebase your changes. #310 caused a breakage in the Vivado builds in CI, which was reverted by #442.

phamhnh:
This PR now contains only the clean-ups for ML-KEM and its gadgets; all whitening-related commits have been split out into #447.

sw/acc/crypto/mlkem/mlkem_indcpa_enc_cmp.s
```diff
@@ -799,20 +799,14 @@ _compute_k4_consts:
...
+  addi x2, x2, -64
 
   /* The first share of r is (1 << N) - 1. The other shares are 0. */
   add     x5, x2, x0
-  bn.xor  w0, w0, w0
-  bn.subi w0, w0, 1
+  bn.subi w0, w31, 1
+  bn.sid  x0, 0(x5++)
+  bn.xor  w0, w31, w31
   bn.sid  x0, 0(x5++)
```
mkannwischer, line 809: 02a80493658d2739222c4c28fd82d6b694875262
Why are we not just storing `w31` here?

sw/acc/crypto/mlkem/mlkem_indcpa_dec.s
```diff
@@ -287,17 +287,12 @@ indcpa_dec:
...
-      bn.subvm.16h w0, w31, w0
-      bn.sid       x0, 0(x11++)
-    endloop
-    /* Whitening. */
-    bn.xor w0, w0, w0
+  loopi 16, 3
+    bn.lid       x0, 0(x11)
+    bn.subvm.16h w0, w31, w0
+    bn.sid       x0, 0(x11++)
   endloop
```
mkannwischer, line 296: 02a80493658d2739222c4c28fd82d6b694875262
Unlike before, this now leaves behind michroarchitectural state in w0.
So masked_poly_tomsg needs to be aware that it has to clear w0 before filling it with any secret data. What's the convention you are using? It would be good if we can document the convention for the ML-KEM implementation.

sw/acc/crypto/mlkem/gadgets.s
mkannwischer, line 1: cc4191e0b3ebd0eae839e4358c24419d88a71658 introduces trailing whitespace. Can you squash it with the later commit that remove it again please?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -42,28 +42,28 @@
  *           https://eprint.iacr.org/2019/910
  */
 
-/*
- * Name: secand
+/**
```
mkannwischer, line 45: fa3cca892a0917b465adfa0bdc0ce3bde5fcf5c2 promises the unify the x_i, x[i] notation, but it doesn't seem to happen in this commit. squash with the commit that actually does it?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -2077,8 +2049,12 @@ onebitdecompress:
...
 masked_cbd:
-  /* Frame: tmp at 0 (768 B), sum at 768 (64 B), s at 832 (384 B, sized for
-   * the worst case eta = 3), saved registers at 1216. */
+  /* Frame:
+   *   x2 +    0 : b ( 768 B)
+   *   x2 +  768 : a (  64 B)
+   *   x2 +  832 : s ( 384 B)
+   *   x2 + 1216 : save registers. */
+  addi x2, x2, -1248
   addi x2, x2, -1248
```
mkannwischer, line 2058: c8cf6d9b36e12f44566529de4ba2e2255fe64c48 should be comment only, but it changes code in secand and masked_cbc. The code in masked_cbd now looks like it can't work:
```
  addi x2, x2, -1248
  addi x2, x2, -1248
```

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -68,9 +68,9 @@
 .type secand, @function
 secand:
   /* Save addresses. */
-  addi x5, x10, 0
-  addi x6, x12, 0
-  addi x7, x15, 0
+  add x5, x10, x0
+  add x6, x12, x0
+  add x7, x15, x0
```
mkannwischer, line 73: Why was the change in 3d89b98b04120c96305d74f7d8020f2a8f9ca4ec made? The commit message doesn't state a reason. If there is a reason to change it here, we should change it in all of ML-KEM.

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -67,25 +67,21 @@
 .globl secand
 .type secand, @function
 secand:
```
mkannwischer: ffa0b382c6f3f99aef2d0779db135a496a4cef62 isn't a mere clean-up. It's actually optimizing the gadget. We'll need to re-run the benchmarks prior to merging.

sw/acc/crypto/mlkem/pack_keys.s
```diff
@@ -114,44 +114,44 @@ poly_frombytes:
 
     loopi 16, 2
       bn.rshi w1, w0, w1 >> 16
-      bn.rshi w0, w0, w0 >> 12
```
mkannwischer, line 117: What's the motivation for the change in 88914a3e8db1c5c1ed1eb2f76b63c5841fba3287? Stating that in the commit message would help.

mkannwischer (commented):
I've reviewed the first 10 commits and most of the changes look like a great improvement. Thanks @phamhnh! I have a couple of questions.

sw/acc/crypto/mlkem/mlkem_indcpa_enc_cmp.s
phamhnh: Storing `w31` also requires one instruction `addi x4, x0, 31`. So I think performance-wise, it's the same. But this place you pointed out can be done a tiny bit better by saving to `0(x2)` and `32(x2)` directly. I'll update this.

sw/acc/crypto/mlkem/mlkem_indcpa_dec.s
phamhnh: At the time, the convention I was using is to whiten all registers before using them. So in this flow, `w0` will be whitened if used in `masked_poly_tomsg`.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: So the notation is unified only in the comment header blocks. The next commit https://github.com/pavona/pavona/pull/424/commits/c8cf6d9b36e12f44566529de4ba2e2255fe64c48 cleans up and uses these notations for all code comments.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: Nice catch! Thank you. This is fixed in https://github.com/pavona/pavona/pull/424/commits/1bfdd073f1bbe2642ace5d6151c1706e992e9946. But I'll update https://github.com/pavona/pavona/commit/c8cf6d9b36e12f44566529de4ba2e2255fe64c48 to fixing comments only.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: It's just a coding convention that I would follow myself. This has been done for all ML-KEM routines in the previous PR. I can squash this change with the "Clean up..." commit chain later if that makes more sense.

sw/acc/crypto/mlkem/pack_keys.s
phamhnh: The motivation is to make it similar to other functions where all the used bits are rolled out using `w31 = 0` register. Functional correctness is not affected. I'll add explanation in the commit message.

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -296,33 +293,28 @@ secadd:
...
-    bn.xor w3, w0, w1
-    bn.xor w3, w3, w2
-    bn.sid x7, 0(x15)
-    /* Adjust addresses. */
     add    x10, x10, x11
+    bn.lid x4, 0(x12)
     add    x12, x12, x13
-    add    x5, x5, x29
+    bn.xor w0, w0, w1
+    bn.lid x4, 0(x30++)
```
mkannwischer, line 307: Do we need whitening of w1 before loading c_i here?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -990,8 +989,8 @@ seca2bmodq:
     bn.not w3, w0
     bn.xor w1, w3, w2
     bn.sid x4, 0(x5++)
-    bn.xor w2, w0, w2
-    bn.and w2, w3, w2
+    bn.xor w2, w2, w0
+    bn.and w2, w2, w3
```
mkannwischer, line 993: What's the motivation for this change?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -1079,25 +1075,19 @@ seca2bmodq:
...
-    bn.xor w3, w0, w1
-    bn.xor w3, w3, w2
-    bn.sid x7, 0(x15)
-    /* Adjust addresses. */
     add    x10, x10, x11
+    bn.lid x4, 0(x12)
     add    x12, x12, x13
-    add    x17, x17, x29
+    bn.xor w0, w0, w1
+    bn.lid x4, 0(x30++)
```
mkannwischer, line 1088: Do we need to add whitening before this lid?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -2380,24 +2373,27 @@ _getnoise_eta_2:
...
+    bn.sid x4, 0(x10++)
     addi   x4, x4, 1
-    bn.sid x4, 0(x28++)
+    bn.sid x4, 0(x11++)
     addi   x4, x4, 1
-    bn.sid x4, 0(x28++)
+    bn.sid x4, 0(x11++)
+
+    /* Whitening. */
+    bn.xor w17, w17, w17
```
mkannwischer, line 2396: Why is this needed? If I understand the code above correctly, w17 is already zero here.

sw/acc/crypto/tests/mlkem/gadgets/finalize_cmp_test.s
```diff
@@ -8,32 +8,31 @@
...
-    ecall
+  ecall
 
 .data
 .balign 32
 stack:
-    .zero STACK_SIZE
+  .zero STACK_SIZE
 stack_end:
-    .zero 0
```
mkannwischer, line 39: 48373f3b69907563810b78a28436492fdfebdc6d claims to be whitespace only, but it removes `.zero 0` which is restored in the next commit. Please make it whitespace only.

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_getnoise_test.s
mkannwischer, line 1: bf80f3dbff2628b609413bf77ec1383bf44f000f hints at a test gap. Can we improve these tests so we would catch such an issue in the future?

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -1499,9 +1499,9 @@ secb2amodq:
  * clobbered flag groups: FG0
  */
 
-.globl poly_hocompress
-.type poly_hocompress, @function
-poly_hocompress:
+.globl poly_hocompress_dv
+.type poly_hocompress_dv, @function
+poly_hocompress_dv:
```
mkannwischer, line 1504: I've been wondering about this for a while: Does it still make sense to call it HOCompress for 2 shares? No need to change anything.

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_dv_testgen.py
```diff
@@ -39,21 +39,38 @@ def random_shared_poly() -> Tuple[bytes, List[int]]:
...
+
+    # Generate the packed reference ciphertext.
+    cv_dv4_bytes = compress_poly(rv, 4)
+    cv_dv5_bytes = compress_poly(rv, 5)
+
+    if invalid:
+        # Pick a random index in the ciphertext and modify a random bytes.
+        idx = random.randrange(128)
+        cv_dv4_bytes = cv_dv4_bytes[:idx] + bytes(cv_dv4_bytes[idx] ^ 1) + \
+            cv_dv4_bytes[idx + 1:]
```
mkannwischer, line 59: This isn't quite doing what you want it to do. `bytes(n)` gives you n zero bytes, so this is generating lots of zero bytes instead of one byte with a bitflip.
`bytes([cv_dv4_bytes[idx] ^ 1]) ` is what you want.

sw/acc/crypto/tests/mlkem/gadgets/BUILD
mkannwischer, line 1: nit: 40c904d6931e96cc61e3f058181409a1163a40a4 uses the wrong function names in the commit message.

sw/acc/crypto/mlkem/pack_keys.s
```diff
@@ -23,6 +23,7 @@
  * @param[in]  x10: dmem pointer to the input polynomial
  * @param[out] x11: dmem pointer to the output byte array
  * @param[in]  w31: all-zero register
+ * @param[in]  mod: q = 3329
```
mkannwischer, line 26: Please apply this uniformly. I believe it's also missing from basemul_acc, poly_tomont, and indcpa_dec.

mkannwischer (changes requested):
Thanks @phamhnh for this large PR that makes a lot of improvements to the ML-KEM gadgets. I have managed to review the remaining commits and left a couple of comments. Everything looks correct to me. 

I'd appreciate if we can still fix the commit structure. Instead of having 19 commits saying "Clean-up gadget x" and no description of the motivation for the changes, I think it is better to squash those into a single commit and describe in the commit message what is being cleaned up and why. Ultimately, you are doing the same cleanup for all of them.
Maybe you can also combine 48373f3b69+8c915bb0f8 (test file formatting),  ab601a2e0d+c5004ac23c+2d6b3f4ed2 (renames), fc085d06b9+69ad638229 (constants).

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_dv_testgen.py
phamhnh: fixed

sw/acc/crypto/mlkem/gadgets.s
phamhnh: I would keep it as `hocompress` because that's the name of the original algorithm.

sw/acc/crypto/tests/mlkem/gadgets/BUILD
phamhnh: fixed

sw/acc/crypto/tests/mlkem/gadgets/finalize_cmp_test.s
phamhnh: This is merged with https://github.com/pavona/pavona/pull/424/commits/8c915bb0f8eca374940d8f05640065fb95cb2cfc.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: Consistency with the other blocks that do the same thing and readability.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: Registers are whitened each iteration so that no register ever holds values from both shares. Here `w1` only ever holds first shares, so the whitening is not needed in my opinion.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: Registers are whitened each iteration so that no register ever holds values from both shares. Here `w1` only ever holds first shares, so the whitening is not needed.

sw/acc/crypto/mlkem/gadgets.s
phamhnh: `w17` is whitened at this step because when entering the loop it holds the first digest. But you're right it's already 0 here. I'll remove this.

sw/acc/crypto/mlkem/pack_keys.s
phamhnh: Fixed. I added this for `poly_tomont`. For `basemul_acc`, it's already there. `indcpa_dec` doesn't need `mod` as input.

phamhnh:
Thanks @mkannwischer for reviewing. I have addressed all your comments above.

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_dv_test.s
```diff
@@ -0,0 +1,61 @@
...
+  /* r <- masked_poly_compare_dv(xv, cv_dv5), k = 4 path (dv = 5). */
+  la   x2, stack_end
+  la   x10, xv
+  la   x11, cv_dv5
+  addi x12, x0, 160
+  la   x14, r
+  addi x15, x0, 4
+  jal  x1, masked_poly_compare_dv
+
+  /* r &= masked_poly_compare_dv(xv, cv_dv4), k != 4 path (dv = 4). */
```
mkannwischer: Can we not & the two results here?
If one of them were broken, the negative test would still pass.

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_du_test.s
```diff
@@ -0,0 +1,61 @@
...
+  /* r <- masked_poly_compare_du(xu, cu_du11), k = 4 path (du = 11). */
+  la   x2, stack_end
+  la   x10, xu
+  la   x11, cu_du11
+  addi x12, x0, 352
+  la   x14, r
+  addi x15, x0, 4
+  jal  x1, masked_poly_compare_du
+
+  /* r &= masked_poly_compare_du(xu, cu_du10), k != 4 path (du = 10). */
```
mkannwischer: same here.

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -75,1231 +42,1221 @@
...
+  bn.xor  w3, w3, w3   /* Whitening. */
+  bn.xor  w3, w6, w8   /* w3 = t_0 ^ s */
+  bn.and  w3, w5, w3   /* w3 &= a_1 */
+  bn.xor  w9, w9, w9   /* Whitening. */
+  bn.not  w9, w5       /* w9 = a_1 ^ 1 */
+  bn.and  w9, w9, w8   /* w9 &= s */
+  bn.xor  w3, w3, w9   /* w3 ^= w9 */
+  bn.xor  w2, w2, w3   /* w2 ^= w3 */
+  bn.xor  w3, w3, w3   /* Whitening. */
+  bn.xor  w3, w2, w1   /* cout_0 = x_1 ^ w2 */
```
mkannwischer:
```suggestion
  bn.xor  w3, w2, w1   /* cout_1 = x_1 ^ w2 */
```

sw/acc/crypto/mlkem/gadgets.s
```diff
@@ -1308,1239 +1265,1107 @@ seconebitb2amodq:
...
+  sw   x8, 1540(x2)
+  sw   x9, 1544(x2)
+
+  /* Create 2^(alpha - 1). */
+  bn.subi   w30, w31, 1
+  bn.shv.8s w30, w30 >> 31
+  bn.shv.8s w30, w30 << 12
+
+  /* Select alpha-dependent parameters: w30 = 2^(alpha - 1), x8 = the
+   * extraction byte offset alpha * 32, x9 = dv. */
```
mkannwischer:
```suggestion
   * extraction byte offset alpha * 32, x9 = du. */
```

mkannwischer (changes requested):
Thanks @phamhnh. Here are a few more comments.

https://github.com/pavona/pavona/pull/424#discussion_r4016380447 is still open. Easiest way would be to extend the testgen to test w16 has the correct value in the end of the test. That should be applied to `refreshmodq`, `secb2amodq`, `seconebitb2amodq`, `masked_poly_frommsg` and `masked_poly_getnoise` .

`addi xY, xZ, 0` is still used in some other ML-KEM asm files. Could you replace those while at it?

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_dv_test.s
phamhnh: Thanks! Fixed. The two paths now report into separate registers (`w1` for `k = 4`, `w0` for `k != 4`) instead of being AND-ed into a single r, so the false test fails if either path wrongly reports a match.

sw/acc/crypto/tests/mlkem/gadgets/masked_poly_compare_du_test.s
phamhnh: fixed

phamhnh:
Thanks @mkannwischer again! All the issues above are addressed and merged into relevant commits.

PR #425, opened 2026-09-01
Fix obsolete references to master branches

fragglet:
While reviewing an update to docs.pavona.org, I noticed some of the links there point to files in `pavona/pavona/master` - which is obviously nonsensical, although the links do work because Github does a helpful redirect. This fixes that (in `mdbook_code_snippet.py`), but I also followed up by fixing some other references to 'master' both in Pavona and in vendored projects that have stopped using that name for their main branches.


adurbin-zr (approved):

hw/vendor/lowrisc_ibex/vendor/google_riscv-dv/verilog_style/build-verible.sh
vlpannel, line 1: I'm not sure if CI checks for this, but I would be wary of directly editing vendored files; they may get overwritten to whatever is upstream when re-running the vendoring script.

vlpannel (approved):
LGTM with one small comment, though make sure that slow lint (which checks vendored files) passes. Thanks for the thoroughness!

hw/vendor/lowrisc_ibex/vendor/google_riscv-dv/verilog_style/build-verible.sh
fragglet: Ah, you're right - sorry, I missed that this was a vendored file. Fixed.

mkannwischer (approved):

PR #426, opened 2026-09-01
[dvsim] Apply job limits to Slurm jobs and reuse the results repository

LouisTheLuis:
This PR contains three commits:
- The first commit stops publishing from deleting and re-cloning the dashboard staging directory whenever the target changes, so a run that publishes to both the private and the public dashboard no longer clones a dashboard twice. `clone_or_pull` keeps one remote per dashboard and returns the one to push to.
- The second commit moves the run timeout, the license wait detection and the memory ceiling out of `LocalLauncher` and into `Launcher`, so a job dispatched through Slurm is subject to all three and reports its wall clock time instead of leaving that column empty.
- The third commit gives `SlurmLauncher` a way to measure a running job: the generated script reports its own job id and the launcher asks `sstat` for the peak. `max_job_mem_gb` is therefore enforced for Slurm jobs, and `force_kill` records the figure before the job goes.

### DIFF
--- a/util/dvsim/SlurmLauncher.py
+++ b/util/dvsim/SlurmLauncher.py
@@ -8,10 +8,36 @@ import shlex
 import shutil
 import subprocess
 import tempfile
+import time
 from pathlib import Path
+from typing import Union
 
 from Launcher import ErrorMessage, Launcher, LauncherError
 from utils import VERBOSE
+
+
+def _parse_size_mb(text: str) -> Union[float, None]:
+    """Convert a size as Slurm writes it into MB, or None if it is not one.
+
+    sacct and sstat report sizes with a single letter unit and sometimes a
+    decimal point, as in `1234K` or `1.50G`. A bare number means bytes. An
+    empty field, which is what a step that has reported nothing yet gives, is
+    not a size.
+    """
+    text = text.strip()
+    if not text:
+        return None
+
+    scale = {'K': 1 / 1024, 'M': 1.0, 'G': 1024.0, 'T': 1024.0 * 1024}
+    factor = 1 / (1024 * 1024)
+    if text[-1].upper() in scale:
+        factor = scale[text[-1].upper()]
+        text = text[:-1]
+
+    try:
+        return float(text) * factor
+    except ValueError:
+        return None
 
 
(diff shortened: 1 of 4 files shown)


util/dvsim/SlurmLauncher.py
```diff
@@ -8,12 +8,37 @@
...
 import subprocess
 import tempfile
+import time
 from pathlib import Path
 
 from Launcher import ErrorMessage, Launcher, LauncherError
 from utils import VERBOSE
 
 
+def _parse_size_mb(text: str):
```
ngildenhuys: Set return type

util/dvsim/SlurmLauncher.py
```diff
@@ -154,6 +195,81 @@ def kill(self):
             'K',
             ErrorMessage(line_number=None, message='Job killed!', context=[]))
 
+    def _job_id(self):
```
ngildenhuys: Return type needed here as well

util/dvsim/SlurmLauncher.py
```diff
@@ -154,6 +195,81 @@ def kill(self):
...
+        nothing to re-read after the first success.
+        """
+        if self._slurm_job_id is None:
+            try:
+                self._slurm_job_id = self.slurm_jobid_file.read_text().strip() or None
+            except OSError:
+                return None
+        return self._slurm_job_id
+
+    def _peak_mem_mb(self):
```
ngildenhuys: return type

ngildenhuys (changes requested):
Just a few nits on function signatures

ysekhar (commented):
LGTM as long as @ngildenhuys concerns are addressed

util/dvsim/SlurmLauncher.py
LouisTheLuis: Fixed.

util/dvsim/SlurmLauncher.py
LouisTheLuis: Fixed.

util/dvsim/SlurmLauncher.py
LouisTheLuis: Fixed.

ngildenhuys (approved):

PR #385, opened 2026-08-18
[bazel,sw] Add partial macOS support

mkannwischer:
This PR adds partial support for running tests on macOS. Not everything works
yet, but the current state should be enough for the tutorial we are giving
at CHES 2026.

The following tests work:
 - ACC crypto tests using the instruction-set simulator
 - Chip-level crypto tests using Verilator
 - Verilated uart_smoketest
 - ACC smoke tests 

For the details on changes required see the individual commits.

Tested on macOS 26.6.2 (Apple Silicon):

```sh
brew bundle install --file=Brewfile
python3 -m venv .venv
source .venv/bin/activate
pip3 install -r python-requirements.txt --require-hashes

# the CI job's steps
bazelisk build //sw/device/examples/hello_world:hello_world_sim_dv
bazelisk test --test_output=errors --//hw:verilator_options=--threads,1 \
    //sw/device/tests:uart_smoketest_sim_verilator
make -C hw/ip/acc/dv/accsim test bazel=bazelisk

# ACC smoke tests
bazelisk test --test_output=errors --//hw:verilator_options=--threads,1 \
    //hw/ip/acc/dv/smoke:run_smoke_test //hw/ip/acc/dv/smoke_pqc:run_smoke_test

# ACC crypto tests using the ISS
bazelisk test --test_output=errors //sw/acc/crypto/...
```

~~This one should also work, but I'm still running it~~ This has failing tests, but they also fail on Linux. I'll investigate this separately. 
```sh
# chip-level crypto tests using Verilator
bazelisk test --test_output=errors --//hw:verilator_options=--threads,1 \
    --test_timeout=7200 \
    $(bazelisk query 'filter("_sim_verilator$", tests(//sw/device/tests/crypto/...))')
```

- Depends on https://github.com/pavona/pavona/pull/369
- Resolves #412

PR #428, opened 2026-09-02
[setup] Debug macOS 15

mkannwischer:


mkannwischer:
- Fixed and added to #385

PR #432, opened 2026-09-02
[ci] Migrate all non-PQ CW340 tests to nightly CI

rtorok-zr:
CW340 runner wait time has been long due to limited hardware bandwidth, so this PR migrates the non-PQ CW340 tests to the nightly CI to speed up PR approvals.

### DIFF
--- a/.github/workflows/egret-cw340-nightly.yml
+++ b/.github/workflows/egret-cw340-nightly.yml
@@ -12,104 +12,89 @@ on:
 
 jobs:
   # CW340 FPGA jobs
-  execute_long_test_rom_fpga_tests_cw340:
+  execute_test_rom_fpga_tests_cw340:
     name: CW340 Test ROM Long Tests
     uses: ./.github/workflows/fpga.yml
     secrets: inherit
     with:
       name: CW340 Test ROM Long Tests
-      job_name: execute_long_test_rom_fpga_tests_cw340
+      job_name: execute_test_rom_fpga_tests_cw340
       bitstream: ${{ inputs.bitstream }}
       board: cw340
       interface: cw340
       tag_filters: cw340_test_rom
-      timeout_filters: 'long,eternal'
+      timeout_filters: 'short,moderate,long,eternal'
       timeout: 300
 
-  execute_long_rom_fpga_tests_cw340:
+  execute_rom_fpga_tests_cw340:
     name: CW340 ROM Long Tests
     uses: ./.github/workflows/fpga.yml
     secrets: inherit
     with:
       name: CW340 ROM Long Tests
-      job_name: execute_long_rom_fpga_tests_cw340
+      job_name: execute_rom_fpga_tests_cw340
       bitstream: ${{ inputs.bitstream }}
       board: cw340
       interface: cw340
       tag_filters: "cw340_rom_with_fake_keys,cw340_rom_with_real_keys,-manuf"
-      timeout_filters: 'long,eternal'
+      timeout_filters: 'short,moderate,long,eternal'
       timeout: 300
 
-  execute_long_rom_ext_fpga_tests_cw340:
+  execute_rom_ext_fpga_tests_cw340:
     name: CW340 ROM_EXT Long Tests
     uses: ./.github/workflows/fpga.yml
     secrets: inherit
     with:
       name: CW340 ROM_EXT Long Tests
-      job_name: execute_long_rom_ext_fpga_tests_cw340
+      job_name: execute_rom_ext_fpga_tests_cw340
       bitstream: ${{ inputs.bitstream }}
       board: cw340
       interface: cw340
       tag_filters: cw340_rom_ext
-      timeout_filters: 'long,eternal'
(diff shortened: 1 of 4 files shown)


.github/workflows/egret-cw340-pr.yml
```diff
@@ -103,7 +23,6 @@ jobs:
       board: cw340
       interface: cw340
       tag_filters: cw340_pqc_test_rom
-      timeout_filters: 'short,moderate,long,eternal'
```
rtorok-zr, line 106: Question for @mkannwischer : Was it intentional the PQ tests were configured to run in the on-demand tests, even those with timeout > `moderate`?

.github/workflows/nightly.yml
```diff
@@ -26,6 +26,16 @@ jobs:
     uses: ./.github/workflows/ci.yml
     secrets: inherit
 
+  # Build non-PQC CW340 bitstream
+  chip_egret_cw340:
```
rtorok-zr, line 30: Aside: we should test if the existing CW340 tests work on the PQC-enabled bitstream. If they do, we could probably remove the non-PQ one from the CI entirely.

mkannwischer: They should work just fine, but I haven't tested it yet.
I just didn't want to touch the other tests and not duplicate any tests.

.github/workflows/nightly.yml
```diff
@@ -26,6 +26,16 @@ jobs:
...
+    name: Egret for CW340
+    needs: quick_lint
```
mkannwischer: I don't think this works because `quick_lint`  is defined in the other workflow.

.github/workflows/nightly.yml
mkannwischer: does this need a `needs: chip_egret_cw340` so it only starts after the bitstream is built?

.github/workflows/egret-cw340-pr.yml
mkannwischer: The PQC tests have a `long` timeout, so I'm afraid that's needed. 
With this change this job will be a no-op.

.github/workflows/egret-cw340-pr.yml
mkannwischer, line 32: I'm surprised you didn't have to modify the verify-fpga job. How does this not complain about the missing CW340 runs?

.github/workflows/egret-cw340-nightly.yml
```diff
@@ -102,7 +102,7 @@ jobs:
     secrets: inherit
     with:
       name: CW340 PQCEn=1 Long Tests
-      job_name: execute_long_pqc_fpga_tests_cw340
+      job_name: execute_pqc_fpga_tests_cw340
```
mkannwischer: Looks like an accidental edit? I believe we need to keep that job_name unique.

.github/workflows/egret-cw340-nightly.yml
```diff
@@ -12,88 +12,88 @@ on:
...
+      job_name: execute_manuf_fpga_tests_cw340
       bitstream: ${{ inputs.bitstream }}
       board: cw340
       interface: cw340
       tag_filters: "manuf,-cw310"
-      timeout_filters: 'long,eternal'
+      timeout_filters: 'short,moderate,long,eternal'
       timeout: 300
 
   execute_long_pqc_fpga_tests_cw340:
```
mkannwischer: Thinking about it more, this tests just duplicates the per-PR one. 
Let's remove the job from the nightly and keep the `long` timeout filter in the PR?

mkannwischer (changes requested):
Thanks @rtorok-zr for starting this! I support this change and think we should merge it as soon as possible. 
If I understand correctly, a couple of things still need to change before this will work as intended.

.github/workflows/egret-cw340-pr.yml
rtorok-zr: It will, I fixed this just now.

.github/workflows/egret-cw340-nightly.yml
rtorok-zr: I removed this job, per your other comment.

.github/workflows/nightly.yml
```diff
@@ -39,7 +48,7 @@ jobs:
   # CW340 nightly FPGA jobs
   execute_nightly_tests_egret_cw340:
     name: Egret CW340 Nightly Tests
-    needs: on-demand
+    needs: chip_egret_cw340
```
mkannwischer: With this change I think it can happen that the nightly's `Verify` job starts before on-demand has finished and then it would fail. Easiest would be stick to on-demand here.

mkannwischer (approved):
Thanks @rtorok-zr. I think there is one race condition left, but the rest looks good to me.

.github/workflows/nightly.yml
rtorok-zr: Ah, it should be 

```
needs:
- on-demand
- chip_egret_cw340
```

PR #433, opened 2026-09-03
[bazel] Make bazelisk.sh work on AArch64 macOS 

### LINKED ISSUE #413: [macOS] `bazelisk.sh` is incompatible with macOS
https://github.com/pavona/pavona/pull/385 added partial macOS support, but did not touch the `bazelisk.sh` script which is compatible with the sh shipped in macOS. Users can still run everything through `bazelisk` which is part of the Brewfile, but since most of Pavona's documentation is using `bazelisk.sh`, it would be preferable to make the script portable.

mkannwischer:
#385 added partial macOS support but skipped making bazelisk.sh portable. As bazelisk.sh is the recommended way of using Pavona right now, this PR makes bazelisk.sh portable. I've done a few simplifications on the way. See the individual commits. 

- Resolves #413

### DIFF
--- a/.github/workflows/setup.yml
+++ b/.github/workflows/setup.yml
@@ -89,17 +89,17 @@ jobs:
           echo "$HOME/.local/share/venv/bin" >> "$GITHUB_PATH"
           ~/.local/share/venv/bin/pip install -r python-requirements.txt --require-hashes
       - name: Build device software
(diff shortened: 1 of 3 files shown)


fragglet (approved):
Nice! Thanks for this.

PR #437, opened 2026-09-03
[dvsim] Publish with a key that needs no agent and no passphrase

LouisTheLuis:
`_setup_ssh_env` returned `env` on the path where neither an agent nor `SSH_KEY_PASSPHRASE` is set, having never assigned it, so a machine authenticating with a passwordless deploy key raised `UnboundLocalError` at the moment it went to publish. That is the configuration the log line on that path describes, so it could never have worked.

It also replaced `GIT_SSH_COMMAND` with a bare `ssh` on every path. A caller holding one key per repository selects between them with an ssh configuration file passed that way, and overwriting it leaves ssh to guess the identity and authenticate as whoever owns the first key it offers. An inherited value is now kept.

### DIFF
--- a/util/dvsim/FlowCfg.py
+++ b/util/dvsim/FlowCfg.py
@@ -678,19 +678,28 @@ class FlowCfg():
         log.info("Dashboard branch %r found on %s.", self.branch, repository)
 
     def _setup_ssh_env(self) -> dict:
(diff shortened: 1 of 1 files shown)


qmn (approved):
LGTM, thank you!

PR #438, opened 2026-09-03
[dvsim] Keep the caller's ssh command when checking repository access

LouisTheLuis:
`check_ssh_git_access` replaced `GIT_SSH_COMMAND` with a bare `ssh` in order to
add `BatchMode`. A caller holding one deploy key per repository selects between
them with an ssh configuration file passed that way, and the URLs it publishes
to name aliases from that file, which a bare `ssh` cannot resolve at all. The
check therefore reported no access to every repository, and it runs before
anything is published, so `--publish` and `--publish-prev` both failed there
rather than at the push.

`BatchMode` is now added to whatever the caller asked for, which leaves the
configuration file in place and keeps the check non-interactive.


qmn (approved):
Looks good, thank you!

PR #454, opened 2026-09-10
[bazel] Silence clang's GCC installation detection warning

### LINKED ISSUE #397: [openocd] Environment dependent warning breaks build
### Description

On some systems (namely shared-workstation) the openocd build fails with this warning (which is upgraded to an error). I'm unsure whether we should simply ignore it by adding `"-Wno-gcc-install-dir-libstdcxx"` to `third_party/openocd/BUILD` or if this is a larger error in the bazel hermetic build setup.

```
$ bazel build //third_party/openocd:openocd_bin
<snip>
clang: error: future releases of the clang compiler will prefer GCC installations containing libstdc++ include directories; '/usr/lib/gcc/x86_64-linux-gnu/11' would be chosen over '/usr/lib/gcc/x86_64-linux-gnu/12' [-Werror,-Wgcc-install-dir-libstdcxx]
```

mkannwischer:
On hosts that have a newer GCC without libstdc++ headers next to a complete older one, clang 22 warns that a future release will pick the older installation instead.  openocd is built with -Werror, so there the warning is an error and `bazel build //third_party/openocd:openocd_bin` fails.

The choice does not matter for us: nothing resolves out of the GCC installation anyway.

Fixes #397

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -277,6 +277,8 @@ register_toolchains("//third_party/foreign_cc:preinstalled_make_toolchain_macos"
 llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
 llvm.toolchain(
     name = "llvm_toolchain",
+    # TODO: Drop once clang's GCC detection prefers complete installations.
+    extra_compile_flags = {"": ["-Wno-gcc-install-dir-libstdcxx"]},
     # macOS links the system libc++ dylib, so use its headers too.
     extra_cxx_flags = {"darwin-aarch64": [
         "-nostdinc++",


rtorok-zr (approved):
This seems okay for now, but I think it's better to put C/C++ compiler flags in `//toolchain/BUILD`.

mkannwischer:
> This seems okay for now, but I think it's better to put C/C++ compiler flags in `//toolchain/BUILD`.

Thanks!
`//toolchain/BUILD` is for the device toolchain, right? Only host code is affected here, so that won't work.
One could possibly move it to `.bazelrc`, but I'm unsure if that reaches all builds and I don't have time to properly test it today.

PR #456, opened 2026-09-11
[rom] Fix fault behavior in dual-stage ROM

rtorok-zr:
The ROM1 ePMP region did not include the interrupt vector, which caused exceptions that occurred during ROM1 execution to trigger a double-fault when attempting to run the exception handler. This PR fixes this and also fixes the shutdown handler to properly disable key manager on DPE platforms.

### DIFF
--- a/sw/device/silicon_creator/lib/BUILD
+++ b/sw/device/silicon_creator/lib/BUILD
@@ -410,8 +410,8 @@ dual_cc_library(
         ["HAS_FLASH_CTRL"],
         [],
     ) + pavona_if_ip(
-        "keymgr",
-        ["HAS_KEYMGR"],
+        "keymgr_dpe",
+        ["HAS_KEYMGR_DPE"],
         [],
     ),
     deps = dual_inputs(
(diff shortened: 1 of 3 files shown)


sw/device/silicon_creator/lib/BUILD
pqcfox, line 561: Would it make sense to toss up an issue for possibly going through and eliminating `HAS_KEYMGR`, instead using `#ifndef HAS_KEYMGR_DPE` to check for presence of keymgr? (since `shutdown.c` in this PR appears to treat the absense of `HAS_KEYMGR_DPE` as implying the presence of keymgr)

pqcfox (commented):
Looks great! One possible follow up to mention, let me know if it makes sense and I can toss up the issue/handle that real fast.

pqcfox (approved):

sw/device/silicon_creator/lib/BUILD
rtorok-zr: Maybe, but regular keymgr is going away soon, so we'll be able to get rid of these entirely.

pqcfox: Makes sense, I'll hold off in that case; thanks!

PR #457, opened 2026-09-11
[rom,test] Fix ROM ePMP test on Dragonfly

rtorok-zr:
Previously, the ROM ePMP test used too large a region of CTN as the executable region, which caused load faults when running the test on Dragonfly, because not all of the CTN region is accessible by Ibex.


sw/device/silicon_creator/rom/rom_epmp_test.c
```diff
@@ -347,10 +361,8 @@ static void test_noexec_ctn(void) {
   // Ideally we'd check all of eFlash but that takes a very long time in
   // simulation. Instead, check the first and last words are not executable and
   // check a sample of other addresses.
-  uint32_t *ctn =
-      (uint32_t *)dt_soc_proxy_memory_base(kDtSocProxy, kDtSocProxyMemoryCtn);
-  size_t ctn_len = dt_soc_proxy_memory_size(kDtSocProxy, kDtSocProxyMemoryCtn) /
-                   sizeof(ctn[0]);
+  uint32_t *ctn = (uint32_t *)TOP_DRAGONFLY_SOC_PROXY_RAM_CTN_BASE_ADDR;
+  size_t ctn_len = TOP_DRAGONFLY_SOC_PROXY_RAM_CTN_SIZE_BYTES;
```
pqcfox: This `ctn_len` value appears to be in bytes, but the `ctn_len` value prior appears to be in words, maybe instead:
```suggestion
  size_t ctn_len = TOP_DRAGONFLY_SOC_PROXY_RAM_CTN_SIZE_BYTES / sizeof(ctn[0]);
```
?

sw/device/silicon_creator/rom/rom_epmp_test.c
```diff
@@ -401,22 +413,19 @@ static void test_unlock_exec_region(void) {
...
-                   sizeof(ctn[0]);
-  uint32_t *image = &ctn[ctn_len / 5];
-  size_t image_len = ctn_len / 7;
+      sizeof(exec_region[0]);
+#elif defined(OPENTITAN_IS_DRAGONFLY)
+  uint32_t *exec_region = (uint32_t *)TOP_DRAGONFLY_SOC_PROXY_RAM_CTN_BASE_ADDR;
+  size_t exec_region_size_words =
+      TOP_DRAGONFLY_SOC_PROXY_RAM_CTN_SIZE_BYTES / sizeof(exec_region[0]);
 #endif
+  uint32_t *image = &exec_region[exec_region_size_words / 5];
```
pqcfox, line 427: (clarification q) where do the `/ 5` and `/ 7` come from in this line and the next?

pqcfox (changes requested):
Thanks for this! One possible typo to mention and one quick clarification question, otherwise this looks great :)

sw/device/silicon_creator/rom/rom_epmp_test.c
rtorok-zr: They were arbitrary values chosen when the test was first added to OpenTitan, that are intended to avoid the region of flash the test itself is executing from. With CTN this is not a problem, but there was no need to hard-code specific offsets for both memory types, either.

pqcfox: Makes sense! Thank you.

pqcfox (approved):
Saw the fix for the one typo, approved--thanks!

PR #459, opened 2026-09-15
[ci] Fix `Lint (slow)` and enforce all steps to pass

### LINKED ISSUE #440: [lint,ci] rom_e2e_self_hash: non-existent target '//sw/device/silicon_creator/rom/e2e/release:rom_e2e_self_hash_test_sim_dv'
https://github.com/pavona/pavona/actions/runs/33828301864/job/100886791507#step:9:28:

```
Errors in hw/top_egret/dv/chip_egret_sim_cfg.hjson:
- chip_sw_pwrmgr_sleep_wake_5_bug: non-existent target '//sw/device/tests:pwrmgr_sleep_wake_5_bug_test_sim_dv' in sw_image=//sw/device/tests:pwrmgr_sleep_wake_5_bug_test:1:new_rules
Errors in hw/top_egret/dv/chip_rom_tests.hjson:
- rom_e2e_self_hash: non-existent target '//sw/device/silicon_creator/rom/e2e/release:rom_e2e_self_hash_test_sim_dv' in sw_image=//sw/device/silicon_creator/rom/e2e/release:rom_e2e_self_hash_test:1:new_rules
Error: Process completed with exit code 1.
```

I'm surprised that CI still lets this through.

mkannwischer:
The slow linting had the `Bazel test suite tags` and `DV software images` steps marked as `continue-on-error: true` and they were in fact failing for some time, see, e.g., [today's nightly](https://github.com/pavona/pavona/actions/runs/34919985194/job/104228178289).

This PR fixes the failures and removes the `continue-on-error: true`. There are two more `continue-on-error: true` in `sw_build_test` and  the `df_sw_build_test` jobs , but those look like more effort to fix. 

 - Resolves https://github.com/pavona/pavona/issues/440


hw/top_egret/dv/chip_rom_tests.hjson
vlpannel, line 1: Should the same also be done to the analogous Dragonfly file (`hw/top_dragonfly/dv/chip_rom_tests.hjson`)?

vlpannel (approved):
LGTM with one comment.

hw/top_egret/dv/chip_rom_tests.hjson
rtorok-zr: Probably, but I'm planning to uncomment the target it points to in a later commit once the ROM is more stable. It was commented out in OpenTitan because the expected hash values are hard-coded in the test and you have to update them every time you change the ROM source code.

mkannwischer: check_dv_sw_images.sh is egret only, so for that test it's definitely not needed.

I briefly looked at hw/top_dragonfly/dv/chip_rom_tests.hjson and it doesn't look like it has the same issue, but it has a lot of other missing targets.

vlpannel: Ah, you're right; I noticed that Dragonfly has its own self_hash test, but it is reliant on a different Bazel target (not the one commented out). Fixes for those missing targets could probably be a different PR since they're out-of-scope for solving the CI test failures.

fragglet:
Thank you so much for doing this!

rtorok-zr (approved):

PR #460, opened 2026-09-15
[rv_dm,test] Fix rv_dm_lc_disabled_tl test

rtorok-zr:
This test was broken by https://github.com/pavona/pavona/commit/6230bb6527f1a929bd67ea31bd68291a76bcd71e, which removed the explicit Debug ROM entry in the Ibex reset ePMP configuration from Egret. This happened because the test attempts to check the debug ROM is operational in the correct lifecycle states by writing to it from Ibex, which, unlike RV_DM itself, is still subject to ePMP constraints. This commit fixes this defect by adding an explicit ePMP entry for the debug memory in the test script, and also upgrades the test to use DT for multi-top compatibility.


mkannwischer (approved):
Thanks @rtorok-zr! 

I've checked that these tests pass now:
- //sw/device/tests:rv_dm_lc_disabled_tl_dev_fpga_cw340_sival
- //sw/device/tests:rv_dm_lc_disabled_tl_rma_fpga_cw340_sival
- //sw/device/tests:rv_dm_lc_disabled_tl_test_unlocked1_fpga_cw340_sival

The fix looks reasonable to me, but admittedly I'm not very familiar with the memory protection. @pqcfox, could you also take a look?

pqcfox (approved):
Looks excellent to me! Nice catch @rtorok-zr!

PR #461, opened 2026-09-16
[rom] Fix the load access fault in the OTP measurement

mkannwischer:
rom_e2e_keymgr_init_otp_meas fails on the CW340 with BFV:05495202.

8596886cfd replaced the `+` joining the OTP base address to the window and digest offsets with a `,`, so rom_measure_otp_partitions dereferences the bare offset.

I'll separately check if `-Wcomma` could have maybe avoided this bug.


rtorok-zr (approved):
Thanks for the catch. TIL that `(a, b)` is actually a valid expression in C, which evaluates to the value of `b`.

This code path was never going to work as a result, because the offset within OTP would try to be used as a full address.

mkannwischer:
> Thanks for the catch. TIL that `(a, b)` is actually a valid expression in C, which evaluates to the value of `b`.

Thanks for the review!
I think `-Wunused-value` (part of `-Wall`) will complain about `(a,b)` if a is just some value because it's just dead, but if a is a function call that may have side-effects it isn't dead and it won't complain. `-Wcomma` would have caught it.

adurbin-zr:
Were the function signatures for `dt_otp_ctrl_reg_block` not forward declared? When I search I only see this:

```
$ git grep dt_otp_ctrl_reg_block
sw/device/silicon_creator/rom/boot_rom_ext.c:      (unsigned char *)(dt_otp_ctrl_reg_block(kDtOtpCtrl,
sw/device/silicon_creator/rom/boot_rom_ext.c:      (unsigned char *)(dt_otp_ctrl_reg_block(kDtOtpCtrl,
```

This seems unexpected to me. Can we not use: `-Wimplicit-function-declaration -Werror` ?

adurbin-zr:
> Were the function signatures for `dt_otp_ctrl_reg_block` not forward declared? When I search I only see this:
> 
> ```
> $ git grep dt_otp_ctrl_reg_block
> sw/device/silicon_creator/rom/boot_rom_ext.c:      (unsigned char *)(dt_otp_ctrl_reg_block(kDtOtpCtrl,
> sw/device/silicon_creator/rom/boot_rom_ext.c:      (unsigned char *)(dt_otp_ctrl_reg_block(kDtOtpCtrl,
> ```
> 
> This seems unexpected to me. Can we not use: `-Wimplicit-function-declaration -Werror` ?

I think all C compilation units should be compiled with  `-Wall -Werror`, fwiw. Is it not currently?

PR #464, opened 2026-09-16
[toolchain] Enable -Wcomma

mkannwischer:
The comma operator turns typos into valid C by discarding the left operand: `(f(), b + c)` evaluates to `b + c`, as in https://github.com/pavona/pavona/pull/461. -Wall only catches this when the left operand has no side effects.

- Depends on #461


fragglet (approved):
Nice! Was not aware of this warning option, might add it for some of my own projects!

PR #465, opened 2026-09-17
[test] Mark the failing CW340 ROM_EXT tests as broken

mkannwischer:
It looks like these tests have been failing since #305 landed: 39 of the tests hang before they print anything,
and the epmp_test expects an ePMP table that no longer matches. Mark them as failing until someone has time to investigate.


rtorok-zr (approved):
LGTM. These tests will be re-evaluated during the ROM_EXT port to Dragonfly.

PR #467, opened 2026-09-18
[dvsim] Remove directories by type rather than by errno

mkannwischer:
rm_path fell back to shutil.rmtree only when os.remove raised IsADirectoryError. macOS reports EPERM rather than EISDIR for a directory, so the fallback never ran and every second dvsim run into the same scratch root failed while recreating dispatched:

  PermissionError: [Errno 1] Operation not permitted: '.../dispatched'

Unlink files and symlinks and remove everything else with rmtree.

This was encountered when running the following on macOS:
```
  ./util/dvsim/dvsim.py hw/ip/uart/dv/uart_sim_cfg.hjson -i uart_smoke -t vcs --dry-run
  ./util/dvsim/dvsim.py hw/ip/uart/dv/uart_sim_cfg.hjson -i uart_smoke -t vcs --dry-run
```
  The second run ends with:
```
  ERROR: [utils] Failed to remove .../uart-sim-vcs/dispatched:
  [Errno 1] Operation not permitted: '.../uart-sim-vcs/dispatched'.
  ...
  PermissionError: [Errno 1] Operation not permitted: '.../uart-sim-vcs/dispatched'
  ```

NB: Admittedly, `dvsim` doesn't claim any support for macOS, but this looks easy enough to fix and cleaner on Linux as well.


apinise (approved):
Looks good to me, thanks for fixing this for macOS!

PR #468, opened 2026-09-18
[doc] Add experimental macOS steps to the getting started guide

mkannwischer:


mkannwischer:
It's unfortunate that Xcode 27/macOS 27 does not work yet, see #470, but I don't think there is anything we can do about that right now.

